PR #209, opened 2020-10-15
feat: automatic breadcrumbs recalculation based on url hash #171

### LINKED ISSUE #171: Make Breadcrumbs component respond to url changes
Currently, the Breadcrumbs component is static. It would be easier for app devs to make it dynamic based on current URL via @meganjkurka

mturoci:
This PR adds `auto` prop to turn on automatic breadcrumbs based on URL hash. Also contains a fix to select the correct tab (tab routing) if you change url by hand instead of clicking.

The API for this component needs a discussion since it doesn't make sense to allow both `items` and `auto` - mutually exclusive.
Since we have no semantic props validation for now we can:
* Make both props optional - allows specifying an empty breadcrumb component which would have no effect and just break the UI as it makes no sense to specify no props for this component.
* Keep `items` mandatory, `auto` optional (current implementation) and provide documentation that says to specify empty `items` if you want to have `auto` turned on.
* Other suggestions are welcome. 

Closes #171


mturoci:
Closing as not going to merge/implement.

Issue #150, opened 2020-09-23
Textbox with trigger=true clears itself after submitting value


lo5:
**Q SDK Version, OS**
0.6.0

**Steps To Reproduce**
Run this example: https://github.com/h2oai/qd/blob/master/py/examples/table_filter_backend.py

**Actual behavior**
Search field is cleared as soon as you type.

**Expected behavior**
Search field should not be cleared.

PR #229, opened 2020-10-22
Candidate alternative strategy for front-end state management

mturoci:
This PR allows controlled rerendering form components by adding a unique id to each form component that is created at python side. This is truly the only option from my point of view since python server is our single source of truth for the app state as a whole. This allows two cases we needed to cover:
* Do not rerender a form component if python instance is not changed itself - only a particular property (like visible for instance changed). This keeps the form component state as is and just updates the prop.
* Change the whole `items` array on card (wizard tour example). If the references in items list change, we want to destroy the old components and create new ones.

TODO: Add R generation also.

This would also allow us to keep state on component level in UI components - resolves https://github.com/h2oai/wave/issues/150 - I have checked with datepicker component for now and works just fine.

This unique id is considered an implementation detail and should not be exposed to app developers.

### DIFF
--- a/py/h2o_wave/types.py
+++ b/py/h2o_wave/types.py
@@ -4030,6 +4030,7 @@ class VegaVisualization:
             visible,
         )
 
+id_ = 0
 
 class Stat:
     """Create a stat (a label-value pair) for displaying a metric.
@@ -4234,6 +4235,9 @@ class Component:
             stats: Optional[Stats] = None,
             inline: Optional[Inline] = None,
     ):
+        global id_
+        id_ += 1
+        self.id = id_
         self.text = text
         """Text block."""
         self.text_xl = text_xl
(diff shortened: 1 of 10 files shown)


mturoci:
Implemented uuid instead of incrementals.

lo5:
This solution will not work, for the same reason I mentioned above: we don't have a mechanism that will sync the UUID across multiple workers. I'll handle this for #150.

Are there other changes in the PR that need to make it to master?

mturoci:
> This solution will not work, for the same reason I mentioned above: we don't have a mechanism that will sync the UUID across multiple workers. I'll handle this for #150.

Okay.
> Are there other changes in the PR that need to make it to master?

None that I am aware of.

lo5:
Thanks @mturoci - closing.

PR #245, opened 2020-10-26
Theming API

### LINKED ISSUE #51: Theming API
Let's tackle this in two phases:
1. No user-specified themes
2. User-specified themes

For phase 1, we'll ship with the three themes that already exist, and the theme can be specified using a `theme: str` attribute on the `meta_card`. 

When a `theme` is specified, the UI should use that theme. 

Let's discuss and agree on an approach on a branch.

mturoci:
First implementation of Theming API for a discussion.

We have 2 architectural routes to take:
* Paint the App with default colors and repaint it if meta includes `theme` - current implementation of this PR.
* Create a separate logic for requesting the theme during initial loading and socket connection setups - pros - don't need to repaint the whole app again, cons - makes code more ugly.

CSS variables used instead of going CSS-in-JS way. This has a few advantages like not forcing React to use its diff rerendering algorithm - everything is handled in CSS. Allows easy future dark/light mode implementation (similar to current docs) without need for above mentioned rerendering and possible loosing component state. This approach also simplifies the change detection. Previously, each component had per file styles that used imported color values from `palette`. It would be a headache trying to both listen for theme changes and rerender the correct component instances to apply the new colors.

I moved all the color related stuff to global `index.css`. Apart from above mentioned advantages, it also saves us from initial JS work that was done before in `toPalette` that generated particular text tones on the fly. Having this already prebuilt and cached in CSS seems more elegant to me. Same goes for font generating as well. This saves a bit of performance - main thread can be used for more important stuff) and memory - no need to keep the jsons in app memory whole time. This is not a significant saving ofc. If we agree on upfront styles building,  we might consider using SASS or some other css preprocessor to keep it maintainable.

We have 4 types of colors:
* base colors - do not change (red, yellow, green etc.)
* theme colors - card, text, page
* fluent palette colors
* text tones - alpha (opacity) based text colors

I had problems with Fluent palette - the currently implemented solution is rather considered as a workaround than a real solution, check [this issue](https://github.com/microsoft/fluentui/issues/13814)  for more info.
I also could not handle text tones elegantly as rgba function in pure CSS doesn't take hex value as a first arg. 

TODO: Add corresponding Fluent palettes for `dark` and `neon` themes. Note that Fluent theme generator (and some components) do not work 100%.

Closes #51

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,6 +1,7 @@
 {
   "recommendations": [
     "dbaeumer.vscode-eslint",
+    "michelemelluso.code-beautifier",
     "ms-python.python",
     "golang.Go",
     "DavidAnson.vscode-markdownlint"
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,7 +1,7 @@
 {
   "recommendations": [
     "dbaeumer.vscode-eslint",
-    "michelemelluso.code-beautifier",
+    "sibiraj-s.vscode-scss-formatter",
     "ms-python.python",
     "golang.Go",
     "DavidAnson.vscode-markdownlint"
--- a/py/examples/dashboard.py
+++ b/py/examples/dashboard.py
@@ -8,8 +8,8 @@ from synth import FakePercent, FakeCategoricalSeries
 
 fake = Faker()
 
-light_theme_colors = 'red pink purple violet indigo blue azure cyan teal mint green amber orange tangerine'.split()
-dark_theme_colors = 'red pink blue azure cyan teal mint green lime yellow amber orange tangerine'.split()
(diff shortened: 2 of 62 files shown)


mturoci:
Rebased.

@lo5 note that only 2 built-in color themes are available - light and neon (dark theme needs Fluent theme to be generated).
Also we should file an issue for specifying custom themes for app devs.

lo5:
> dark theme needs Fluent theme to be generated.

Hmm. I overlooked this. Let's discuss before merging.

mturoci:
@lo5 PR rebased and checked, feel free to have a look.

https://user-images.githubusercontent.com/64769322/105325012-66bc3b00-5bcc-11eb-9147-79c1764f1a13.mov

lo5 (approved):
Thanks @mturoci!

PR #268, opened 2020-11-02
Improved git hooks setup

### LINKED ISSUE #274: Move non-js / non-node.js assets out of the root dir
Let's get the package.json, eslint, husky, etc. configs out of the root directory into ui/ or ide/ where they should belong.

mturoci:
* removed unnecessary Typescript typechecking step - already done via `@typescript/eslint`
* added `.eslintignore` config files to corresponding project modules to prevent unnecessary linting (like `node_modules` for example)
* installed `eslint`, `jest` into root `node_modules` - this way we can use already downloaded binaries for git hooks instead of pulling from npm on each commit
* refactored lint step for particular file types into it's corresponding npm run tasks

Closes #274


lo5 (approved):
Looks good - thanks @mturoci!

mturoci:
* rebased
* minor linting erros resolved
* a few config files added to `.eslintignore`

Note: `make all` needs to be run after merging in order to get everything set up correctly.

cc @lo5

lo5 (approved):
Thanks @mturoci!

Issue #17, opened 2020-07-02
Block UI while waiting for app backend to respond


lo5:
**Goal**
Prevent unhandled app errors due to double-submission.

PR #275, opened 2020-11-04
Update blocking overlay for css only solution

geomodular:
If PR merged, blocking overlay will be updated of css solution.

The original solution worked but had opacity transition to non-blocking overlay as well. i.e.: if no blocking state on server happened in 2s+, user would have experienced the same - correct transition.

This PR separates classes and prevents transition mixing.

Continuation of #17 and #236

### DIFF
--- a/ui/src/app.tsx
+++ b/ui/src/app.tsx
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import * as Fluent from '@fluentui/react'
-import { box, on, WaveErrorCode, WaveEventType } from 'h2o-wave'
+import { box, WaveErrorCode, WaveEventType } from 'h2o-wave'
 import React from 'react'
 import { stylesheet } from 'typestyle'
 import Dialog from './dialog'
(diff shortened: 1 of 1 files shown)


lo5:
@mturoci - It would be good if you can review and try this out so that we can close it.

mturoci (commented):
I have verified this PR and it works. Also the code is much cleaner - good job @geomodular!

Small nit worth discussion:
We could increase `zIndex` within `busyOverlay` to something like 999 to be absolutely sure this won't get accidentally overwritten in future - despite the fact that z-index should be used sporadically, there are usecases when it might make sense.

ui/src/app.tsx
```diff
@@ -43,12 +43,17 @@ const
...
       left: 0, top: 0, right: 0, bottom: 0,
+      opacity: 0,
+      zIndex: -1,
     },
     busyOverlay: {
-      display: 'block',
+      position: 'fixed',
+      left: 0, top: 0, right: 0, bottom: 0,
+      opacity: 0.8,
+      zIndex: 1,
```
mturoci, line 55: Can we increase `zIndex` to some higher value (e.g. `999`) to make sure it won't get overriden by accident in future?

mturoci (changes requested):
Thanks @geomodular! Great job, this CSS only solution is way cleaner.

One comment from me.

ui/src/app.tsx
geomodular: Yes, sorry, I meant to but forgot.

PR #278, opened 2020-11-05
IDE UI

mturoci:
Basic IDE UI with the following features:
* Create an app by name.
* Go to recent apps.
* Various views (split, code, preview).
* Dirty state indicator.
* Single file support only for now (app.py).
* Saving cursor position when travelling between file tabs.

Minor build features:
* App reloads on each file change (hot reload does not work 100% - still experimental - https://github.com/pmmmwh/react-refresh-webpack-plugin)
* Added minification

Possible features to be added to built:
* Split configs to prod, dev and base to make it cleaner
* Add eslint webpack integration.
* Add asset optimizations (images mostly)
* The initial app.js chunk has 2.7mb - we should probably try to split it into separate chunks. Load editor javascript just after visiting the AppPage, not Homepage already.

Needs discussion:
Listing files in an expected way (see comments below).

### DIFF
--- a/ide/__mocks__/monaco-editor.js
+++ b/ide/__mocks__/monaco-editor.js
@@ -5,6 +5,7 @@ export const editor = {
     addAction: jest.fn(),
     dispose: jest.fn(),
     onDidChangeModelContent: jest.fn(),
+    onDidChangeCursorPosition: jest.fn(),
     focus: jest.fn()
   }))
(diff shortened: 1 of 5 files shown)


mturoci:
Merged in 84e9aaff8f59cf70dbaeb32cace75eb94d1df6de.

Issue #250, opened 2020-10-28
Package server binaries with PyPI package


lo5:
Our getting-started experience is a bit complicated: you have to both (a) download a release and (b) perform a`pip install` of the Python driver.

We should improve this experience by including the release binaries and examples inside the `.whl`.

Tasks:
- Build OS/arch specific .whl files; include server and examples.
- Make `wave run` automatically start the Wave daemon using the included executable if not running already.
- Documentation.

lo5:
Via https://github.com/h2oai/wave/discussions/1124

This is mainly packaging work:
- Modify setuptools / `setup.py` and test against PyPI (90% of the work)
- Make `wave run` automatically launch `waved` (using `subprocess` or `POpen`).
 
Today we push one single platform-neutral `none-any` `whl` file to PyPI. Instead, we need to push at least 3 platform-specific whl files - `x86_64` for windows, osx and linux. We already have makefile tasks that produce `waved` for windows, osx and linux - it's just a matter of building them first and copying them into the whl build directory before invoking `setup.py bdist_wheel`, so that they're available inside the `whl` file. Once pip-installed, `wave run` should be able to run `waved` using a relative path.

PR #288, opened 2020-11-09
Update tour.md

mtanco:
Added example of running `wave` command from the environment as I spent 20 minutes trying to figure out why `wave run --no-reload examples.tour` was giving me permission errors. I did not have venv sourced as that is not how I was running wave apps previously.

### DIFF
--- a/website/docs/tour.md
+++ b/website/docs/tour.md
@@ -52,6 +52,10 @@ Finally, run the tour:
 wave run --no-reload examples.tour
 ```
 
+:::tip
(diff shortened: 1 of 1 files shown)


lo5:
@mtanco - This will also need the corresponding Windows instructions, and overall would make the docs a tad too noisy, since the docs already mention right there that you have to source the venv. 

When #250 ships, this will no longer be a problem, since we'll be recommending a global `pip install`.

mtanco:
Makes sense to me, thanks!

PR #303, opened 2020-11-18
feat(dropdown): display a dialog with filtering for > 100 choices

### LINKED ISSUE #153: Ability to search values in the dropdown component by typing the text
In case a dropdown contains too many values, it would be helpful to have possibility to search for the value by typing in the first letters.

mturoci:
I am not sure I understand the meaning of `caption` in `ui.choice` as proposed in #153, so not implemented yet. What is the difference between `caption` and `label`? 

Closes #153

### DIFF
--- a/ui/src/dropdown.test.tsx
+++ b/ui/src/dropdown.test.tsx
@@ -231,6 +231,35 @@ describe('Dropdown.tsx', () => {
       expect(getByTestId(name)).toHaveValue('Choice 1')
     })
 
+    it('Closes dialog on cancel', () => {
+      const { getByText, getByTestId, getByRole } = render(<XDropdown model={{ ...dialogProps, values: ['1'] }} />)
+
+      fireEvent.click(getByTestId(name))
+      fireEvent.click(getByText('Cancel'))
+
+      expect(getByTestId(name)).toHaveValue('Choice 1')
+      expect(getByRole('dialog')).not.toBeVisible()
+    })
+
+    it('Closes dialog on submit', () => {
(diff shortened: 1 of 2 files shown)


shanaka-rajitha:
@lo5 @mturoci  I have added few tweaks to the existing dropdown component.

**Variations on default and active state of dropdown**
Dropdown list

**Default dialog with filtering using checkbox**

- Since user may select any number of choices we need to use checkbox, radio buttons are used when user must select exactly one choice.

Modal - Dropdown list


**Default dialog with filtering using search option**

Modal - Dropdown list-2

mturoci:
I have updated design according to Fluent:

https://user-images.githubusercontent.com/64769322/118106622-13216780-b3de-11eb-9efd-53c89f79df7d.mov

* used Fluent Searchbar instead of custom one to keep consistency
* used Detaillist with always visible checkboxes instead of regular checkboxes. The technical reason is that Detaillist is virtualized and checkboxes are not. Considering we already had cases where people wanted to filter large data using `ui.picker` and faced perf issues, this seem like a good quick solution. Other alternative would be to write virtualization ourselves, but I think it would be a waste of time - cc @lo5 
* Added select / unselect all as proposed in design, but used styles we use for other components currently to keep consistency.

Needs discussion:
* The current UX clears current selection when filtering. This is caused because Fluent tightly couples items in Detaillist with selection items for some reason. The question is whether this is good UX as it is or we should keep the selection throughout filtering - cc @lo5 @shanaka-rajitha

PR #316, opened 2020-11-23
feat: allow submitting text phrases within markdown

### LINKED ISSUE #293: Add mechanism to submit click events from text/phrases inside ui.text*()
Via @Far0n 

e.g. `The quick brown [fox](?fox) jumps over the lazy [dog](?dog)` can submit `q.args.fox=True` or `q.args.dog=True` when clicked.

mturoci:
This PR intercepts `markdown-it` text rendering process and adds an `onclick` listener if the href attribute is prefixed with `?`. The listener code is inlined, otherwise it would require a globally available function. Apart from the fact that globals in general are not a good idea, it would be of no use since webpack changes function names during prod build, so the strings would not match either way.

This solution proposes use of custom events, both dispatched and listened on `window` object.

Closes #293

### DIFF
--- a/py/examples/markdown_submit_text.py
+++ b/py/examples/markdown_submit_text.py
@@ -1,0 +1,13 @@
+# Markdown / Submit / Text
+# Use "?" to prefix the desired q.args.key that you want to have submitted after clicking a phrase.
+# ---
+from h2o_wave import site, ui, main
+
+page = site['/demo']
+
+page['example'] = ui.markdown_card(
+    box='1 1 3 -1',
+    title='I was made using markdown!',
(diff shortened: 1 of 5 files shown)


mturoci:
@lo5 does the proposed API (`The quick brown [fox](?fox) jumps over the lazy [dog](?dog) can submit q.args.fox=True or q.args.dog=True when clicked.`) still look good to you? If yes, I will merge.

lo5:
Yes, looks good. It would be good if the example did not use index based access since we want to discourage that practice.
```
q.page['clicked'].items[1]
```

PR #358, opened 2020-12-07
feat: add ability to show/hide legend for plots #351

### LINKED ISSUE #351: Allow hiding legend in plots
via @isarth23

Proposal: Add `legend` prop to `plot_card`, defaults to `True`.

Implementation note: `chart.legend(false)` is all it takes.

cc @lo5 @geomodular

mturoci:
Closes #351


lo5:
We need to think this API through. Let's punt this for post 0.11.

PR #360, opened 2020-12-07
Sort table group titles when grouped

### LINKED ISSUE #336: Sort Table Values when Grouped
It would be nice to automatically sort grouped values ... or at least make that an option in the table. 
image

mturoci:
* Added default sorting for string | number | date types.
* Fixed minor deprecation warning for small_series_stat.
* Added unit tests.

Closes #336

### DIFF
--- a/ui/src/table.test.tsx
+++ b/ui/src/table.test.tsx
@@ -488,6 +488,58 @@ describe('Table.tsx', () => {
       expect(container.querySelectorAll('.ms-GroupedList-group')).toHaveLength(tableProps.rows.length)
     })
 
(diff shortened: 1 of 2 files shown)


lo5:
I got confused by the title of this PR.

Is there a corresponding example that demonstrates that the sorting works as expected? If not, it would be good to add one.

mturoci:
> I got confused by the title of this PR.

This was leftover from other PR by accident. Fixed.

>Is there a corresponding example that demonstrates that the sorting works as expected? If not, it would be good to add one.

This sorting is just for initial groups creation meaning if you select group by column, the groups will be shown automatically sorted. Not sure what examples you mean.

PR #362, opened 2020-12-07
Markdown / HTML within message_bar component

### LINKED ISSUE #356: Markdown or HTML for message_bar
image

It would be nice to format the text in message bars. Low priority, just a thought. 

Tested both:
```
ui.message_bar(type='info', text='<b>App Developer:</b> <br />'
                                         'Write the requirements and expectations of the dataset here!'),

ui.message_bar(type='info', text='**App Developer:** \n'
                                         'Write the requirements and expectations of the dataset here!'),
```

mturoci:
Generated HTML code has some default margins for some of the elements what does not align with default message bar type icon, had to adjust first-child / only-child appropriately.

Closes #356

### DIFF
--- a/ui/src/message_bar.tsx
+++ b/ui/src/message_bar.tsx
@@ -40,6 +40,7 @@ const
   css = stylesheet({
     messageBar: {
       $nest: {
+        // Adjust spacing to align with Fluent Messagebar icon.
         '.wave-markdown > *:first-child': { marginTop: 0 },
         '.wave-markdown > *:only-child': { marginBottom: 0 },
       }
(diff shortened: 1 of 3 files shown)


ui/src/message_bar.tsx
```diff
@@ -22,6 +24,14 @@ export interface MessageBar {
 }
 
 const
+  css = stylesheet({
+    messageBar: {
```
lo5, line 28: Can we add a comment here explaining what this CSS is doing / why this is required?

lo5 (changes requested):
Thanks @mturoci - minor comment about a comment ;-)

ui/src/message_bar.tsx
mturoci: Good catch @lo5. This CSS definitely deserves a comment. Will add.

Issue #353, opened 2020-12-04
Suggestions for Improving Tour.py


mtanco:
Reoccuring feedback from new Wave Developers: There are too many examples in documentation (tour), and many of them are too niche or small in functionality to help me get up and coding fast. 

Suggestions:

- [x] There should be a `hello_world_script` (exactly the same as the current `hello_world`) and a `hello_world_app` (write in the format of `@app`)

- [x] None of the apps should use `faker` - this is our own special library which confuses people. Also, examples that are repeatable are easier to understand so it would be better to have a set dataset to use rather than make a fake one every time. Maybe use fake to created a csv to ship and always use?

- [ ] Notes for all "real" apps (To Do List, Wizard, Issue Tracker, Dashboard)
  * Have a header at the top explaining what the app does
  * Meta card 
  * Should only use the `app` function for routing, all content rendering should be in a single function per view
  * Should not use Faker or synth 
  * If cypress is used it needs to have comments explaining what it is and how to use it - probably all of these should have testing
  * Are global variables okay, I thought this was bad practice?
  * I cannot imagine a new to wave person being able to learn from the current dashboard example but I don't have specific solid feedback on how to fix this, it's just... a lot going on!

- [ ] Notes for Widget Examples
  - [ ] Stats Cards
    - [x] 12 current examples should be rolled into one app
    - [ ] Avoid column names like foo and qux as this is hard to internalize
    - [ ] Avoid faker

  - [ ] Layout / Position
    * This should be the first app - explaining the grid I think is really important
    * Should this be rewritten to be responsive, or can we add a responsive version of this?
    * The notes in this example are AWESOME

  - [x] Form (all of them)
    * I like that the current form.py file has everything - i think we should focus on making this really nice and get rid of all of the other `Form / Thing`. There should be one example that is a form with non-interactive elements and then one form with interactive stuff (Just pick one element, color picker or checkbox or something - people are smart and can use the UI docs to replace if they have one good example to start with)
    * Each element should follow the same format for any text: `element name: some example`
      * `ui.text_xl(content='text_xl: use for headings')`
    * Add new lines between different types of conent
    * Do general clean up (like right now all 3 checkboxes have the same `name` element)
    * Add any important notes (like explain that the name element will be used in q.args when you click the thing)
    * I think if everything is in one function then all the text from `html`, `menu`, and `spec` should be too
    * avoid Faker if possible

[...]

PR #370, opened 2020-12-08
Move all stat card examples to single example

dulajra:
For https://github.com/h2oai/wave/issues/353

Move all stat card examples to a single example


lo5 (changes requested):

dulajra:
This is not needed because of dashbaord.py

PR #371, opened 2020-12-08
style: center breadcrumb component vertically #337

### LINKED ISSUE #337: Tab Text Not Vertically Centered
This is a question not a bug per se. I noticed that the text in this tab isn't vertically centered. Is that expected? 

image

https://github.com/h2oai/wave-apps/pull/10

mturoci:
Closes #337

### DIFF
--- a/py/examples/breadcrumbs.py
+++ b/py/examples/breadcrumbs.py
@@ -13,7 +13,7 @@ from h2o_wave import main, app, Q, ui
 async def serve(q: Q):
     if '#' in q.args:
         hash_ = q.args['#']
-        q.page['breadcrumbs'] = ui.form_card(box='1 1 2 5', items=[
+        q.page['breadcrumbs'] = ui.form_card(box='1 1 2 2', items=[
             ui.text(f'#={hash_}'),
             ui.button(name='show_breadcrumbs', label='Back', primary=True),
         ])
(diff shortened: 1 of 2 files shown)


lo5 (approved):
Thanks @mturoci.

mturoci:
Rebased.

lo5 (changes requested):
Needs a rebase. Also, can we use the same strategy as https://github.com/h2oai/wave/commit/565b6f60c604970bb9d83ab2cc2e2a83774073fe? I'd like to avoid using `Fluent.Stack`.

mturoci:
* Rebased
* The reason why this was not centered vertically was that Fluent's implementation set top margin to 11px and bottom margin to 1px. I have set them both to 15px instead (feel free to adjust as you see fit).

image

(The card has `CardEffect.Transparent`, I just temporarily removed in order to clearly show the centering in screenshot)

lo5:
@mturoci - it's not centered. Better to remove Fluent's 11px top-margin. 

Screenshot from 2021-01-20 09-28-23

mturoci:
The margin is already out. Are you sure you are running with latest client changes?
image

lo5 (approved):
I see it now - my bad. 
Thanks @mturoci!

PR #373, opened 2020-12-09
ML API for H2O-3

geomodular:
If PR merged, `ML API` will be added to the Wave project. This is the first version of API based on `H2O-3` backend and it's capable of building the model, predicting *on instance*, loading and saving the binary model. `ML API` is independent part of a *Wave* with no internal connections between other Python modules.

## API

- To build a model use [`build_model()`](https://github.com/h2oai/wave/blob/f43d5bd202f7b248064b81e23af3890d14fd03a2/py/h2o_wave/ml.py#L187).

- To predict on an backend instance use a method available to a built model: [`model.predict()`](https://github.com/h2oai/wave/blob/f43d5bd202f7b248064b81e23af3890d14fd03a2/py/h2o_wave/ml.py#L68).

- To save and load models use [`save_model()`](https://github.com/h2oai/wave/blob/f43d5bd202f7b248064b81e23af3890d14fd03a2/py/h2o_wave/ml.py#L230) and [`load_model()`](https://github.com/h2oai/wave/blob/f43d5bd202f7b248064b81e23af3890d14fd03a2/py/h2o_wave/ml.py#L245).

- To get a model from a running `H2O-3` instance (if any) use [`get_model()`](https://github.com/h2oai/wave/blob/f43d5bd202f7b248064b81e23af3890d14fd03a2/py/h2o_wave/ml.py#L212).

## Dependencies

`ML API` needs two packages to be able to operate inside the Python environment: `H2O` and `datatable`.

The `extras_require` was introduced inside the [`py/setup.py`](https://github.com/h2oai/wave/blob/f43d5bd202f7b248064b81e23af3890d14fd03a2/py/setup.py#L37). It separates dependencies from regular *Wave* using the `pip` brackets syntax. If extra parameter `ml` used the whole package plus the `ML API` dependencies will be installed.

For regular `pip` package:

```bash
pip install h2o-wave[ml]
```

For dev installation inside the `py/` directory:

```bash
pip install --editable .[ml]
```

## Examples

Full example with a histogram:
```python3
from h2o_wave import main, data, ui, site
from h2o_wave.ml import build_model

page = site['/prediction']

train_set = './creditcard_train_cat.csv'
test_set = './creditcard_test_cat.csv'

model = build_model(train_set, target='DEFAULT_PAYMENT_NEXT_MONTH')
prediction = model.predict(test_set)
# This will output: [(False, 0.8590936346577605, 0.14090636534223955), (True, 0.13186347630483963, 0.8681365236951604), (False, 0.7730244054420021, 0.22697559455799787), ...]

true_count = len([1 for p in prediction if p[0]])
false_count = len([1 for p in prediction if not p[0]])

v = page.add('example', ui.plot_card(
    box='1 1 4 5',
    title='Credit card category',
    data=data('category count', rows=[['First category', true_count], ['Second category', false_count]]),
    plot=ui.plot([ui.mark(type='interval', x='=category', y='=count', y_min=0)])
))

page.save()
```

Save model:
```python3
from h2o_wave.ml import build_model, save_model

model = build_model('./creditcard_train_cat.csv', target='DEFAULT_PAYMENT_NEXT_MONTH')
path = save_model(model, './')
```

Load model:
```python3
from h2o_wave.ml import load_model

model = load_model('./StackedEnsemble_AllModels_AutoML_20201209_155503')
prediction = model.predict('./Datasets/creditcard_test_cat.csv')
```

Predict using just one row:
```python3
from h2o_wave.ml import build_model

data = [["ID", "LIMIT_BAL", "SEX", "EDUCATION", "MARRIAGE", "AGE", "PAY_1", "PAY_2", "PAY_3", "PAY_4", "PAY_5", "PAY_6",
         "BILL_AMT1", "BILL_AMT2", "BILL_AMT3", "BILL_AMT4", "BILL_AMT5", "BILL_AMT6", "PAY_AMT1", "PAY_AMT2",
         "PAY_AMT3", "PAY_AMT4", "PAY_AMT5", "PAY_AMT6"],
        [24001, 50000, "male", "university", "single", 23, 2, 2, 0, 0, 0, 0, 51246, 49758, 48456, 44116, 21247, 20066, 8,
         2401, 2254, 2004, 704, 707]]

model = build_model('./creditcard_train_cat.csv', target='DEFAULT_PAYMENT_NEXT_MONTH')
prediction = model.predict(data)
# [(True, 0.31995987356563516, 0.6800401264343648)]
```

### Datatable

`Datatable` is used to transform the prediction output from an H2O-3 [here](https://github.com/h2oai/wave/blob/f43d5bd202f7b248064b81e23af3890d14fd03a2/py/h2o_wave/ml.py#L183). The reason for this is to provide the same output interface both for H2O-3 and DAI having both saving the predictions on disk. H2O-3 has also another ways to obtain predictions and those are:
- Using Python object. This produces string values and we would end up with retyping directly in Python.
- Using `pandas` object. That introduces a new package.

`Datatable` seems like a good option considering DAI is saving predictions onto disk.

Another thing to mention; this is needed because predictions are made on instance. In case of a future `deploy_model()` the predictions will AFAIK be row based. But I think it's good to support both batch and row (instance vs deployed).

### Storing models

Saving and loading of a model is available in `ML API`. This functionality is temporary however as `save_model()` will became `deploy_model()` later, once we have a functional infrastructure in place with `H2O-3` support. The goal was to provide a way to store and load model for `H2O-3` instance.

Since the models are stored as binary models, the downside is also the incompatibility between the `H2O` versions.

### H2O-3 instance

`ML API` spawns own instance of `H2O-3` unless the `H2O_WAVE_H2O3_URL` ENV variable is specified. If specified, the running instance will be used. e.g.:

```bash
 H2O_WAVE_H2O3_URL=http://192.168.1.6:54321/ python3 examples/ml_ex.py
```

### Automatic conversion

The target column is checked for a type and changed appropriately using [this function](https://github.com/h2oai/wave/blob/f43d5bd202f7b248064b81e23af3890d14fd03a2/py/h2o_wave/ml.py#L116). The conversion to a categorical type is made for all `str` types and `int` types if unique values for a column is below `INT_TO_CAT_THRESHOLD` (set to 50).

For the future release the separate attribute setting *classification* vs *regression* might be a required addition (considering DAI).

## Notes

If desired, a DAI backend using the static instance of DAI can be added fairly quickly as well. The backend will be using the same interface with an exception of saving and loading models. `get_model()`, however, will be working.

Part of https://github.com/h2oai/h2o-ai-cloud/issues/139 and #342

### DIFF
--- a/py/h2o_wave/core.py
+++ b/py/h2o_wave/core.py
@@ -28,6 +28,7 @@ class _Config:
         self.hub_address = _get_env('ADDRESS', 'http://127.0.0.1:10101')
         self.hub_access_key_id: str = _get_env('ACCESS_KEY_ID', 'access_key_id')
         self.hub_access_key_secret: str = _get_env('ACCESS_KEY_SECRET', 'access_key_secret')
+        self.h2o3_url = _get_env('H2O3_URL', '')
 
 
 _config = _Config()
--- a/py/h2o_wave/ml.py
+++ b/py/h2o_wave/ml.py
@@ -89,6 +89,7 @@ class _H2O3ModelBackend(WaveModelBackend):
     INIT = False
     MAX_RUNTIME_SECS = 60 * 60
     MAX_MODELS = 20
+    INT_TO_CAT_THRESHOLD = 50
 
     def __init__(self, model: H2OEstimator):
         super().__init__(WaveModelBackendType.H2O3)
@@ -113,14 +114,25 @@ class _H2O3ModelBackend(WaveModelBackend):
                 h2o.init()
             cls.INIT = True
 
-    @staticmethod
-    def build(filename: str, target: str, metric: WaveModelMetric, **aml_settings) -> WaveModelBackend:
+    @classmethod
+    def _is_classification_task(cls, frame: h2o.H2OFrame, target: str) -> bool:
+        target_type = frame.type(target)
+        if target_type == 'str':
+            return True
+        if target_type == 'int':
+            uniques = frame[target].unique()
+            if len(uniques) < cls.INT_TO_CAT_THRESHOLD:
+                return True
+        return False
 
-        _H2O3ModelBackend._ensure()
+    @classmethod
+    def build(cls, filename: str, target: str, metric: WaveModelMetric, **aml_settings) -> WaveModelBackend:
 
-        id_ = _H2O3ModelBackend._make_project_id()
-        aml = H2OAutoML(max_runtime_secs=aml_settings.get('max_runtime_secs', _H2O3ModelBackend.MAX_RUNTIME_SECS),
-                        max_models=aml_settings.get('max_models', _H2O3ModelBackend.MAX_MODELS),
+        cls._ensure()
+
+        id_ = cls._make_project_id()
+        aml = H2OAutoML(max_runtime_secs=aml_settings.get('max_runtime_secs', cls.MAX_RUNTIME_SECS),
+                        max_models=aml_settings.get('max_models', cls.MAX_MODELS),
                         project_name=id_,
                         stopping_metric=metric.name,
                         sort_metric=metric.name)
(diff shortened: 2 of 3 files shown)


geomodular:
Closing PR as it will be part of a separate repo and library.

PR #374, opened 2020-12-10
Removing Faker and Merging Table Examples

LenaRampula:
- Merged table_download.py, table_groupby.py into table.py
- Merged table_filter.py, table_sort.py,  table_search.py into table_column_settings.py
- Removed Faker from Table, flex, grid, list, repeat


lo5 (changes requested):

PR #378, opened 2020-12-10
feat: add items prop to header card #367

### LINKED ISSUE #310: header_card should accept a clickable logo
**Is your feature request related to a problem? Please describe.**
It should be possible to add a custom logo to an application's header. In many web apps, the logo also serves as a link back to the homepage, so it would also be nice to support this functionality. 

**Describe alternatives you've considered**
Perhaps I don't understand the `icon` argument. Is this in the docs somewhere? I'm assuming [these](https://developer.microsoft.com/en-us/fluentui#/controls/web/icon) are the options but I could be way off.

### LINKED ISSUE #367: Header card items
We should allow always visible toolbar at the rightmost side of header card for cases:
* When there are just a few menu items and it doesn't make sense to put them into expandable sidenav.
* When there is a need for always on screen visible commands.

cc @lo5 @geomodular

### LINKED ISSUE #722: Improve header card

Many apps use similar features such as user information, searching, alerts, etc. Currently, using individual components like ui.header and ui.section_cards which lacks an integrated feel

https://xd.adobe.com/view/714ead4b-4020-4ba6-ab95-efdc3a61bb0d-241d/

image

### LINKED ISSUE #811: Navigation and Side Menu Improvements ( To use as a template )
https://xd.adobe.com/view/05dadb5a-1157-4225-b74b-72e4c355e3cb-5089/

Based on the xd link attached , Shall we think of a way to improve the header and the side navigation ( collapsible ) across all the applications ? We might need to think about a way to improve it, by letting the user to customise the header, side navigation and the body color according to the needs. How can we achieve  this ?

The main points we should think is , 

1. Header 
2. Side nav bar
3. Breadcrumb with the Main Title of the Context 
4. Button ( colors and make buttons with icons, having a 4px border radius )  
5. Body 
6. Color configuration to the header, side nav and the body.  

It would be really great, if we can add the user's name and the picture , notification and setting icon , search bar on the header. ( sample design attached )

image

@lo5 @shanaka-rajitha @mtanco @mturoci @venky74  @surenH2oai @VijithaEkanayake

mturoci:
Closes #367
Closes #811 
Closes #722
Closes #310


lo5:
@mturoci - I haven't had a chance to try this - do you mind posting a screencap of this so that I can understand what it's doing. I have some local changes to the header as well.

mturoci:
Sure @lo5, basically it just adds a toolbar to header rhs.

https://user-images.githubusercontent.com/64769322/102590442-e14af400-4110-11eb-86e7-bd69526861bc.mov

lo5:
@mturoci The `CommandBar` looks odd here, especially because the hover effect and menus are white.

Is it possible to style the hover/dropdowns so that they are less jarring? It would be good to have the menu's background color match the header's background color.

mturoci:
@lo5 it would be doable, but requires very ugly hack. The command menu is not really a part of `CommandBar`, but works in a similar fashion like Dialog. It is positioned absolutely, away from the rest of the components and uses [portals](https://reactjs.org/docs/portals.html) to work. Also per Fluent implementation, the hover background is computed off other colors (probably background and text) so I would need to dig in their code to find the desired formula if we want to be 100% exact. Not sure if it is worth the time, but lmk if I shall proceed.

lo5:
We don't have to use the `CommandBar`. It could be `ContextualMenu` instances tied to links laid out horizontally. 
If a link is tied to a `ContextualMenu`, display a chevron. If not, leave it as is. We don't need icons prefixed to the links either (they look odd). See https://themes.getbootstrap.com/ for examples.

mturoci:
Good idea @lo5! I have used `CommandButtons` instead of links since it has already built in support for `ContextualMenu`. Here is the result:
image

mturoci:
Rebased on top of current master and redesigned:

image
image
image

## Proposed API

```ts
/**
 * Render a page header displaying a title, subtitle and an optional navigation menu.
 * Header cards are typically used for top-level navigation.
 */
interface State {
  /** The title. **/
  title: S
  /** The subtitle, displayed below the title. **/
  subtitle: S
  /** The icon, displayed to the left. **/
  icon?: S
  /** The icon's color. **/
  icon_color?: S
  /** The logo displayed to the left. Mutually exclusive with icon. **/
  image?: S
  /** The navigation menu to display when the header's icon is clicked. Recommended for mobile screens only. **/
  nav?: NavGroup[]
  /** Items that should be displayed on the right side of the header. */
  items?: Component[]
  /** Items that should be displayed in the center of the header. */
  secondary_items?: Component[]
  /** Header background color. Defaults to 'primary'. */
  color?: 'card' | 'transparent' | 'primary'
}
```

cc @lo5 ^^

Also there was a [special syntax](https://github.com/h2oai/wave/blob/master/ui/src/header.tsx#L68-L72) used for docstrings within API - used for web editor. @lo5 should I keep (and continue) in this syntax or can it be considered obsolete?

lo5:
The changes look good, including the removal of the special docstrings.

This bit is not quite clear to me:
> 'color?: 'card' | 'transparent' | 'primary'`

In what cases would `card` or `transparent` be helpful? Wouldn't we always want it to be `primary` (like the default in the PR), in which case we can get rid of `color` altogether.

mturoci:
@lo5 I am a bit hesitant about adding a dropdown context menu directly to `ui.persona` as I already foresee people putting it into `form_card` and doing funky stuff. I decided to reuse existing `commands` attr instead and override the existing header 3 dots menu with the rightmost chevron (see screenshots). Let me know what you think about this approach.

The original Fluent design for the context menu seems to be more clear than the proposed one, so I went with that instead.

image
image
image

>In what cases would card or transparent be helpful? Wouldn't we always want it to be primary (like the default in the PR), in which case we can get rid of color altogether.

Proposed designs suggest card color and I assume it's because using primary color for the header can be sometimes too noisy.  

I browsed a lot of dashboard designs and these 3 header colors were used the most.
https://www.behance.net/gallery/85516221/Dashboard-Design
https://dribbble.com/search/dashboard
https://pinterest.com/zerpixelung/dashboard-ui/

lo5:
> Proposed designs suggest card color and I assume it's because using primary color for the header can be sometimes too noisy.

OK. API looks good then.

> I am a bit hesitant about adding a dropdown context menu directly to ui.persona

Agree. It might be better to have a _new_ `Menu`/`ui.menu()` component for the profile/dropdown. The normal state would be just the (circular) image, and clicking on the image could display the dropdown. Similar to the top-right Github menus: 



The menu could have both `icon` and `image` properties, similar to your implementation for the info cards.

Having such a component would allow authors to plonk them into the header card's `items`, producing results similar to the Github screencap above.

mturoci:
>Agree. It might be better to have a new Menu/ui.menu() component for the profile/dropdown.

Agree.

## Latest implementation (with menu)

image
image
image

## Proposed menu API

```ts
/**
  Create a contextual menu component. Useful when you have a lot of links and want to conserve the space.
*/
export interface Menu {
  /** Commands to render. */
  items: Command[]
  /** The card's icon. Mutually exclusive with the image. */
  icon?: S
  /** The card’s image, preferably user avatar. Mutually exclusive with the icon. */
  image?: S
  name?: Id
}
```

lo5 (approved):
I'm assuming the menu's down-chevron is still visible when both `icon` and `image` are not supplied?

Looks good - let's merge. Thanks, @mturoci!

mturoci:
> I'm assuming the menu's down-chevron is still visible when both icon and image are not supplied?

Correct, it's still visible.

PR #381, opened 2020-12-11
Single native plot app

LenaRampula:
#353 
- Native plots include: Line, Area, Point, Heatmap, Histogram, Polar, Transpose, Labels, Titles, Annotation, Color, Size
- Not included: Smooth (added in description), Range, Negative, Helix (added in description), Map, Step, Path - those are not deleted yet
- Plot event and plot form should stay - minor changes.

Consider if not included plot types could be removed.


lo5 (changes requested):

PR #383, opened 2020-12-12
Form cleanup

LenaRampula:
- Clean up form.py and add comments
- removed all other form / X
- Leaving form_menu.py as there is no menu.py


lo5 (changes requested):

PR #387, opened 2020-12-14
Single Text App

LenaRampula:
- Single text app that includes header, label, text_sizes, markdown, markup, separator
- Merged Frame and Frame_path
- Merged template and template_data


lo5 (changes requested):

Issue #372, opened 2020-12-09
Add enter handler for textbox


VijithaEkanayake-zz:
Add enter handler with similar logic `onChange` has.

Steps to follow:
1. Setup project for dev.
2. Find `textbox.tsx`.
3. Implement the feature.
4. Find `textbox.py` and showcase your new feature there.
5. Add needed unit tests in `textbox.test.tsx`.

lo5:
The current textbox with a search icon (already supported) can double as a search box if we add `enter` key handling to it.

mtanco:
related to https://github.com/h2oai/wave/issues/348

Nicolapps:
I’m working with @sherrychen2, @monal20, and @kitty3190 on a PR for this issue (and also #1645, #495, and #1535) as part of a [Software Engineering class assignment](https://cmu-313.github.io//assignments/hw6).

Sachin-Ram:
is this issue still active?

PR #390, opened 2020-12-15
Implement searchbar_card for wave

VijithaEkanayake:
searchbar_card can be used to horizontally align textbox and button on the requirement. Fixes: https://github.com/h2oai/wave/issues/372


lo5:
Is there a strong reason for this to be in a card of its own? I'm not sure I understand the need for a separate search bar component (and nested/multiple search textboxes). The current textbox with a search icon can serve this purpose if we add `enter` key handling to it. 

What do you think @mturoci?

VijithaEkanayake:
@lo5 The issue we face was, we had to use two different cards to submit the text-box values if we are not using triggers.
image
Basically wanted to have a textbox and a button next to each other and thought it is a very common use case that many people will definitely try to put stuff next to each other. For our case having a search icon with enter key handling also fine I guess.

mturoci:
> The current textbox with a search icon can serve this purpose if we add enter key handling to it

Good suggestion. I see the point of this card only in solving the problem people were asking about since the birth of Wave itself - how to put a button next to a textbox.

I personally prefer searchbars with debounce / enter instead of buttons, but I can be biased as a software dev.

Nonetheless, if we still want to proceed with this card, I would do the following:
* Remove all the positioning and go with reasonable defaults - just put the button on the right side and align it. Right now this is not a searchbar implementation, but rather flexbox form I would say.
* Change the API to match textbox API.
* Add proper unit tests.

lo5:
> I personally prefer searchbars with debounce / enter instead of buttons

Agree. Button-less search boxes are almost universal now - Chrome, Google Search, Github, etc., so I don't see a strong UX reason to put a button next to a textbox. Let's go with the enter key handling.

VijithaEkanayake:
> > I personally prefer searchbars with debounce / enter instead of buttons
> 
> Agree. Button-less search boxes are almost universal now - Chrome, Google Search, Github, etc., so I don't see a strong UX reason to put a button next to a textbox. Let's go with the enter key handling.

Yeah, we can use enter key handling instead of separate card.

PR #394, opened 2020-12-16
Safari quirks

### LINKED ISSUE #382: Header Bar Changing Height
Wave 0.10.0 running locally on OS X - app was recently refactored by @dulajra to responsive AI which may be related?

https://h2oai.slack.com/files/UFY03NQ4V/F01GRGDMBNF/screen_recording_2020-12-11_at_12.21.04_pm.mov?origin_team=T0329MHH6&origin_channel=CFF49UG65

mturoci:
* AddListener Safari - https://github.com/mdn/sprints/issues/858
* Also had to specify a min-height in addition to height since Safari Flexbox implementation differs as well.

We can also use a addListener only in case addEventListener is not available, but I checked and this works in other browsers as well so didn't want to pollute the code.

Closes #382

### DIFF
--- a/ui/src/page.tsx
+++ b/ui/src/page.tsx
@@ -97,13 +97,13 @@ on(layoutsB, layouts => {
         listener = (mq: MediaQueryListEvent) => { if (mq.matches) layoutB({ layout, index }) },
         bp: Breakpoint = { layout, min, max, mq, listener }
 
-      mq.addEventListener('change', listener)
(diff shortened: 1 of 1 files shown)


lo5 (approved):
Thanks for catching this @mturoci!
Looks like this should be OK per https://caniuse.com/mdn-api_mediaquerylist_addlistener. Fingers crossed.

PR #395, opened 2020-12-16
feat: add gdpr notice to docs

mturoci:
* Added a notice that complies with rest of H2O pages - https://www.h2o.ai/
* LocalStorage is used to not bother user that has already dismissed the notice.
* Removed the red color as I consider it being a leftover from previous design.

image

cc @NidhiMehta


geomodular (dismissed):
Thanks @mturoci !

lo5 (dismissed):
Thanks @mturoci!

lo5 (approved):
Thanks @mturoci @NidhiMehta @mmalohlava!

PR #403, opened 2020-12-17
docs: add shields to main readme

mturoci:
Added a few shields to readme that might be worth. Looking for opinions.

image
vs
image

Further possible Readme improvements - do not use links directly, but via `[](url)` to make it consistent.

Note that I golang and nodejs versions are hardcoded and simply taken from my machine, feel free to point out if we support lower versions as well.

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,3 +1,5 @@
 {
-  "line-length": false
+  "line-length": false,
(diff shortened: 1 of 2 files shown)


lo5 (approved):
Looks awesome! Thanks @mturoci!

PR #405, opened 2020-12-17
Forward refresh token to python client

### LINKED ISSUE #365: Propagate OIDC refresh token to Python client.
`Access` token is already accessible in Python client. For us to be able to access DAI and MLOps we need also `refresh` token.

## Goal
Propagate `refresh` token the same way as `access` token to Python client. It should be present in `q.auth.refresh_token` if OIDC enabled.

geomodular:
If PR merged, the OIDC `refresh` token will be propagated to python client the same way as other properties. This change is needed because of API for DAI and MLops which require `refresh` token to be able to handle long running tasks.

The token is accessible in `q.auth.refresh_token`.

Closes #365

### DIFF
--- a/client.go
+++ b/client.go
@@ -48,19 +48,20 @@ var (
 
 // Client represent a websocket (UI) client.
 type Client struct {
-	id          string          // unique id
-	addr        string          // remote address
-	username    string          // username, or "default-user"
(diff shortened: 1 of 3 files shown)


lo5 (approved):
Thanks @geomodular!

PR #418, opened 2020-12-19
Add new plotting example(s) directly from dataframe

mtanco:
Addresses question from discourse with a new example that shows how you can use `df.values.tolist()` with Wave's `data` class to easily visualize data from a dataframe. I included two examples in a single file which should help people get started: one with visualizing ALL the data and one where you are aggregting in pandas and visualizing some trend. 

Hopefully this + the more advanced plot examples using Faker will help pandas folks get started. 

This would be the first example using `data(fields, rows)` vs. `data(fields, size)`

### DIFF
--- a/py/examples/plot_dataframe.py
+++ b/py/examples/plot_dataframe.py
@@ -8,39 +8,44 @@ import numpy as np
 page = site['/demo']
 
 n = 100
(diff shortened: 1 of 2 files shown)


pelumiojebode:
Support@bitpay.con

lo5 (approved):
Thanks @mtanco!

Issue #397, opened 2020-12-16
Tour not working on Windows 10


letsfigureout:
#### Wave SDK Version, OS

0.10.0, Windows 10, Python 3.7.8

#### Actual behavior

When I try and run the tour, I get a NotImplementedError message in the console. The webapp fails to load also giving me the same error as in the console.

#### Expected behavior

I expected the tour to run

#### Steps To Reproduce

1. Created a new virtual environment (virtualenv)
2. Extracted the wave tarball 
3. Activate the newly created virtualenv and install dependencies (pip install -r examples\requirements.txt)
4. Start waved
5. Run the tour using "wave run --no-reload examples.tour"

```
(.venv) C:\code\python\h2oai\wave-0.10.0-windows-amd64>wave run --no-reload examples.tour
----------------------------------------
 Welcome to the H2O Wave Interactive Tour!

 Go to http://localhost:10101/tour        
----------------------------------------  
INFO:     Started server process [22080]  
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     127.0.0.1:51930 - "POST / HTTP/1.1" 200 OK
Unhandled exception
Traceback (most recent call last):
  File "c:\code\python\h2oai\wave-0.10.0-windows-amd64\.venv\lib\site-packages\h2o_wave\server.py", line 265, in _process
    await self._handle(q)
  File ".\examples\tour.py", line 197, in serve
    await show_example(q, catalog[route])
  File ".\examples\tour.py", line 166, in show_example
    await active_example.start()
  File ".\examples\tour.py", line 48, in start
    sys.executable, os.path.join(example_dir, self.filename)
  File "C:\Python\Python37\Lib\asyncio\subprocess.py", line 217, in create_subprocess_exec
    stderr=stderr, **kwds)
  File "C:\Python\Python37\Lib\asyncio\base_events.py", line 1544, in subprocess_exec
    bufsize, **kwargs)
  File "C:\Python\Python37\Lib\asyncio\base_events.py", line 462, in _make_subprocess_transport
    raise NotImplementedError
NotImplementedError
```

I did find these posts relating to asyncio and creating a sub-process in windows but I the suggestion there didn't work.

https://stackoverflow.com/questions/45699932/asyncio-stdout-failing
https://docs.python.org/3/library/asyncio-subprocess.html#windows-event-loop

shaz13:
[...]

PR #429, opened 2020-12-21
Replace asyncio.create_subprocess_exec() with python subprocess for tour

VijithaEkanayake:
Windows do not support SelectorEventLoop. To support tour in all the platforms, replaced asyncio.create_subprocess_exec() with python subprocess. Fixes: https://github.com/h2oai/wave/issues/397

Verified on: Windows 10. python 3.7.9 and python 3.8.6, python 3.9.1, MacOS Catalina 10.15.5

### DIFF
--- a/py/examples/tour.py
+++ b/py/examples/tour.py
@@ -33,16 +33,15 @@ class Example:
         self.is_app = source.find('@app(') > 0
 
     async def start(self):
+        # The environment passed into Popen must include SYSTEMROOT, otherwise Popen will fail when called
+        # inside python during initialization if %PATH% is configured, but without %SYSTEMROOT%.
+        env = dict({'SYSTEMROOT': os.environ['SYSTEMROOT']}) if sys.platform.lower().startswith('win') else {}
         if self.is_app:
-            # The environment passed into Popen must include SYSTEMROOT, otherwise Popen will fail when called
-            # inside python during initialization if %PATH% is configured, but without %SYSTEMROOT%.
             self.process = subprocess.Popen(
                 [sys.executable, '-m', 'uvicorn', '--port', _app_port, f'examples.{self.name}:main'],
-                **dict(env={'H2O_WAVE_EXTERNAL_ADDRESS': f'http://{_app_host}:{_app_port}', **(
-                    {'SYSTEMROOT': os.environ['SYSTEMROOT']} if sys.platform.lower().startswith('win') else {})}))
+                env=dict(H2O_WAVE_EXTERNAL_ADDRESS=f'http://{_app_host}:{_app_port}', **env))
         else:
-            self.process = subprocess.Popen([sys.executable, os.path.join(example_dir, self.filename)], **dict(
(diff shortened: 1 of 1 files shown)


Nenginyo (commented):
0 9

py/examples/tour.py
```diff
@@ -29,24 +29,45 @@ def __init__(self, filename: str, title: str, description: str, source: str):
...
         else:
-            self.process = await asyncio.create_subprocess_exec(
-                sys.executable, os.path.join(example_dir, self.filename)
-            )
+            self.process = subprocess.Popen([sys.executable, os.path.join(example_dir, self.filename)])
 
     async def stop(self):
         if self.process and self.process.returncode is None:
             self.process.terminate()
+            self.process.wait()
```
lo5, line 46: Is this `wait()` necessary?

py/examples/tour.py
```diff
@@ -29,24 +29,45 @@ def __init__(self, filename: str, title: str, description: str, source: str):
...
     async def start(self):
         if self.is_app:
-            self.process = await asyncio.create_subprocess_exec(
-                sys.executable, '-m', 'uvicorn', '--port', _app_port, f'examples.{self.name}:main', env=dict(
-                    H2O_WAVE_EXTERNAL_ADDRESS=f'http://{_app_host}:{_app_port}'
-                ),
-            )
+            self.process = subprocess.Popen(
+                [sys.executable, '-m', 'uvicorn', '--port', _app_port, f'examples.{self.name}:main'],
+                env=self.adjust_env_for_platform(dict(H2O_WAVE_EXTERNAL_ADDRESS=f'http://{_app_host}:{_app_port}')))
```
lo5, line 39: If I'm not mistaken, all this is doing is:
"if the platform is Windows and `SYSTEMROOT` is present in the OS env, copy it to `env`"

If so, it's simpler to inline it here instead of additional methods - the methods are static anyway.

Also, it would be good to add a comment why `SYSTEMROOT` needs to be copied to `env`.

py/examples/tour.py
```diff
@@ -29,24 +29,45 @@ def __init__(self, filename: str, title: str, description: str, source: str):
...
         else:
-            self.process = await asyncio.create_subprocess_exec(
-                sys.executable, os.path.join(example_dir, self.filename)
-            )
+            self.process = subprocess.Popen([sys.executable, os.path.join(example_dir, self.filename)])
```
lo5, line 41: Does the env for this `Popen()` call require `SYSTEMROOT`?

lo5 (changes requested):
Thanks for the PR @VijithaEkanayake!

py/examples/tour.py
VijithaEkanayake: If we don't call wait(). Zombie process entries remain in the process table until the user terminates the parent process in MasOS.
Screenshot 2020-12-22 at 08 11 59

py/examples/tour.py
VijithaEkanayake: Probably needed I think, Will check the behavior for scripts and update as necessary.

lo5 (approved):
Thanks @VijithaEkanayake !

PR #430, opened 2020-12-21
fix: force dropdown callout to be placed at bottom

### LINKED ISSUE #149: Dropdown list weird when many items
**Q SDK Version, OS**
0.5.0 (I think) q8s.h2o.ai - see Telco Churn App v 0.1.0

**Actual behavior**
drop down list looks weird and moves to the left when there are too many items for a certain resolution

This is on my 14" Mac 
image

This is the same page but I moved to my additional monitor which is larger / different resolution (I guess)
image

When I had 100 items in my drop down list I saw the same behavior on both screens

**Expected behavior**
Don't move location when there are lots of items

**Steps To Reproduce**
I have not figured out the exact repo steps in regards to how many items is "too many" for a screen

mturoci:
Closes #149

### DIFF
--- a/githooks/pre-commit.js
+++ b/githooks/pre-commit.js
@@ -37,7 +37,7 @@ const
 
 if (files.uiTs.length) {
   runLinter('eslint', 'ui/node_modules/eslint/bin/eslint.js --resolve-plugins-relative-to ui -c ui/.eslintrc.js', files.uiTs)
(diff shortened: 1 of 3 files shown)


lo5:
I'm not sure if this is the best strategy. Why not let the dropdown follow its default behavior? Why force a hint? 
If the dropdown is trying to fit too many items in there, maybe using a dropdown for those use cases is poor UX.

mturoci:
Comments addressed in #149.

PR #431, opened 2020-12-21
Edited index.js

tomott12345:
Fixed grammar issue with 'great good' and spelled out RPi as Raspberry Pi (RPi) Zero.

### DIFF
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -57,7 +57,7 @@ const features = [
     icon: 'fas fa-terminal',
     description: (
       <>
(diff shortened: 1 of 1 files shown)


mtanco:
The spelling out of Raspberry Pi is nice, but I'm for keeping the Haskel joke :) 
image

lo5:
Yes, it was a hat tip to that book. This change sort of takes the punch out of the punchline :-)
Plus, the device is commonly abbreviated that way (even "Pi Zero").

tomott12345:
So this joke went right over my head. I don't know how many people will get it, so perhaps we can remove this from the front page and put it somewhere else?

Issue #391, opened 2020-12-15
Running two apps side-by-side has port interference


tomkraljevic:
**Q SDK Version, OS**

- waved version 0.10.0
- h2o-wave version 0.10.0
- python version 3.6, 3.7
- macos catalina

**Actual behavior**

App 1 output:

```
/Users/tomk/anaconda/envs/license37/bin/python -m h2o_wave run --no-reload generator_app.py
INFO:     Started server process [59358]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
```

App 2 output:

```
/Users/tomk/anaconda/envs/license37/bin/python -m h2o_wave run --no-reload claiming_app.py
INFO:     Started server process [59359]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8001 (Press CTRL+C to quit)
```

waved output:

```
tomk@tomkmbp wave-0.10.0-darwin-amd64 % ./waved
2020/12/15 01:10:12 # 
2020/12/15 01:10:12 # ┌─────────────────────────┐
2020/12/15 01:10:12 # │  ┌    ┌ ┌──┐ ┌  ┌ ┌──┐  │ H2O Wave
2020/12/15 01:10:12 # │  │ ┌──┘ │──│ │  │ └┐    │ 0.10.0 20201129141903
2020/12/15 01:10:12 # │  └─┘    ┘  ┘ └──┘  └─┘  │ © 2020 H2O.ai, Inc.
2020/12/15 01:10:12 # └─────────────────────────┘
2020/12/15 01:10:12 # 
2020/12/15 01:10:12 # {"address":":10101","t":"listen","webroot":"/Users/tomk/0xdata/wave-0.10.0-darwin-amd64/www"}
2020/12/15 01:10:28 # {"host":"http://127.0.0.1:8000","route":"/generator_app","t":"app_add"}
2020/12/15 01:10:33 # {"host":"http://127.0.0.1:8000","route":"/claiming_app","t":"app_add"}
```

Notice that the two app_add are both for port 8000.  But the second app's output says it should be on port 8001.
And if you try to go to both routes with a browser, they both show app 1.

**Expected behavior**

For the two apps to not be wired together.

**Steps To Reproduce**

1.  Start waved.
2.  Start app 1
3.  Start app 2

tomkraljevic:
[...]

PR #436, opened 2020-12-22
Infer correct app address from ENV

VijithaEkanayake:
Wave CLI dynamically generates host:port combinations to be paased to uvicorn and update the ENV with that address(H2O_WAVE_APP_ADDRESS). However wave server doesn't inter the updated value when registering an application. This PR fixes that issue.  Fixes: https://github.com/h2oai/wave/issues/391.


lo5 (approved):
Thanks @VijithaEkanayake!

PR #438, opened 2020-12-22
Update routing.md

sergiomora03:
just put `await` in code. The app doesn't work becouse that.


lo5 (approved):
Thanks for catching this @sergiomora03!

Issue #31, opened 2020-08-04
Hard-code synthetic data in examples


lo5:
The use of `Fake*` data generators make the plotting examples harder to decipher. 
Simplify by hard-coding data in the body of the examples.

PR #443, opened 2020-12-23
fix(examples): Remove Faker and Synth from examples

VijithaEkanayake:
The use of Fake data generators make the examples harder to understand. Simplify by hard-coding data in the body of the examples. Relates: https://github.com/h2oai/wave/issues/31


mturoci:
Closing in favor of #1337

Issue #423, opened 2020-12-20
Add a master table of contents page for all examples


lo5:
The current gallery page displays only thumbnails and titles without descriptions. 

It would be good to have a parallel ToC that lists all examples along with their descriptions.

dulajra:
@lo5 Did you mean to keep the Gallery page as it is and add a new TOC page with titles and short descriptions (only text - no images)? I hope we can reuse the same description already available in each component example here also.

lo5:
@dulajra Exactly! This line: https://github.com/h2oai/wave/blob/master/py/sync_examples.py#L116 generates the ToC, but might need additional implementation/configuration in Docusaurus. Needs the `useBaseUrl` patch as well, similar to `make_gallery_thumbnail()`.

Issue #413, opened 2020-12-18
h2oai.github.io/wave/docs/examples dead links


0100101001010000:
h2oai.github.io/wave/docs/examples links are dead, for example if you click on Hello World! the link it targets is:

    https://h2oai.github.io/wave/docs/examples/examples/hello-world

however the actual link seems to be:

https://h2oai.github.io/wave/docs/examples/hello-world

seems like you've got an extra _examples_ in the href that shouldn't be there.

Minor UX issue but the more people look at your product the more important it'll get :)

shaz13:
The link - [https://h2oai.github.io/wave/docs/examples/hello-world/](https://h2oai.github.io/wave/docs/examples/hello-world/) works fine for me.

mturoci:
Thanks for reporting the issue @0100101001010000!

PR #444, opened 2020-12-23
Add TOC for Wave examples

dulajra:
For https://github.com/h2oai/wave/issues/423

### DIFF
--- a/py/sync_examples.py
+++ b/py/sync_examples.py
@@ -75,7 +75,7 @@ def load_example(filename: str) -> Example:
 
 def make_toc(examples: List[Example]):
     return '''---
-title: Contents
+title: Table of Content
 slug: /examples
 ---
 
@@ -89,7 +89,7 @@ def make_gallery_thumbnail(e: Example):
 def make_gallery(examples: List[Example]):
     return '''---
 title: Gallery
-slug: /examples
+slug: /examples/gallery
 ---
 
 import useBaseUrl from '@docusaurus/useBaseUrl';
@@ -112,9 +112,10 @@ def main():
 
     example_items = [dict(slug=e.slug) for e in examples]
     example_items.insert(0, dict(slug='index'))
+    example_items.insert(1, dict(slug='gallery'))
     write_file(os.path.join(website_dir, 'examples.js'), f'module.exports={json.dumps(example_items)}')
-    # write_file(os.path.join(example_md_dir, 'index.md'), make_toc(examples))
-    write_file(os.path.join(example_md_dir, 'index.md'), make_gallery(examples))
+    write_file(os.path.join(example_md_dir, 'index.md'), make_toc(examples))
+    write_file(os.path.join(example_md_dir, 'gallery.md'), make_gallery(examples))
 
 
 if __name__ == '__main__':
--- a/py/sync_examples.py
+++ b/py/sync_examples.py
@@ -75,7 +75,7 @@ def load_example(filename: str) -> Example:
 
 def make_toc(examples: List[Example]):
     return '''---
-title: Table of Content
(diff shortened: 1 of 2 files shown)


website/docusaurus.config.js
```diff
@@ -26,7 +26,7 @@ module.exports = {
         src: 'img/logo.svg',
       },
       items: [
-        { to: 'docs/examples', label: 'Gallery', position: 'left' },
+        { to: 'docs/examples', label: 'Examples', position: 'left' },
```
lo5, line 29: Let's preserve the current Gallery page at /examples.

py/sync_examples.py
```diff
@@ -112,9 +112,10 @@ def main():
 
     example_items = [dict(slug=e.slug) for e in examples]
     example_items.insert(0, dict(slug='index'))
+    example_items.insert(1, dict(slug='gallery'))
     write_file(os.path.join(website_dir, 'examples.js'), f'module.exports={json.dumps(example_items)}')
-    # write_file(os.path.join(example_md_dir, 'index.md'), make_toc(examples))
-    write_file(os.path.join(example_md_dir, 'index.md'), make_gallery(examples))
+    write_file(os.path.join(example_md_dir, 'index.md'), make_toc(examples))
+    write_file(os.path.join(example_md_dir, 'gallery.md'), make_gallery(examples))
```
lo5, line 118: It would be good to preserve the gallery page as-is, and add a new page for the ToC.

py/sync_examples.py
```diff
@@ -112,9 +112,10 @@ def main():
 
     example_items = [dict(slug=e.slug) for e in examples]
     example_items.insert(0, dict(slug='index'))
+    example_items.insert(1, dict(slug='gallery'))
     write_file(os.path.join(website_dir, 'examples.js'), f'module.exports={json.dumps(example_items)}')
-    # write_file(os.path.join(example_md_dir, 'index.md'), make_toc(examples))
-    write_file(os.path.join(example_md_dir, 'index.md'), make_gallery(examples))
+    write_file(os.path.join(example_md_dir, 'index.md'), make_toc(examples))
```
lo5, line 117: The `make_toc()` function needs to use `useBaseUrl()` like the `make_gallery_thumbnail()` function. See #413

py/sync_examples.py
```diff
@@ -75,7 +75,7 @@ def load_example(filename: str) -> Example:
 
 def make_toc(examples: List[Example]):
     return '''---
-title: Contents
+title: Table of Content
```
lo5, line 78:
```suggestion
title: All Examples
```

lo5 (changes requested):
Thanks @dulajra - minor comments.

website/docusaurus.config.js
dulajra: Done

py/sync_examples.py
dulajra: Done

py/sync_examples.py
dulajra: @lo5 This was fixed by just using the relative path as `e.slug` instead of `examples/e.slug`. I think this should work without `useBaseUrl()` function. I tested this on my local machine too. Isn't it?

dulajra: Now I get it :) The relative URL is not when working if the user has landed to `http://localhost:3000/wave/docs/examples/all/` instead of `http://localhost:3000/wave/docs/examples/all` and click on a link. But this still works with `useBaseUrl()`

dulajra: Fixed

lo5 (approved):
Thanks @dulajra!

Issue #399, opened 2020-12-17
Set up CI using Github actions


lo5:
**Goal**
Canonical CI set up and configuration using Github actions; automate steps outlined in https://github.com/h2oai/wave/blob/master/DEV.md#cut-a-release

mturoci:
@lo5 wdyt about automated changelog off commits? Proposal:

* Squash each PR with appropriate changlelog message / introduce some format for PR title.
* Generate the changelog off `git log`

lo5:
I'd love to do that, but we can only make it work if:
- Every commit message is end-user (app author) facing.
- Commits are not too granular (results in a verbose change log).

mturoci:
Wouldn't this be resolved by squashing? You can pick a custom commit message for this it if I recall correctly.

lo5:
I change almost every message when I squash. Even then, the commit log is far too verbose for *users* of wave and needs curation/pruning: https://github.com/h2oai/wave/commits/master

mturoci:
I see, will investigate other possibilities / flows that might be interesting for us.

dulajra:
Squashing will make the log more meaningful. But still, I also think it will contain messages that a user doesn't expect in a changelog. One suggestion I have here is,

- Maintain a `develop` branch and merge all the feature/bug fix/typos PR to `develop` daily.
- Merge `develop` branch to `main` squashing unwanted messages when we are ready for a release. 
- Generate changelog from `main` branch commit history

But yeah there will be some additional work here.

PR #448, opened 2020-12-24
Github actions pipeline to cut a release in Wave

dulajra:
Closes https://github.com/h2oai/wave/issues/399

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -86,7 +86,14 @@ jobs:
           asset_name: "wave-${{ env.VERSION }}-windows-amd64.tar.gz"
           asset_content_type: application/gzip
 
-      - name: Publish distribution 📦 to PyPI
+      - name: Publish to PyPI
         uses: pypa/gh-action-pypi-publish@master
         with:
           password: ${{ secrets.PYPI_API_TOKEN }}
+
+      - name: Publish to Conda
+        uses: fcakyon/conda-publish-action@v1.3
+        with:
+          subdir: 'conda'
+          anacondatoken: ${{ secrets.ANACONDA_TOKEN }}
+          platforms: 'win osx linux'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,7 +20,9 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Build
-        run: make all
+        run: |
+          cd ui
+          make setup
 
       - name: Unit Test
         run: make test-ui-ci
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -97,7 +97,7 @@ release-os:
 	cd build && tar -czf $(REL).tar.gz  --exclude='*.state'  --exclude='__pycache__' $(REL)
 
 build-website: ## Build website
-	cd website && npm run build
+	cd website && npm install && npm run build
```
lo5, line 100:
```suggestion
	cd website && npm ci && npm run build
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,91 @@
...
+
+      - name: Upload Windows Release Asset
+        uses: actions/upload-release-asset@v1
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: "./build/wave-${{ env.VERSION }}-windows-amd64.tar.gz"
+          asset_name: "wave-${{ env.VERSION }}-windows-amd64.tar.gz"
+          asset_content_type: application/gzip
+
+# TODO: Enable this before merge
```
lo5, line 89: It would be good to remove the `publish-pypi` task and use [this procedure](https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/) instead.

lo5 (changes requested):
Thanks for putting this together @dulajra!

The publish-PyPI task needs an overhaul (see link). I can supply API tokens.

.github/workflows/release.yml
dulajra: Done!

.github/workflows/release.yml
```diff
@@ -0,0 +1,85 @@
...
+
+jobs:
+  ci-test:
+    name: CI Test
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build
+        run: make all
```
mturoci, line 23: Is this really necessary since we run just ui unit tests here? What do you think about just doing `cd ui && npm run ci` ?

mturoci (commented):
Good job @dulajra! Just one quick question from me.

.github/workflows/release.yml
dulajra: @mturoci Yes that's enough. I will change it. Thanks!

dulajra:
Publish to Pypi was tested with Test Pypi: https://test.pypi.org/manage/project/h2o-wave/releases/

dulajra:
`0.10.0` release was created using this pipeline here - https://github.com/h2oai/wave-ci-dummy/releases
Publish to Pypi was tested with Test Pypi

lo5 (approved):
Thanks @dulajra!

PR #461, opened 2020-12-30
Update file name to run in doc tutorial-monitor

angelosaleh:
system_stats.py is documented in https://h2oai.github.io/wave/docs/tutorial-monitor when it is time to run the program; However, system_monitor.py is what is created during the example, thus system_monitor.py should be used to run the example.


lo5:
Thanks for catching this @angelosaleh. I've modified the source .md file in master: https://github.com/h2oai/wave/commit/fc82547b0d90b282fa050460c52f4730cfca38eb

Issue #445, opened 2020-12-23
Trailing slash breaks routing in docs website


mtanco:
### Actual behavior
When a URL has a trailing slash, any internal links that are not explicit to the file fail. 

### Expected behavior
Page to render instead of 404

### Steps To Reproduce
Click on `Gallery` in the second to last section of the below page

#### Works
`https://h2oai.github.io/wave/docs/tutorial-monitor` goes to `https://h2oai.github.io/wave/docs/examples`

#### Doesn't Work
`https://h2oai.github.io/wave/docs/tutorial-monitor/` goes to `https://h2oai.github.io/wave/docs/tutorial-monitor/examples`

## Proposed Solution (needs testing)
This seems to be a docusaurus problem and they suggest, for now, using .md extensions and relative paths. 
We'll need to do a bulk-replace of all our local links `[Gallery](examples)` to use extensions and relative paths

dulajra:
@mtanco Even though we use the `.md` extension in URL if the user lands with a trailing `/` (i.e. `examples/index.md/` ) then still the issue remains.

Here also using `useBaseUrl` function instead of markdown hyperlink will fix the problem.

`<a href={useBaseUrl('docs/examples')}>Gallery</a>`

@lo5 WDYT?

lo5:
@dulajra It would be good to figure out a way to fix this using plain markdown, without `useBaseUrl` everywhere. This is a problem with docusaurus: https://github.com/facebook/docusaurus/issues/3372

MarkFarmiloe:
In docusaurus/packages/docusaurus/src/client/normalizeLocation.ts
any trailing /index.html is knocked off the path.
If you knock off any trailing / as well with
pathname = pathname.replace(/\/$/, '');
on line 23, this seems to fix this bug.

But I don't know whether this would break anything else.
But it does seem to me that if you want to normalize, then allowing both /foo/bar/ and /foo/bar doesn't seem to be a very good job of normalizing to me.

PR #469, opened 2021-01-04
fix api/core#Expando with api/core/#Expando

sergiomora03:
Just add backslash to the right link.


sergiomora03:
Hi @lo5, I just add backslash. Please [check it](https://github.com/h2oai/wave/pull/469/files) to merge.

lo5:
@sergiomora03 Thanks for the PR. The bug is caused by #445 and affects the entire docs website, so the local fix in this PR does not address the general case.

PR #471, opened 2021-01-04
fix: Make stats cards work correctly for responsive layouts.

lo5:


lo5:
@mturoci 

I reverted the changes to the red dashboard.
Per the layout spec, this is what is supposed to happen:
1, 2 should be 50% of available width. It's not (2 is wider than 1).
3, 4 should be 50% of 385px (i.e. 3+4 should be 385px).

lo5:
More fixes (see notes in commit above):



1 - 50% width behavior improved
2, 3, 4, 5 - SVG rendering computes `clientWidth`/`clientHeight` incorrectly, causing (why?)
6 - `flex-grow:1` still not perfect (why?)

---



1 - same problem: `flex-grow:1` still not perfect (why?)

---




1, 1, 2 - more examples of incorrect `clientWidth`/`clientHeight` used by SVG rendering.

@mturoci - in addition to the above, the grey dashboard would be a good layout to debug `flex-grow:1` issues.

The major reason SVG + flex behavior is better than before is that the rendering is now using the parent `div`'s size to compute bounds: https://github.com/philipwalton/flexbugs/issues/184

lo5:
Since the flex/responsive layout no longer applies a padding/margin to individual cards, each card now has to apply its own padding. 

The major reason for doing is that a couple of the cards that were rendering their visualizations flush with the border got broken (e.g. `tall_series_stat_card` - notice how the screenshot in the readme has its area plot flush with the border).

Additional tasks:
- Make grid layout not apply any padding (see `css.slot` in `layout.tsx`).
- Make each `*_card` apply its own `grid.gap` padding (the ones in my last commit already do this - need to check the rest).
- Verify that all examples are rendering properly.

mturoci:
Thanks for the imagery, really helped.

> Per the layout spec, this is what is supposed to happen:
1, 2 should be 50% of available width. It's not (2 is wider than 1).

Why should they be 50% width? They should just both have `flex-grow: 1`, but that does not mean equal width, but rather equal distribution of remaining space. They don't have equally wide contents.

>2, 3, 4, 5 - SVG rendering computes clientWidth/clientHeight incorrectly, causing (why?)

Not 100% correct, try adding `debugger` to the end of `renderViz` function and reload page with devtools open. You will see that the initial flex layout is rendered correctly. This is the point that `renderViz` uses for width/height calculations - these are correct. The problem is that after flushing changes to the DOM, cards resize for some reason - I couldn't figure out why yet. D3 does not get rerendered after the resize, thus looking ugly.

> 6 - flex-grow:1 still not perfect (why?)

The biggest question of all. For some reason, the card still takes extra few px (which are part of either plot container or card container), even if you set `min-height` for plot div container, the space is assigned to card itself, thus I suspect the problem is somewhere higher.
Note: I tried my previous (non-flex-grow) approach and still got the extra space - feels like some weird SVG issue.

>The major reason SVG + flex behavior is better than before is that the rendering is now using the parent div's size to compute bounds

I am not sure I understand the benefit of this approach. The previous one just filled entire parent (width / height set to `100%`) and took its own size. The advantage was that you could control the total sizing from outside D3 component.

lo5:
> Why should they be 50% width? 

Per *our* `ui.layout` spec, they should be 50%. I was not referring to flexbox.

> I am not sure I understand the benefit of this approach. 

1. It fixed the problem shown here: https://github.com/h2oai/wave/pull/471#issuecomment-754115901.
2. No `calc()` / `%` hacks with hard-coded title heights. Now we can add more elements, like captions, icons, etc. without having to go fix the hard-coding over and over again, and handle edge cases like browser-zoom.

PR #474, opened 2021-01-07
Flex layout fixes

### LINKED ISSUE #473: Layout/SVG rendering issues in flex layout
This is a catch-all issue for layout/rendering issues holding up release 0.11.

**Grey Dashboard**: Footer renders on top of bottom row:

mturoci:
* Use absolute positioning (to avoid normal document flow) to avoid weird flexbox behavior with inline SVG - probably due to scalability capabilities.
* Fix some flexbox browser quirks - flex child didn't take full height on all major browsers - tested with Chrome, Safari, Firefox, Edge.
* Recompute svgs on `resize` event.
* Fixed grey dashboard as commented in issue.

Closes #473

### DIFF
--- a/py/demo/dashboard_grey.py
+++ b/py/demo/dashboard_grey.py
@@ -15,7 +15,7 @@ async def show_grey_dashboard(q: Q):
                     ui.zone('top', direction=ui.ZoneDirection.ROW, size='25%'),
                     ui.zone('middle', direction=ui.ZoneDirection.ROW, size='25%'),
                     ui.zone('middle2', direction=ui.ZoneDirection.ROW, size='25%'),
-                    ui.zone('bottom', direction=ui.ZoneDirection.ROW, size='25%'),
+                    ui.zone('bottom', direction=ui.ZoneDirection.ROW, size='20%'),
                 ]),
                 ui.zone('footer', size='0'),
             ]
--- a/py/demo/dashboard_grey.py
+++ b/py/demo/dashboard_grey.py
@@ -15,7 +15,7 @@ async def show_grey_dashboard(q: Q):
                     ui.zone('top', direction=ui.ZoneDirection.ROW, size='25%'),
                     ui.zone('middle', direction=ui.ZoneDirection.ROW, size='25%'),
                     ui.zone('middle2', direction=ui.ZoneDirection.ROW, size='25%'),
-                    ui.zone('bottom', direction=ui.ZoneDirection.ROW, size='20%'),
+                    ui.zone('bottom', direction=ui.ZoneDirection.ROW, size='25%'),
                 ]),
                 ui.zone('footer', size='0'),
             ]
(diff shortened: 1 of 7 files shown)


mturoci:
Other bugs worth noting:
image
* Plot has overflowing axes, does not take full width/height of card.

py/demo/dashboard_grey.py
```diff
@@ -15,7 +15,7 @@ async def show_grey_dashboard(q: Q):
                     ui.zone('top', direction=ui.ZoneDirection.ROW, size='25%'),
                     ui.zone('middle', direction=ui.ZoneDirection.ROW, size='25%'),
                     ui.zone('middle2', direction=ui.ZoneDirection.ROW, size='25%'),
-                    ui.zone('bottom', direction=ui.ZoneDirection.ROW, size='25%'),
+                    ui.zone('bottom', direction=ui.ZoneDirection.ROW, size='20%'),
```
lo5: This does not address the problem why the footer overlaps the body, regardless of the height of the body. For example, the problem persists even if you change the body size to `800px`.

lo5: So we have to revert this to 25%, and instead investigate what's wrong with the layout system.

ui/src/parts/microbars.tsx
```diff
@@ -58,5 +59,5 @@ export const MicroBars = ({ value, category = 'x', color, data, zeroValue }: Pro
   }, [])
   React.useLayoutEffect(renderViz, [value, category, color, data, zeroValue])
 
-  return <div ref={ref} style={{ width: '100%', height: '100%' }}>{content}</div>
+  return <div ref={ref} style={{ width: '100%', height: '100%', display: 'flex', position: 'relative', flexGrow: 1 }}>{content}</div>
```
lo5: The `display: 'flex'` seems to be unnecessary here, since the svg child is absolutely positioned.

ui/src/parts/microarea.tsx
```diff
@@ -82,5 +83,5 @@ export const MicroArea = ({ value, color, data, zeroValue, curve }: Props) => {
   }, [])
   React.useLayoutEffect(renderViz, [value, color, data, zeroValue, curve])
 
-  return <div ref={ref} style={{ width: '100%', height: '100%' }}>{content}</div>
+  return <div ref={ref} style={{ width: '100%', height: '100%', display: 'flex', position: 'relative', flexGrow: 1 }}>{content}</div>
```
lo5: The `display: 'flex'` seems to be unnecessary here, since the svg child is absolutely positioned.

ui/src/plot.tsx
```diff
@@ -735,6 +735,11 @@ const
       flexGrow: 1,
       display: 'flex',
     },
+    plotContainer: {
+      $nest: {
+        '>div': { textAlign: 'center' }
```
lo5: This does not fix the problem, which is that the G2 vis for some reason does not use all available width. This change simply centers the plot.

lo5 (approved):
Thanks @mturoci - The fix for SVG rendering looks good - thanks for figuring this out without any ugly hacks!
I've reverted a couple of changes here (see comments why), and merging into master - let's continue to investigate.

PR #480, opened 2021-01-12
fix/flex plot

mturoci:
Additional layout fixes

* Managed to fix g2 plots kind of - they take all available free space also within flex layout - to make it work with form cards as well, I had to display form items as flexbox children with col direction for vertical alignment. There is one catch though - the margins don’t collapse within flexbox (https://stackoverflow.com/questions/43882869/margin-collapsing-in-flexbox) which in combination with default margins for p tags for examples caused large gaps between text form elements. I managed to work around it, but it is not ideal.
* Discovered scrolling bug on g2 plots - seems like scroll event bubbles outside the component and interferes with native scroll event handler. (video 1)
* Discovered overflow bugs on g2 plots (img1, img2) - didn’t dig deeper for now
* Added vega resizing
* Centered vega in forms and grid layouts
* Checked tour and dashboards and it seems to be fine
image (2)

image (1)

https://user-images.githubusercontent.com/64769322/104325706-39470180-54e9-11eb-85f8-dc83ef1f1646.mov

### DIFF
--- a/py/demo/dashboard_purple.py
+++ b/py/demo/dashboard_purple.py
@@ -272,7 +272,6 @@ async def show_purple_dashboard(q: Q):
                         ('In-store', next(sales_days_2), next(sales_amounts_2)) for i in range(15)],
                     pack=True
                 ),
-                height='210px',
             )
         ],
     )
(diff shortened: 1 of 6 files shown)


lo5 (approved):
Thanks @mturoci!

PR #449, opened 2020-12-24
feat: support tags/hashtags on examples to improve discovery #422

srini-x:
Closes #422 

1. Added a new page `Tags` between `Gallery` and `Hello World!`
2. The `tags` are subheadings on this page, sorted alphabetically. Under each `tag`, links to the examples with those tags are available.
3. Added a `Tags` field under each example with tags. These tags are links that link back to the Tags page mentioned above.
4. Added `keywords` metadata in the header of the each generated `.md` file. Docusaurus [markdown headers](https://v2.docusaurus.io/docs/markdown-features#markdown-headers) provide only `keywords` under Docs and `tags` under Blog.
5. In the python example files, tags can be provided as the last line of description.
6. Tags line can not have `.` in it. Tags must be `,` separated. A tag can not contain spaces.

ToDo:
1. Comment the code
2. Add tags to all remaining examples.

@lo5, If you like this implementation, I will go ahead and complete the ToDos listed above.

example:

`buttons.py`

```py
# Form / Buttons
# Use the `ui.buttons()` function to group related buttons.
# form, buttons
# ---
from h2o_wave import main, app, Q, ui
```

`buttons.md`

```md
---
title: Form / Buttons
keywords:
  - form
  - buttons
---
```

wave_docs_examples_tags

wave_docs_examples_buttons_tags

PR #489, opened 2021-01-13
feat: add tag parsing from the descriptions for the examples in tour.…

### LINKED ISSUE #488: Add tag parsing from the descriptions for the examples in tour.py
We can add twitter style hash-tags to example descriptions to make it easy for docs.
we need to implement the same functionality in `tour.py` to remove the hash-tags from appearing in the raw form in the wave tour.

srini-x:
Closes #488
Similar implementation as #449

Added `parse_tags` from
https://github.com/h2oai/wave/blob/1bc6e309cf37fadf827b3d275fb858096252c206/py/sync_examples.py#L104

### DIFF
--- a/py/examples/tour.py
+++ b/py/examples/tour.py
@@ -1,9 +1,10 @@
 import collections
 import os
 import os.path
+import re
 import subprocess
 import sys
-from typing import List, Optional, Dict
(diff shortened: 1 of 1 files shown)


lo5 (approved):
Thanks, @srini-x!

PR #490, opened 2021-01-14
Update Conda package name

dulajra:
Explicitly set Conda package name as `h2o-wave`


dulajra:
Addressed with https://github.com/h2oai/wave/pull/491

PR #492, opened 2021-01-14
Rdriver

ashrith:
First pull Rq for rdriver. You can build, install and use the h2owave library from r.


r/R/examples/stat_large.R
```diff
@@ -0,0 +1,36 @@
+#source("../core.R")
+#source("../ui.R")
```
lo5, line 2: We can probably remove these imports from all the examples.

r/R/examples/stat_large.R
```diff
@@ -0,0 +1,36 @@
...
+library(h2owave)
+library(tokenizers)
+base_url <- "https://programminghistorian.org/assets/basic-text-processing-in-r"
+url <- sprintf("%s/sotu_text/236.txt", base_url)
+text <- paste(readLines(url), collapse = "\n")
+sentences <- tokenize_sentences(text)
+captions <- sentences[[1]][sample(seq(1,length(sentences[[1]])),1)]
```
lo5, line 9: It would be good to hard-code the data right here (or use one of R's built-in datasets) - makes the example easier to read. In general, the shorter the example, the better.

r/R/examples/stat_large.R
```diff
@@ -0,0 +1,36 @@
...
+while(TRUE)
+{
+  Sys.sleep(3)
+  card_pc <- as.integer(runif(1,1,100))
+  card_value <- as.character(sample(mtcars[['mpg']],1))
+  captions <- sentences[[1]][sample(seq(1,length(sentences[[1]])),1)]
+  test_page[[1]]$value$data$foo <- card_value
+  test_page[[1]]$value$data$bar <- card_pc/100
+  test_page[[1]]$value$caption <- captions
+  page.save(test_page,"/page_demo")
```
lo5, line 35: Having to call `page.save(page, url)` is repetitive. 

It would be good to do something like:

```
my_page = get page '/foo' from site
...
...
save my_page
```

The above pseudo-code might look like:
```
my_page = get_page(site, '/foo')
...
save_page(my_page)
```

or

```
my_page = site$get_page('/foo')
...
my_page$save()
```

I don't know what's idiomatic in R.

r/R/examples/stat_large.R
```diff
@@ -0,0 +1,36 @@
...
+text <- paste(readLines(url), collapse = "\n")
+sentences <- tokenize_sentences(text)
+captions <- sentences[[1]][sample(seq(1,length(sentences[[1]])),1)]
+
+card_title <- sample(row.names(mtcars),1)
+card_value <- as.character(sample(mtcars[['mpg']],1))
+card_pc <- as.integer(runif(1,1,100))
+
+test_page <- page("/page_demo")
+test_page <- page.add(test_page,"/page_demo","test",ui_large_stat_card(box="1 1 2 2"
```
lo5, line 16: We need to provide:
1. A canonical way to grab a page, given a URL.
2. A canonical way to add a card to a page.

In the above lines, having `page.add()` return a `test_page` is confusing to me.

In Python, `page.add()` is the mechanism to add a card to a page, and returns a reference to the card. In case the reference to the card is not required, the Python API provides a `page['card_name'] = ...` mechanism as well.

r/R/examples/stat_large.R
```diff
@@ -0,0 +1,36 @@
...
+                                                                       ,title=card_title
+                                                                       ,value='=${{intl foo minimum_fraction_digits=2 maximum_fraction_digits=2}}'
+                                                                       ,aux_value='={{intl bar style="percent" minimum_fraction_digits=2 maximum_fraction_digits=2}}'
+                                                                       ,data=list(foo=card_value,bar=card_pc/100)
```
lo5, line 20: We need the equivalent of Python's `data()` function here. That function provides hints to the server about how to allocate data for this card (size, etc.).

lo5 (changes requested):
Thanks @ashrith.

I would suggest that we work on `stat_large.R` and a couple of other examples and work on 
 the API until the examples look near-perfect, and then move on to the next steps. 

Further, it would be good to .gitignore roxygen-generated files from the PR (and repo) - we can generate those just in time during the CI build.

r/R/examples/stat_large.R
ashrith: removed from examples

r/R/examples/stat_large.R
ashrith: Updated with a standard text list.

r/R/examples/stat_large.R
ashrith:
```
page <- Site("/demo")

columns = 12
rows = 10

for(column in 1:columns){
        for(row in 1:rows){
               box = paste0(column," ",row," 1 1") 
        page$add.card(paste0("card_",column,"_",row),
                              ui_markdown_card(box=box,title=box,content=''))
        }}  
page$page.save()
```

The Rdriver now is able to access the page canonically, and is able to `save` the page by `page$page.save()` (the `page` is the canonical object)

r/R/examples/stat_large.R
ashrith: The rdriver now is able to allocate a canonical object for a page by
`page <- Site("demo")`
similarly, cards can be added as follows
```
page$add.card(paste0("card_",column,"_",row),
                              ui_markdown_card(box=box,title=box,content=''))
```
you can drop the page as follows
`page$page.drop()`
or save the page (send to the server)
`page$page.save()`

r/R/examples/stat_large.R
```diff
@@ -0,0 +1,35 @@
+#library(h2owave)
+
+
+captions <- list("Our Constitution begins with those three simple words."
```
lo5, line 4: Let's see if we can make this example a bit more understandable (and improve over the Python example).

**Improvement 1**

This example is odd to read: 
- The captions are unrelated to mtcars.
- The percentage displayed is a random value.
- The primary value originates from mtcars.

To the reader, it takes more work to read through the example and distill it down to its essence. To simplify this, either use random data for all three attributes, or use mtcars for all three attributes.

**Improvement 2**

Depending on what information is being displayed, change `foo` and `bar` to the names of the variables (e.g. 'mpg', 'efficiency', 'readout') so that it's clear which variable is being mapped to which part of the widget.

Also rename `card_value`, `card_pc`, etc. to reflect the actual data being displayed ('mpg', 'efficiency', etc.).

**Improvement 3**

The indentation looks wrong (e.g. the `$add.card(` block is indented to col 72)

**Improvement 4**

Rename the card `test` to something that reflects the data being displayed (e.g. `mtcars_stats`) so that the `$page$set("test", ...` is more understandable.

lo5 (changes requested):
Thanks @ashrith - `stat_large.R` looks better now. Let's continue to work on refining it before we make similar changes to the other examples.

lo5:
@ashrith:

can you please make a dockerfile in the /r directory that has includes system-level deps?

`r-base` isn't enough. seems like the transitive dependencies require curl, etc - they don't build and consequently `devtools` fails:

```
    $ docker run -it r-base bash
    $ git clone --branch rdriver --single-branch https://github.com/h2oai/wave
    ...
    ...
    ...
    ERROR: dependencies ‘usethis’, ‘httr’, ‘roxygen2’, ‘rversions’ are not available for package ‘devtools’
    * removing ‘/root/R_LIBS/devtools’
```

lo5:
also, summarizing changes - we need Python-equivalent:
- `site = Site('http://example.com')` (grabbing a reference to a site)
- `page = site['/foo/bar']` (grabbing a reference to a page in the site)
- `site.upload(...)`, `site.unload(...)`, `site.download(...)` (site manipulation)

post-implementation, the following should be possible:

manipulate multiple pages from the same site:
```
s <- .Site$new('http://foo.com')
p1 <- s$get('/stats/cricket') # mints a new page from the site s
p2 <- s$get('/stats/hockey') # mints another page from the site s
p1$set(....)
p2$set(....)
```

manipulate multiple pages on multiple sites:
```
s1 <- .Site$new('http://foo.com')
p1 <- s1$get('/stats/cricket')
s2 <- .Site.new('http://bar.com')
p2 <- s2$get('/stats/hockey')
p1$set(....)
p2$set(....)
```

manipulate multiple sites:
```
s1 <- .Site$new('http://foo.com')
s1$upload('cricket_scores.csv')
s2 <- .Site.new('http://bar.com')
s2$upload('hockey_scores.csv')
```

lo5:
more notes:

the python driver exports both `site` (noun) and `Site` (verb / constructor). 
difference is: `Site()` mints a new instance of a site, and `site` is shorthand for a `Site()` that already points to `$H2O_WAVE_ADDRESS` using `$H2O_WAVE_ACCESS_KEY_ID`/`$H2O_WAVE_ACCESS_KEY_SECRET` for auth.

so for now, it's fine to not expose the R6 `SIte()` in the API, but structure the API in a way that let's us demonstrate the 3 examples above in a future release.

in other words: only export a pre-created `site`, and the `$get` on the `site` to get an instance of the page, and we can do the `Site()` part later.

ashrith:
> @ashrith:
> 
> can you please make a dockerfile in the /r directory that has includes system-level deps?
> 
> `r-base` isn't enough. seems like the transitive dependencies require curl, etc - they don't build and consequently `devtools` fails:
> 
> ```
>     $ docker run -it r-base bash
>     $ git clone --branch rdriver --single-branch https://github.com/h2oai/wave
>     ...
>     ...
>     ...
>     ERROR: dependencies ‘usethis’, ‘httr’, ‘roxygen2’, ‘rversions’ are not available for package ‘devtools’
>     * removing ‘/root/R_LIBS/devtools’
> ```

Added in this commit. There is a Dockerfile now. 
https://github.com/h2oai/wave/commit/234ebb11f3538c67ab1bd97604d5ff733f5eb54a

lo5:
Dockerfile good to go :rocket: 

The Dockerfile can look like:
```
FROM ashrith/wave-r-base:latest
RUN git clone --depth 1 https://github.com/h2oai/wave
RUN cd wave/r && make package
```
The other lines can go into a static image on docker hub (`ashrith/wave-r-base`?) to speed up CI.

ashrith:
> Dockerfile good to go 🚀
> 
> The Dockerfile can look like:
> 
> ```
> FROM ashrith/wave-r-base:latest
> RUN git clone --depth 1 https://github.com/h2oai/wave
> RUN cd wave/r && make package
> ```
> 
> The other lines can go into a static image on docker hub (`ashrith/wave-r-base`?) to speed up CI.



> Dockerfile good to go 🚀
> 
> The Dockerfile can look like:
> 
> ```
> FROM ashrith/wave-r-base:latest
> RUN git clone --depth 1 https://github.com/h2oai/wave
> RUN cd wave/r && make package
> ```
> 
> The other lines can go into a static image on docker hub (`ashrith/wave-r-base`?) to speed up CI.

@lo5 done. Now we have our own image  😄 
https://github.com/h2oai/wave/pull/492/commits/121ebc1eab222a19c594ed5a36df09ea58c9f5a2

lo5 (approved):
Thanks @ashrith - that was quite the marathon - reverse engineering the command protocol on your own! Your effort on this also forced some much required simplification to the backend app protocol. Here's to the upcoming R interactive app support :champagne: !

ashrith:
thank you!

PR #507, opened 2021-01-18
Fix: form visibility

### LINKED ISSUE #484: Visible prop on textbox not working
via @5675sp 
#### Wave SDK Version, OS

0.10.0

#### Actual behavior

After setting `visible` prop as `False`, textbox component is visible nonetheless.

#### Expected behavior

Textbox should not be visible.

### LINKED ISSUE #817: Message bar visibility doesn't hide component
#### Wave SDK Version, OS
v0.15.0 in Chrome on Mac

#### Actual behaviour
`ui.message_bar` with `visible=False` still shows the background.

#### Expected behaviour
It should be hidden completely.

#### Steps to Reproduce
```python
ui.message_bar(type='success', text='Predictions successfully completed!', visible=False)
ui.message_bar(type='success', text='Predictions successfully completed!', visible=True)
```

Screenshot 2021-05-16 at 1 04 23 AM

Maybe related to #484 ?

mturoci:
Moved the visibility handling logic to parent form component instead of dealing with it at component level.

There was also a minor spacing issue - we applied `marginTop: 0` to `:not(:first-child)` selector. The problem is, this selector looks at the DOM (hidden elements are still present in the DOM), not whether the element is visible or not. I managed to fix it via CSS.

Closes #484
Closes #817

### DIFF
--- a/py/examples/form_visibility.py
+++ b/py/examples/form_visibility.py
@@ -1,0 +1,41 @@
+# Form / Visible
+# Use "visible" property to control whether form element should be shown / hidden.
+# ---
+from h2o_wave import main, app, Q, ui
+
+
+@app('/demo')
+async def serve(q: Q):
+    if not q.client.initialized:
+        q.page['example'] = ui.form_card(box='1 1 4 10', items=[
+            ui.text_xl(content='First text'),
+            ui.text_l(content='Second text'),
+            ui.text_m(content='Third text'),
+            ui.text_s(content='Fourth text'),
+            ui.inline([
+                ui.button(name='left1', label='Left1'),
+                ui.button(name='left2', label='Left2'),
+                ui.button(name='left3', label='Left3'),
+            ]),
+            ui.buttons(justify='end', items=[
+                ui.button(name='right1', label='Right1'),
+                ui.button(name='right2', label='Right2'),
+                ui.button(name='right3', label='Right3'),
+            ]),
+            ui.buttons(items=[ui.button(name='show', label='Show'), ui.button(name='hide', label='Hide')])
+        ])
+        q.client.initialized = True
+    items = q.page['example'].items
+    items_to_hide = [
+        items[0].text_xl,
+        items[2].text_m,
+        items[4].inline.items[0].button,
+        items[5].buttons.items[2].button,
+    ]
+    if q.args.hide:
+        for i in items_to_hide:
+            i.visible = False
+    if q.args.show:
+        for i in items_to_hide:
+            i.visible = True
+    await q.page.save()
--- a/py/examples/tour.conf
+++ b/py/examples/tour.conf
@@ -20,6 +20,7 @@ layout.py
 layout_size.py
 layout_responsive.py
 form.py
+form_visibility.py
 text.py
 text_sizes.py
 label.py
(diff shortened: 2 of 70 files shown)


lo5:
Any reason why this uses `visibility: hidden` instead of the previous `display: none`? 
Also, we should find a non-js solution to spacing.

mturoci:
> Any reason why this uses `visibility: hidden` instead of the previous `display: none`?

Futurewise, `display: none` cannot be animated. This means that after resolving #229 (form components will be properly updated instead of recreated on props change), we can make toggled elements leave smoothly instead of snapping. Apart from that, there is not really much of a difference.

> Also, we should find a non-js solution to spacing.

I tried to look for something better, but can try harder.

mturoci:
Found a pure CSS approach for spacing. Added custom data attr `data-visible` to get info whether the element is currently visible or not. Then used `> [data-visible="true"] ~ div'` to select all direct form elements that are preceded by visible elements. This is a working solution for "first **visual** child and onwards.

Also added the needed functionality to horizontal (inline) part.

mturoci:
* Rebased.
* Replaced `visibility: hidden` with `display:none` as discussed.
* Added example demonstrating the feature.

https://user-images.githubusercontent.com/64769322/105163171-dcb09b80-5b13-11eb-814f-cd7077a0b241.mov


Needs discussion: @lo5 this approach will not work for cases when `visible` is defined on elements in `ui.buttons` as it creates its own wrapper at component level. A few ways to resolve this:
* Shouldn't `ui.buttons` be deprecated in favor of `ui.inline` or you left it there just for the sake of not introducing a breaking change?
* Special case it within `button` component.

mturoci:
Rebased against current master.

lo5 (commented):
This PR causes a breaking change: the `visible` attribute has been removed from the API.

mturoci:
TBH I don't remember the reasoning why `visible` was removed from each component. Adding it back to API should cause no harm as it is used anyway.

mturoci:
* rebased
* reverted back `visible` prop per component to avoid breaking change
* special cased the button component to allow using `visible` prop even inside `ui.buttons` component.

mturoci:
Rebased.

mtanco:
@mturoci seeing similar behavior for Toggles, would that be handled automatically in this PR too or should I open a separate issue?

image


Wave 0.16.0
```python
from h2o_wave import site, ui

page = site['/']

page['toggle'] = ui.form_card(
    box='4 4 4 4',
    items=[
        ui.toggle(name='hidden', label='You can\'t see me', visible=False),
        ui.toggle(name='shown', label='You can see me', visible=True),
    ]
)
page.save()
```

mturoci:
@mtanco  This PR fixes visibility for all the form items at once so no need to open a new issue.

lo5:
@mturoci - what would be the minimal fix for this bug (without changes to the generator, or introducing changes to launch.json, precommit.js, etc.)?

mturoci:
The minimal changes are to move visibility handling logic from component scope to form scope. The generator is updated so that every form component has a consistent `visible` prop, but we can go with hardcoding it as it was before ofc.

mturoci:
@lo5 rebased against current master, removed all the extra stuff that is not necessarily part of the fix, removed the generator tweaks and settled back to specifying `visible` per each component.

psinger:
@mturoci What is the status of this issue? Will it be fixed soon? Thanks!

PR #512, opened 2021-01-19
fix: let React handle page updates #508

### LINKED ISSUE #508: Changing box parameters for an existing card fails.
#### Wave SDK Version, OS

0.10.0. Reported via @5675sp . 

```
Hello, I currently have a card name example, and when a button is clicked, I will like the box args to change. I change the card box with the following reference: q.page['example'].box = '1 1 -1 -1'
I am not sure if it's a bug, but when the button is clicked, the box width and height do not change (before the desire change it is: box='1 1 4 5').
```

#### Expected behavior
Expectation is that the user should be allowed to change only the card box attributes for a previously created card.

#### Steps to reproduce
```py
@app('/')
async def serve(q: Q):
    if not q.client.initialized: 
        q.page['wizard'] = ui.form_card(box='1 1 2 4', items=[])
        q.client.initialized = True

    wizard = q.page['wizard']  # Get a reference to the wizard form
    if q.args.step1:
        wizard.items = [ui.buttons([ui.button(name='step2', label='Next', primary=True)])]
        wizard.box = '1 1 3 5'
    else:
        wizard.items = [ui.buttons([ui.button(name='step1', label='Of course!', primary=True)])]
        wizard.box = '1 1 2 4

    await q.page.save()
```

mturoci:
Needs discussion - cc @lo5 

The problem was that websocket sent "box changes" in `msg.d` (I believe that stands for data), which then just calls `changedB` on corresponding card, causing specific card update.  As the page itself didn't really change, `currentPage` stays the same and doesn't call `onSocket` handler that is in charge of updating layout in general.

This PR proposes a solution that allows React itself decide what to update instead of doing it ourselves manually.

Closes #508

### DIFF
--- a/ui/src/qd.ts
+++ b/ui/src/qd.ts
@@ -853,22 +853,19 @@ const
       qd.socket = null
       backoff *= 2
       if (backoff > 16) backoff = 16
-      handle({ t: SockEventType.Message, type: SockMessageType.Warn, message: `Disconneced. Reconnecting in ${backoff} seconds...` })
+      handle({ t: SockEventType.Message, type: SockMessageType.Warn, message: `Disconnected. Reconnecting in ${backoff} seconds...` })
       window.setTimeout(retry, backoff * 1000)
     }
     sock.onmessage = function (e) {
-      if (!e.data) return
-      if (!e.data.length) return
+      if (!e.data?.length) return
       qd.busyB(false)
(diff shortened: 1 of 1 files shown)


lo5:
Fixed via https://github.com/h2oai/wave/commit/b9cff523ebe00a6d1ad0c0393f9c0c831c5498db

PR #521, opened 2021-01-22
Safari iframe fix

### LINKED ISSUE #519: Safari Error: Preview Component in examples.tour is Render Halfway
Wave SDK Version, OS

Wave SDK 0.11.0
OS X / Safari Version Version 14.0.1 (15610.2.11.51.10, 15610)
(This does not happen on OS X/Chrome)

**Error in Safari:** 

Screen Shot 2021-01-21 at 4 36 07 PM

Preview Component(card) is render halfway 
This only happens in Safari NOT in OS X/Chrome 

**Chrome:**

Screen Shot 2021-01-21 at 4 39 08 PM

**Steps To Reproduce**

wave run examples.tour -> Go to **http://localhost:10101/** using Safari -> You will see all Preview Components being render halfway in contrast to how they are being render in Chrome

mturoci:
The core problem is that Safari's `height: 100%` implementation differs from Chrome/FF. It used default height `150px` instead. The iframe wrapper correctly takes up all the space via `flex-grow: 1`, but iframe is still not capable of filling it corectly.

None of the workarounds found could solve this, so I went for absolute positioning.

Discussion: This fix affects `frame_card` only as it works with dynamic height (the only way to control height is to change card's height). I believe it is enough since form frames have the ability to control height manually with default to `150px`. if not, we can exchange the `150px` height default for `flex-grow:1` and absolute positioning in the same way like `frame_card`.

Closes #519

### DIFF
--- a/ui/src/frame.tsx
+++ b/ui/src/frame.tsx
@@ -32,7 +32,12 @@ const
     },
     body: {
       flexGrow: 1,
+      position: 'relative',
     },
+    frame: {
+      position: 'absolute',
+      top: 10, left: 0, right: 0, bottom: 0
+    }
   })
 
 /**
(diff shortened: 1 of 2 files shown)


ui/src/frame.tsx
```diff
@@ -32,7 +32,12 @@ const
     },
     body: {
       flexGrow: 1,
+      position: 'relative',
     },
+    frame: {
+      position: 'absolute',
+      top: 10, left: 0, right: 0, bottom: 0
```
lo5, line 39: The 10px margin seems odd. The inter-component spacing should take care of this, no?

ui/src/frame.tsx
```diff
@@ -32,7 +32,12 @@ const
     },
     body: {
       flexGrow: 1,
+      position: 'relative',
     },
+    frame: {
+      position: 'absolute',
```
lo5, line 38: It would be good to add your PR comment here, so that we don't inadvertently revert the change in the future.

ui/src/frame.tsx
mturoci: Yes, this was a typo.

ui/src/frame.tsx
mturoci: Done.

lo5 (approved):
Thanks @mturoci!

PR #523, opened 2021-01-22
Fix ProgressArc in Safari

### LINKED ISSUE #520: Tall gauge stat not working in Safari
image

mturoci:
`height: 100%` is not working for flex parent with `flex-grow: 1`. Had to use absolute positioning instead.

Closes #520

### DIFF
--- a/ui/src/parts/progress_arc.tsx
+++ b/ui/src/parts/progress_arc.tsx
@@ -27,8 +27,6 @@ interface Props {
 
 const css = stylesheet({
   container: {
-    width: '100%',
-    height: '100%',
     // Height 100% is not working in Safari. Without this, container has 0 height and d3 cannot compute properly.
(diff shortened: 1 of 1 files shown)


ui/src/parts/progress_arc.tsx
```diff
@@ -29,7 +29,9 @@ const css = stylesheet({
   container: {
     width: '100%',
     height: '100%',
-    position: 'relative',
+    // Height 100% is not working in Safari. Without this, container has 0 height and d3 cannot compute properly.
```
lo5, line 32: Maybe we should get rid of the 100% width/height?

mturoci: Good catch!

lo5 (approved):
Thanks @mturoci!

PR #544, opened 2021-02-02
feat: allow using height in vega spec #543

### LINKED ISSUE #543: Allow specifying width/height for vega card in vega spec
Currently, if width/height is specified within vega spec, wave still uses default `300px` height for card without respect to values specified in spec. If these are > `300px`, an overflow occurs.

Repro code:
```py
from h2o_wave import Q, listen, ui, data, app, main, AsyncSite


@app('/')
async def serve(q: Q):
    # Create Layout
    q.page['meta'] = ui.meta_card(box='', title='H2O AutoDoc', layouts=[
        ui.layout(
            breakpoint='xl',
            width='1200px',
            zones=[ui.zone(name='drift_layout', direction=ui.ZoneDirection.COLUMN, justify='center')]
        )
    ])
    await q.page.save()
    # Create vega plot
    vega_spec = '''{
      "width": "container",
      "height": "container",
      "description": "VEGA test",
      "mark": {"type": "bar", "cornerRadiusEnd": 4},
      "encoding": {
        "x": {"field": "drift", "type": "quantitative"},
        "y": {"field": "feature", "type": "ordinal" , "sort": "-x" },
        "color": {"field": "color", "type": "nominal", "scale": null}
      }
    }'''
    scores = [0.6, 0.2, 0.08, 0.05, 0.03, 0.01]
    dif_scores = {f"feat_{i:02d}": scores[i] for i in range(len(scores))}
    features = sorted(dif_scores.keys(), key=lambda k: dif_scores[k], reverse=True)
    vega_data = data(
        fields=["feature", "drift", "color"],
        rows=[[f, float(dif_scores[f]), get_color(dif_scores[f])] for i, f in enumerate(features)])
    q.page['my_vega_plot_container'] = ui.vega_card(
        box=ui.box('drift_layout'),  # order=1, size=2, height='100%'),
        title='Feature drift',
        specification=vega_spec,
        data=vega_data
    )
    await q.page.save()
    vega_spec = '''{
      "width": 500,
      "height": 500,
      "description": "VEGA test",
      "mark": {"type": "bar", "cornerRadiusEnd": 4},
      "encoding": {
        "x": {"field": "drift", "type": "quantitative"},
        "y": {"field": "feature", "type": "ordinal" , "sort": "-x" },
        "color": {"field": "color", "type": "nominal", "scale": null}
      }
    }'''
    q.page['my_vega_plot_with_height'] = ui.vega_card(
        box=ui.box('drift_layout'),  # order=1, size=2, height='100%'),
        title='Feature drift',
        specification=vega_spec,
        data=vega_data
    )
    await q.page.save()


def get_color(drift):
    if drift < 0.2:
        return '#66cc66'
    elif drift < 0.5:
        return '#ff9900'
    else:
        return '#e64759'

```

mturoci:
Needs discussion:

As discussed with @goldentom42, there are basically two ways for manipulating card size in Wave:
1. Card level - specified it in layout definition.
2. Card content - card size is based on its content.

#543 can be resolved easily by (1).

Vega allows controlling its dimensions via [spec](https://vega.github.io/vega-lite/docs/size.html). Which is (2) approach. This PR introduces ability for handling `height` defined in spec, but I don't think it makes sense for `width` as well, since `vega_card` has only plot as content and you always want to stretch it to the full of the container (see images below)
**Stretched**
image
**Static width in vega spec**
image

My opinion is that we should follow general Wave philosophy and treat everything layout related in `layouts` (for flex layout) or `box` (in grid), then let our vega fill whole card by default. This would need a warning in Docs that using Vega width/height directly in spec is discouraged in favor of layouts. This can however be a bit of trouble for seasoned Vega users who are used to controlling dimensions via spec.
wdyt @lo5 ?

Closes #543

### DIFF
--- a/ui/src/vega.tsx
+++ b/ui/src/vega.tsx
@@ -66,10 +66,13 @@ export const
       init = () => {
         const el = ref.current
         if (!el) return
+
+        const spec = JSON.parse(state.specification)
+        if (!isNaN(spec.height)) el.style.height = `${spec.height + 10}px`
         // If card does not have specified height, it uses content. Since the wrapper is empty, it takes very little space - set to 300px by default.
-        if (el.clientHeight < 30) el.style.height = '300px'
+        else if (el.clientHeight < 30) el.style.height = '300px'
+
         const
-          spec = JSON.parse(state.specification),
           data = unpack<any[]>(state.data),
           width = el.clientWidth - 10, // HACK: Vega calculates dimensions with extra 10px for some reason.
           height = el.clientHeight - 10 // HACK: Vega calculates dimensions with extra 10px for some reason.
@@ -123,10 +126,10 @@ export const
   View = bond(({ name, state, changed }: Card<State>) => {
     const
       render = () => {
-        const { specification, data } = state
+        const { specification, data, title } = state
         return (
           <div data-test={name} className={css.card}>
-            <div className='s12 w6'>{state.title}</div>
+            <div className='s12 w6'>{title}</div>
             <div className={css.body}>
               <XVegaVisualization key={xid()} model={{ specification, data }} />
             </div>
--- a/ui/src/vega.tsx
+++ b/ui/src/vega.tsx
@@ -72,7 +72,9 @@ export const
           el.style.height = `${spec.height + 10}px`// HACK: Vega calculates dimensions with extra 10px for some reason, increase container for 10px as well.
         }
         // If card does not have specified height, it uses content. Since the wrapper is empty, it takes very little space - set to 300px by default.
-        else if (el.clientHeight < 30) el.style.height = '300px'
+        else if (el.clientHeight < 30) {
+          el.style.height = '300px'
+        }
 
         const
           data = unpack<any[]>(state.data),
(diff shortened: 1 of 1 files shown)


ui/src/vega.tsx
```diff
@@ -66,10 +66,13 @@ export const
       init = () => {
         const el = ref.current
         if (!el) return
+
+        const spec = JSON.parse(state.specification)
+        if (!isNaN(spec.height)) el.style.height = `${spec.height + 10}px`
```
lo5, line 71: Any reason why the extra 10px is necessary here? Would be good to have a comment here explaining why.

lo5: Also, would be good to include `{}` for the conditionals.

lo5 (changes requested):

ui/src/vega.tsx
mturoci: Thanks, good point, always forget about the comments somehow.

> Also, would be good to include {} for the conditionals

I thought the current style is inline ifs if possible without brackets.

lo5 (approved):
Thanks @mturoci!

PR #564, opened 2021-02-11
fix: remove displaying form items in flex column due to Safari #563

### LINKED ISSUE #563: Safari form regression
#### Wave SDK Version, OS

Wave 0.11 onwards

image

mturoci:
Needs more testing.

Closes #563

### DIFF
--- a/ui/src/form.tsx
+++ b/ui/src/form.tsx
@@ -154,8 +154,6 @@ const
   css = stylesheet({
     card: {
       padding: 15,
-      display: 'flex',
-      flexDirection: 'column'
     },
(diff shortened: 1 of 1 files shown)


lo5:
@mturoci - thanks for catching this. Has the change been verified in Chrome/FF?

mturoci:
I went through the whole `Tour` and `Demo App` without finding any regressions. Seems to be a leftover from intermediary development.

lo5 (approved):
@mturoci - thanks for catching this. Agree - let's adjust the demo app.

PR #566, opened 2021-02-12
Nav initial_value prop

### LINKED ISSUE #565: Allow only initial selection for Nav

A `tour.py` could improve its UX by preselecting nav item based on active hash.

Introduce `initial_value` attr on `ui.nav_card`.

Tried to use `value`, but this allows app developers to manually control which nav item is currently active (clicking on other items does not work out of the box unless handled in python code via `value`). However, I only care about the initial selection without further overhead.

mturoci:
Closes #565

### DIFF
--- a/py/examples/nav.py
+++ b/py/examples/nav.py
@@ -15,7 +15,7 @@ async def serve(q: Q):
     else:
         q.page['nav'] = ui.nav_card(
             box='1 1 2 5',
-            value='#menu/spam',
+            initial_value='#menu/spam',
             items=[
                 ui.nav_group('Menu', items=[
                     ui.nav_item(name='#menu/spam', label='Spam'),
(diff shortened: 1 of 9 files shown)


lo5:
See https://github.com/h2oai/wave/issues/565#issuecomment-778348908

mturoci:
@lo5 I pushed the initial implementation for `Nav` as proposed in #565. If it passes the review, I can apply the same pattern to the rest value-based components (e.g. tabs) to synchronize behavior.

* `value` is used for initial selection by default
* if value is changed explicitly later on, use it to determine active item and prevent further user clicks so as to avoid incompatible states on server and client for active item.

One thing I am not sure about is whether we want to set `q.args` and call `sync` on explicit `value` change from client (latest implementation). I think it makes sense to change the hash url, but not sure about args and sync since the state is already captured on server.

mturoci:
Closing in favor of #1154

PR #569, opened 2021-02-15
fix: Set git PAS to publish docs workflow

dulajra:
Fix for permission denied to commit to protected branch issue in Github actions pipeline.


lo5 (approved):

lo5:
Thanks @dulajra!

PR #585, opened 2021-02-24
feat: wide info card

mturoci:
* Implemented according to spec - https://docs.google.com/document/d/1CcQrHMHihA7tzJNjo7OqItVbxMz6HyL1S7b7Y8EdewI/edit#
* enhanced color contrast function to work with wave custom colors as well - prefixed with `$`
* `PopExpand` icon is not available for Fluent version we use. Used `MiniExpand` instead. Other option is to update Fluent.
* As for "sensible object fit" I went with cover since it looks the best. The cropping effect is minimized if correct image is provided, e.g. wide image for wide card or tall image for tall card. Other possible alternative is to provide an additional switch attribute that would preserve the image ratio even at the cost of not filling entire half of the card.

https://user-images.githubusercontent.com/64769322/109000600-36bcf780-76a4-11eb-9718-9d5eef561f16.mov

I wasn't sure what you meant by _If an icon is specified, cap the max width_ @lo5. Icons are rendered as text elements so I have set default font size to `80px` for now, open for discussion.


shanaka-rajitha:
@lo5 I have updated a sample of an improved wide info card, the layout now provides a smooth transition from top to bottom of the text compared to the previous which highlights the title mostly. I have added extra spacing between image and text to give some breathing space to avoid layout congestion. 
[Design URL](https://xd.adobe.com/view/4dc343a9-48ea-441f-9448-bb2fb83788a4-1890/)


Wide infor card samples



> The typography for category and title need work. Is this something you can tackle @sandruparo?

lo5:
Looks good - what would the card look like with [Inter](https://rsms.me/inter/)? 

[Tracking metrics are here](https://github.com/h2oai/wave/blob/master/ui/src/index.scss#L98-L117).

shanaka-rajitha:
> Tracking metrics are here

@lo5 I have updated the designs to use `Inter` font type.

mturoci:
Updated according to proposed design:
image

lo5 (approved):

PR #586, opened 2021-02-24
fix: stretch slider for available space #582

### LINKED ISSUE #582: Slider does not work with ui.inline
via @tkmaker
image

mturoci:
The sliders were cramped within `ui.inline` because there was no default width specified originally. This PR sets it for 100% which doesn't change anything if slider specified as a form child and takes up available space for inline. This seems reasonable for me since short sliders (not wide enough) are harder to use because of smaller dragging area. Another possiblity is to specify default min width or exposing width attr. 100% width solution feels the best though.

Fixed with 100% width:
Screen Shot 2021-02-24 at 4 40 38 PM

Possible improvements: Make our range slider UX the same as Fluent one, something like:
Screen Shot 2021-02-24 at 4 29 16 PM

Closes #582

### DIFF
--- a/ui/src/form.tsx
+++ b/ui/src/form.tsx
@@ -197,11 +197,12 @@ export const
           // All form items are wrapped by their component name (first and only prop of "m").
           [componentKey] = Object.keys(m),
           { name, visible = true } = m[componentKey],
-          visibleStyles: React.CSSProperties = visible ? {} : { display: 'none' }
+          visibleStyles: React.CSSProperties = visible ? {} : { display: 'none' },
+          width = componentKey === 'slider' || componentKey === 'range_slider' ? '100%' : 'auto'
 
         return (
           // Recreate only if name or position within form items changed, update otherwise.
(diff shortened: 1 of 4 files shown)


shanaka-rajitha:
Few suggestions and improvements on the UI 

- Slider thumb to be increased to minimum `20px` by `20px` and have a dropshadow.
- Slider thumb to user `cursor:pointer` on hover and active state.
- Active slider tracker can have minimum height of `6px` for more visibility and inactive tracker area height to be `4px`. 
- Value labels can be on either sides of the slider, no need to repeat values. 
- For step slider its good to show the increments with a tick mark on the tracker.
- On disabled slider show the `cursor: not-allowed` on hover.

[Design URL](https://xd.adobe.com/view/69c444cc-5907-4479-8f6b-4f3ea7d996da-3c97/)

Range slider UI

mturoci:
I am not sure we want to interfere with Fluent design, cc @lo5.

mturoci:
Merging the fix, leaving the redesign for later.

PR #590, opened 2021-02-25
Update routing.md

tomkraljevic:
on_heads should be on_tails in a couple of places.


lo5 (approved):
Thanks @tomkraljevic.

PR #598, opened 2021-03-03
Spinbox trigger attr

### LINKED ISSUE #595: Add the trigger option to spinbox
Right now, the `spinbox` component doesn't have a `trigger` attribute. I had to use another component in my application instead. `spinbox` would be a better fit.

mturoci:
This PR adds a `trigger` prop to the `Spinbox` component. The implementation is hacky though, since current Fluent version does not support `onChange` callback for `SpinButton` (see https://github.com/microsoft/fluentui/issues/5326). Fluent implements its own `onChange` from version 8. I implemented the `onChange` by hand, removing the need for `onBlur` that was used as a workaround I believe.

Side changes:
* Added default values to docstrings.
* Removed `onBlur` in favor of the new `onChange`.

Closes #595


mturoci:
Managed to get rid of the previous ugly hacky solution. The current on hooks up `onChange` listener directly to an input via `inputProps`.

PR #611, opened 2021-03-15
feat: prevent empty zones from taking up space #610

### LINKED ISSUE #610: Empty zones in Flex layout should not take up space

It is a common scenario that some parts of layout are not initially visible. It is however convenient to have the layout defined upfront. The problem however is, that not yet visible parts take up empty space, making the app look ugly.

https://user-images.githubusercontent.com/64769322/111138728-92322500-8580-11eb-81eb-4da786c4e203.mov

It would be nice if SDK could handle cases when there is no content yet in `ui.zone` in order to achieve this:

https://user-images.githubusercontent.com/64769322/111138566-5e56ff80-8580-11eb-868e-77385b54b9a7.mov

It is possible to achieve the above by handling this manually:
```py
q.page['meta'].layouts[0].zones[1].zones[1].zones[1].size = '0' # for hiding
q.page['meta'].layouts[0].zones[1].zones[1].zones[1].size = '1' # for showing
```
However, this is an unnecessary code pollution IMO.

mturoci:
Closes #610

### DIFF
--- a/ui/src/page.tsx
+++ b/ui/src/page.tsx
@@ -185,7 +185,7 @@ const
     around: 'space-around',
     stretch: 'stretch',
   },
-  toSectionStyle = (zone: Zone, direction?: S): React.CSSProperties => {
+  toSectionStyle = (zone: Zone, cardSlots: CardSlot[], direction?: S): React.CSSProperties => {
     const
       style: React.CSSProperties = {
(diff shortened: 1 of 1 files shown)


mturoci:
Closing in favor of a better solution.

PR #660, opened 2021-03-23
Stop scroll propagation for plots

### LINKED ISSUE #659: Plot zoom via scroll interferes with native browser scroll
#### Wave SDK Version, OS

0.13

#### Actual behavior

https://user-images.githubusercontent.com/64769322/112139742-727bac00-8bd3-11eb-9529-a5d451b09045.mov

#### Expected behavior

Stop scrolling the page when zooming the plot.

mturoci:
It seems like this is a bug from the plotting lib. There are 2 possible solutions:
* Since `zoom-view` is not working on Firefox, one option is to remove it so that we get consistent behavior across all the browsers.
* If we however don't want to do that, this PR stops further event propagation thus page scroll does not happen. This behavior is turned off for Firefox as it makes no sense to forbid the page scroll and not provide zoom-view as well.

Fix in action:

https://user-images.githubusercontent.com/64769322/112176610-8980c500-8bf8-11eb-8b3b-ab5f0219aa0d.mov1

Closes #659

### DIFF
--- a/ui/src/plot.tsx
+++ b/ui/src/plot.tsx
@@ -708,7 +708,6 @@ const
         geometries,
         annotations,
         interactions: [
-          { type: 'view-zoom' },
           { type: 'drag-move' }, // custom
         ],
         tooltip: {
--- a/ui/src/plot.tsx
+++ b/ui/src/plot.tsx
@@ -777,18 +777,11 @@ export const
           init()
         }
       },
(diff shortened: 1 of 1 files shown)


lo5:
Yes, I vote for removing the `view-zoom` interaction altogether.

lo5 (approved):

Issue #661, opened 2021-03-23
Values are not submitted to the server


VijithaEkanayake:
#### Wave SDK Version, OS
<=0.13.0

(e.g. 4.2.0, Windows)

#### Actual behavior
setting a value attribute of the ui.command doesn't submitted to the server. It always passes as True.
This only happens if I use ui.command inside toolbar_card.
(A clear and concise description of what happened.)

#### Expected behavior
value should be submitted to the server if the value attr is defined inside the command.

(A clear and concise description of what you expected to happen.)

#### Steps To Reproduce
Start wave server.
Run below sample wave application and click on the toolbar item of the application. values are not submitted to the server.

from h2o_wave import main, app, Q, ui
```@app('/demo')
async def serve(q: Q):
    print(q.args)
    if q.args.spam:
        q.page['blurb'].content = "Sorry, we're out of spam!"
    elif q.args.ham:
        q.page['blurb'].content = "Sorry, we're out of ham!"
    else:
        q.page['nav'] = ui.toolbar_card(
            box='1 1 4 1',
            items=[
                ui.command(name='spam', label='Spam', value="aaaa"),
                ui.command(name='ham', label='Ham', value="bbb")
            ],
        )
        q.page['blurb'] = ui.markdown_card(
            box='1 2 4 2',
            title='Store',
            content='Welcome to our store!',
        )
    await q.page.save()
```

lo5:
Fixed via https://github.com/h2oai/wave/commit/2e7871dda27bab8f0b1b600c0d0dfdeaa26a9ea5

PR #662, opened 2021-03-23
test: Testcases for Submit command value when clicked, if available

VijithaEkanayake:
testcases for https://github.com/h2oai/wave/issues/661

### DIFF
--- a/ui/src/toolbar.test.tsx
+++ b/ui/src/toolbar.test.tsx
@@ -13,22 +13,60 @@
 // limitations under the License.
 
 import React from 'react'
-import { render } from '@testing-library/react'
+import { fireEvent, render } from '@testing-library/react'
 import { View } from './toolbar'
 import * as T from './qd'
 
 const
   name = 'toolbar',
+  commandName = 'toolbar_command',
+  commandValue = 'toolbar_command_value',
+  commandLabel = commandName,
   toolbarProps: T.Card<any> = {
     name,
     state: { items: [] },
     changed: T.box(false)
+  },
+  toolbarPropsWithValueAttr: T.Card<any> = {
+    name,
+    state: { items: [{ name: commandName, value: commandValue, label: commandLabel }] },
+    changed: T.box(false)
+  },
+  toolbarPropsWithoutValueAttr: T.Card<any> = {
+    name,
+    state: { items: [{ name: commandName, label: commandLabel }] },
+    changed: T.box(false)
   }
 
 describe('Toolbar.tsx', () => {
+  beforeEach(() => {
+    T.qd.args[commandName] = null
+    jest.clearAllMocks()
+  })
 
   it('Renders data-test attr', () => {
     const { queryByTestId } = render(<View {...toolbarProps} />)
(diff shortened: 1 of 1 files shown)


ui/src/toolbar.test.tsx
```diff
@@ -13,22 +13,60 @@
...
-import { render } from '@testing-library/react'
+import { fireEvent, render } from '@testing-library/react'
 import { View } from './toolbar'
 import * as T from './qd'
 
 const
   name = 'toolbar',
+  commandName = 'toolbar_command',
+  commandValue = 'toolbar_command_value',
+  commandLabel = commandName,
```
mturoci, line 24: Not needed as it is the name.

ui/src/toolbar.test.tsx
```diff
@@ -13,22 +13,60 @@
...
+  toolbarPropsWithValueAttr: T.Card<any> = {
+    name,
+    state: { items: [{ name: commandName, value: commandValue, label: commandLabel }] },
+    changed: T.box(false)
+  },
+  toolbarPropsWithoutValueAttr: T.Card<any> = {
+    name,
+    state: { items: [{ name: commandName, label: commandLabel }] },
+    changed: T.box(false)
   }
```
mturoci, line 39: No need for these, use existing `toolbarProps` with base state: ` state: { items: [{ name: commandName, label: commandLabel }] },` and add value within specific test case as needed. Be careful to not change the original base props instance for further tests.

ui/src/toolbar.test.tsx
```diff
@@ -13,22 +13,60 @@
...
+  })
 
   it('Renders data-test attr', () => {
     const { queryByTestId } = render(<View {...toolbarProps} />)
     expect(queryByTestId(name)).toBeInTheDocument()
   })
-})
\ No newline at end of file
+
+  it('Sets args and calls sync on click - with value attr', () => {
```
mturoci, line 52: missing check if sync was called

ui/src/toolbar.test.tsx
```diff
@@ -13,22 +13,60 @@
...
+    const syncMock = jest.fn()
+    T.qd.sync = syncMock
+
+    const {getByText} = render(<View {...toolbarPropsWithValueAttr} />)
+    fireEvent.click(getByText(commandLabel))
+
+    expect(T.qd.args[commandName]).toBe(commandValue)
+  })
+
+  it('Sets args and calls sync on click - without value attr', () => {
```
mturoci, line 62: missing check if sync was called

mturoci (changes requested):
Thanks @VijithaEkanayake. A few comments. Please also include tests for case when value starts with a hash.

VijithaEkanayake:
> Thanks @VijithaEkanayake. A few comments. Please also include tests for cases when the value starts with a hash.

@mturoci All the suggestions are introduced. Kindly have a look.

VijithaEkanayake:
@mturoci  Please refer to PR: https://github.com/h2oai/wave/pull/717 which introduces the requested changes. I don't seem to have the permissions to request a review from anyone it just default reviewer.

PR #716, opened 2021-03-30
feat: compact file upload #623

### LINKED ISSUE #623: file_upload component - Need a simple button like file-upload component
wave sdk 0.13.0

The current file_upload component is very bulky with browse icon and drag and drop features. It will be good to have a simple button like component for file-upload so that it can be embedded in a single row without taking much space.

mturoci:
Closes #623

### DIFF
--- a/py/examples/file_upload.py
+++ b/py/examples/file_upload.py
@@ -7,9 +7,10 @@ from h2o_wave import main, app, Q, ui
 
 @app('/demo')
 async def serve(q: Q):
-    if 'file_upload' in q.args:
+    if 'file_upload' in q.args or 'file_upload_compact' in q.args:
         q.page['example'] = ui.form_card(box='1 1 4 10', items=[
             ui.text(f'file_upload={q.args.file_upload}'),
+            ui.text(f'file_upload_compact={q.args.file_upload_compact}'),
             ui.button(name='show_upload', label='Back', primary=True),
         ])
     else:
@@ -17,7 +18,10 @@ async def serve(q: Q):
             box='1 1 4 10',
             items=[
                 ui.file_upload(name='file_upload', label='Upload!', multiple=True,
-                               file_extensions=['csv', 'gz'], max_file_size=10, max_size=15)
+                               file_extensions=['csv', 'gz'], max_file_size=10, max_size=15),
+                ui.file_upload(name='file_upload_compact', label='Upload!', compact=True,
+                               multiple=True, file_extensions=['jpg', 'png'], max_file_size=1, max_size=15),
+                ui.button(name='submit', label='Submit')
             ]
         )
     await q.page.save()
--- a/ui/src/file_upload.tsx
+++ b/ui/src/file_upload.tsx
@@ -15,8 +15,8 @@
 import * as Fluent from '@fluentui/react'
 import React from 'react'
 import { stylesheet } from 'typestyle'
-import { B, bond, box, F, Id, qd, S, U, xid } from './qd'
-import { centerMixin, clas, dashed, displayMixin, padding } from './theme'
+import { B, bond, box, F, Id, qd, S, U } from './qd'
+import { centerMixin, clas, cssVar, dashed, displayMixin, padding } from './theme'
 
 /**
  * Create a file upload component.
@@ -37,6 +37,8 @@ export interface FileUpload {
   max_size?: F
   /** The height of the file upload, e.g. '400px', '50%', etc. */
   height?: S
+  /** Indicates whether file upload should be displayed in minimal mode without drag & drop functionality. Defaults to false. */
+  compact?: B
   /** True if the component should be visible. Defaults to true. */
   visible?: B
   /** An optional tooltip message displayed when a user clicks the help icon to the right of the component. */
@@ -54,7 +56,7 @@ const
       boxSizing: 'border-box',
     },
     uploadDragging: {
-      border: dashed(2, 'var(--text)'),
+      border: dashed(2, cssVar('$themePrimary')),
     },
     uploadLabel: {
       ...centerMixin(),
@@ -62,9 +64,11 @@ const
       fontSize: 14,
       fontWeight: 600,
       borderRadius: 2,
-      background: 'var(--text)',
-      color: 'var(--page)',
-      minWidth: 80,
+      background: cssVar('$themePrimary'),
+      color: cssVar('$page'),
+      minWidth: 120,
+      boxSizing: 'border-box',
+      height: 32,
       $nest: {
         '&:hover': {
           cursor: 'pointer'
@@ -81,6 +85,9 @@ const
       right: 0,
       cursor: 'pointer'
     },
+    compact: {
+      display: 'flex'
+    }
   })
 const convertMegabytesToBytes = (bytes: F) => bytes * 1024 * 1024
 export const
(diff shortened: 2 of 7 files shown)


ui/src/file_upload.tsx
```diff
@@ -160,11 +163,12 @@ export const
         const fileArr = Array.from(files)
 
         const errMsg = validateFiles(fileArr)
-        if (errMsg) {
-          errorB(errMsg)
-          return
+        if (errMsg) errorB(errMsg)
```
lo5, line 166: I would preserve the if/return pattern. Also, add braces for both if/else blocks if either block has multiple statements.

ui/src/file_upload.tsx
```diff
@@ -37,6 +37,8 @@ export interface FileUpload {
   max_size?: F
   /** The height of the file upload, e.g. '400px', '50%', etc. */
   height?: S
+  /** Indicates whether file upload should be displayed in minimal mode without drag & drop functionality. Defaults to false. */
```
lo5, line 40:
```suggestion
  /** True if the component should be displayed compactly (without drag-and-drop capabilities). Defaults to false. */
```

py/examples/file_upload.py
```diff
@@ -7,17 +7,21 @@
 
 @app('/demo')
 async def serve(q: Q):
-    if 'file_upload' in q.args:
+    if 'file_upload' in q.args or 'file_upload_compact' in q.args:
```
lo5, line 10: Can we break this into two separate examples? The UI of the example looks odd.

py/examples/file_upload.py
```diff
@@ -7,17 +7,21 @@
...
     else:
         q.page['example'] = ui.form_card(
             box='1 1 4 10',
             items=[
                 ui.file_upload(name='file_upload', label='Upload!', multiple=True,
-                               file_extensions=['csv', 'gz'], max_file_size=10, max_size=15)
+                               file_extensions=['csv', 'gz'], max_file_size=10, max_size=15),
+                ui.file_upload(name='file_upload_compact', label='Upload!', compact=True,
+                               multiple=True, file_extensions=['jpg', 'png'], max_file_size=1, max_size=15),
+                ui.button(name='submit', label='Submit')
```
lo5, line 24: primary=True

lo5: Another improvement: 
```
ui.file_upload(name='file_upload', label='Select one or more files to upload', ...
```

lo5 (changes requested):

ui/src/qd.ts
```diff
@@ -807,9 +809,25 @@ export const qd: Qd = {
       }
     return { get, put, set, del, drop, sync }
   },
-  sync: () => {
+  sync: async () => {
```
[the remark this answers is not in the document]
mturoci: 2 reasons:
* When should be the files submitted? It has to be submitted before sync AFAIU so that args can contain references to uploaded files.
* How would I sync busy overlay with both files upload and long-running sync?

ui/src/qd.ts
```diff
@@ -784,6 +785,7 @@ export const qd: Qd = {
   username: null,
   editable: false,
   dialogB: box(null),
+  files: new Map<S, File[]>(),
```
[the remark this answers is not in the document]
mturoci: each instance has its own `model.name` which is key of this map. When multiple components are filled and to be submitted, whole map is iterated and files are sent to server.

ui/src/file_upload.tsx
```diff
@@ -266,41 +264,51 @@ export const
...
+      render = () => (
+        <div style={displayMixin(visible)}>
+          {
+            compact ? (
+              <>
+                {label && <Fluent.Label>{label}</Fluent.Label>}
+                <div className={css.compact}>
+                  <Fluent.TextField data-test={`textfield-${name}`} readOnly value={fileNamesB()} styles={{ root: { flexGrow: 1 } }} errorMessage={errorB()} />
+                  <input id={name} data-test={name} type='file' hidden onChange={onChange} accept={fileExtensions?.join(',')} multiple={multiple} />
+                  <label htmlFor={name} className={css.uploadLabel}>Browse files</label>
```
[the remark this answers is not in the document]
mturoci: Good catch, will fix.

ui/src/file_upload.tsx
mturoci: will do

py/examples/file_upload.py
mturoci: sure

ui/src/qd.ts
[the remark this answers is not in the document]
lo5: Can't you perform the upload in the file-upload component, wait for the upload to complete, and then call `sync()` instead of implementing the file-upload logic inside `sync()`? 

To show/hide the busy overlay, you can call `qd.busyB(true)`/`qd.busyB(false)` from anywhere.

ui/src/qd.ts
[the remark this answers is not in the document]
lo5: This map wouldn't be required if each file upload component maintained its own list of files. I don't see a strong reason to maintain this in `qd.ts`.

mturoci:
Moved the upload logic from global `qd.ts` to local file upload component as suggested. The downside of this approach as noted previously is that 2 sync calls are now required instead of one meaning regular form data are submitted in the first run and uploaded file names in the second.

lo5 (changes requested):
The implementation needs improvement. The component should upload the file(s), wait for the upload to complete, and finally call sync() once so that control is passed to the app. I don't understand why 2 sync() calls are required. The files are uploaded to the server, and the sync calls are redirected to the app via WS, after which the app has to download the files from the server for processing, so as long as the app gets the sync() message *after* the files are guaranteed to be available on the server, we're fine. 

Secondly, relying on `busyB()` state is not the right thing to do.

mturoci:
New implementation: upload files directly to server once user picks the files in OS file picker. The file names (hashes) are stored in qd.args and submitted with the nearest sync.

shanaka-rajitha:
@lo5 @mturoci I have updated sample designs to [Design URL](https://xd.adobe.com/view/8cee53cc-2973-4ad6-b764-e9aeec87c563-d9bd/)

_**Compact Layout**_
File Upload Compact

**_Default Layout_**
File Upload Default

mturoci:
@shanaka-rajitha There are a few more states to be designed:
* Error state
* Success state
* Files selected, but not yet uploaded (Drag and drop component only)
* Upload progress

shanaka-rajitha:
@mturoci I have updated the design file to show error, success and upload states. 

File Upload states

mturoci:
@shanaka-rajitha the submit button is not really a part of this component, but rather part of a whole form which means the loading state proposed has to be changed. I guess temporarily changing `Browse File` right button to loading spinner would suffice. @lo5 wdyt?

lo5:
Demo ^ looks good to me. 
@mturoci I thought you asked for designs on error/success states? If not, we can merge and close.

mturoci:
I asked for success / error / loading state for the drag&drop component mostly since the latest mockups uploaded were about that one. @shanaka-rajitha my apologies, I should have been more explicit.

@lo5 compact part UI / UX looks good to me as well. I suggest merging this one and redesigning the drag&drop component in a separate issue.

shanaka-rajitha:
@mturoci Current implementation with the loading screen looks good to show upload file progress 👍. As for the drag and drop success / error / loading state lets address on a separate task.

PR #717, opened 2021-03-31
test: Testcases for Submit command value when clicked, if available 

VijithaEkanayake:


lo5 (approved):
Thanks @VijithaEkanayake!

PR #718, opened 2021-04-01
Add Wave ML guide to docs

### LINKED ISSUE #683: Add WaveML to Wave docs
It would be good to have all the installation and usage details consolidated in one place (information from the current API, readme, and announcement blog post).

geomodular:
If PR is merged, `Wave ML` guide will be added to the docs. I included a description, current API overview, and function prototypes. We can also add a Wine tutorial to the `Getting Started` section if desired.

Resolves #683

### DIFF
--- a/website/docs/wave-ml.md
+++ b/website/docs/wave-ml.md
@@ -28,12 +28,12 @@ import h2o_wave_ml
 
 Wave ML provides four high-level functions:
 
-- [`build_model()`](wave-ml#build_model): Train a model on a dataset, given the column to be predicted.
-- [`Model.predict()`](wave-ml#modelpredict): Make a prediction.
-- [`save_model()`](wave-ml#save_model): Save your model.
-- [`load_model()`](wave-ml#load_model): Load your previously saved model.
+- [`build_model()`](api/h2o_wave_ml/ml#build_model): Train a model on a dataset, given the column to be predicted.
+- [`Model.predict()`](api/h2o_wave_ml/ml#predict): Make a prediction.
+- [`save_model()`](api/h2o_wave_ml/ml#save_model): Save your model.
+- [`load_model()`](api/h2o_wave_ml/ml#load_model): Load your previously saved model.
 
-Use [`build_model()`](wave-ml#build_model) to train a model. The function accepts a dataset and a *target column* (the column to be predicted):
+Use [`build_model()`](api/h2o_wave_ml/ml#build_model) to train a model. The function accepts a dataset and a *target column* (the column to be predicted):
 
 ```py {3}
 from h2o_wave_ml import build_model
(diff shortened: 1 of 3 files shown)


website/docs/wave-ml.md
```diff
@@ -0,0 +1,200 @@
...
+To load a saved model, use [`load_model()`](wave-ml#load_model):
+
+```py {1}
+model = load_model(model_path)
+predictions = model.predict(file_path='./test.csv')
+```
+
+## Functions
+
+<div className='api'>
```
lo5, line 84: The HTML in this section makes it difficult to maintain - it's better if this is reformatted as markdown, or generated and added to the API section.

lo5 (changes requested):

website/docs/wave-ml.md
geomodular: > The HTML in this section makes it difficult to maintain

It's the autogenerated HTML from Wave docs to match the website API style. That's the only reason I used HTML, to match the style.

> it's better if this is reformatted as markdown, or generated and added to the API section.

True.

We cannot autogenerate it since it's in another repo. I believe we can create a script to clone the repo and autogenerate the docs. Is that the approach we want to try?

lo5: > I believe we can create a script to clone the repo and autogenerate the docs. Is that the approach we want to try?

Yes. Currently we generate the API docs from Python at build-time (the generated .md files are not checked into git). 
The `make` task can pull the repo (clone if missing), extract docs, and then build the main docusaurus site. This way, you can probably reuse the pdoc .md generator in the wave repo.

geomodular: Will try that, thanks!

geomodular:
API for *Wave ML* is now autogenerated. It's done using the regular `make setup && make docs` targets inside `py/` directory. The version is specified here: https://github.com/h2oai/wave/blob/1b424e39dd31f54e83f4c3449b8fa045bfc8d6ef/py/Makefile#L1

The version needs to be specified as we don't want to publish the current master branch.

## Notes
- During the Makefile `setup` stage, `h2o-wave-ml` has to be installed into `venv/` so pdoc can work with dependencies.
- I've chosen API for **v0.3.0** as this is the current published version.

geomodular:
> During the Makefile setup stage, h2o-wave-ml has to be installed into venv/ so pdoc can work with dependencies.

I removed the *Wave ML* installation from the `setup` stage, as it's not a mandatory package. I install it during the `docs` stage.

> I've chosen API for v0.3.0 as this is the current published version.

`v0.4.0` is out so I updated this as well.

lo5 (approved):
Thanks @geomodular!

PR #732, opened 2021-04-14
feat: refresh access token when handling websocket messages.

michal-raska:
Refreshes access token when handling websocket messages. If an error occurs, it's ignored (maybe we want to change this @lo5? )


lo5 (approved):
Thanks @michal-raska  - I'll look into how to best handle the refresh failure.

PR #734, opened 2021-04-14
updated wavegen.ts to handle recursive, list elements in R, ui.R need…

ashrith:
`ui.R` needs to be re-generated. This clears null recursively amongst the `ui.R` elements.


lo5 (approved):
Thanks @ashrith!

PR #736, opened 2021-04-15
updated wavegen with recursive null extractor, period fix, and ui.R 

ashrith:
`ui.R` has been regenerated. period fix.


lo5 (changes requested):
wavegen.ts is not merged properly.

ashrith:
@lo5 I just fixed the `wavegen.tsx` I removed the `diff`'s annotations.

lo5 (approved):

PR #737, opened 2021-04-15
feat: annotator component #227

### LINKED ISSUE #227: Add text annotation component to API
**Goal**
Create a component that allows users to highlight phrases in text content and annotate them.

**Intended use**
NER tagging. See https://prodi.gy/docs/named-entity-recognition#manual-patterns

**Specifics**
The component can have two major properties: `tags: List[str]` and `content: str`. The `content` (markdown allowed, like in `ui.text()`) is rendered as a block of text. The `tags` are rendered as clickable elements that tag the highlighted text.

On submit, the component can send this array:
```
[
   { text: 'Europe', tag: 'continent' }, 
   { text: 'fish and chips', tag: 'food' }, 
]
```

**TBD**
- What should we call this? `ui.annotator()`?
- What should the UX be? Select tag first and then highlight text? Or vice-versa? Or something else?

mturoci:
Closes #227


shanaka-rajitha:
@lo5 @mturoci  I have updated few design samples based on the initial completed design with minor variations. [Design URL](https://xd.adobe.com/view/26f90005-a225-4e48-af10-e6e082222bd9-73aa/)

text annotator layout

mturoci:
Looks good @shanaka-rajitha! @lo5 I suppose we are going with the second (right side) option that resembles our card more.

lo5:
@mturoci - yes

I like the muted colors in the second design. I liked the position of the 'x' in Martin's implementation - allows us to show-on-hover. The inline 'x' in the second design adds to visual noise. It would be good if the highlighted text were inconspicuous.

shanaka-rajitha:
@lo5 @mturoci, Iv updated the annotator design based on feedback and comments. However I feel we can add more line height to allow some breathing space. With current themeing line height is restricted to the font size used.

[Design URL](https://xd.adobe.com/view/26f90005-a225-4e48-af10-e6e082222bd9-73aa/)

text annotator layout

lo5:
leftright

If we go with the right-aligned 'x' button strategy, the 'x' will end up in the lower-middle [1] if the text wraps.

[2] is not affected by word-wrap.

Also the 'x' buttons in the design look extra-small - the 'x' is not legible, and requires more precision to click on.

shanaka-rajitha:
@mturoci I have updated the designs based on the feedback. Following items were improved 

- Close icon size increased from `9px` to `16px`.
- Highlighted text will have a different font color, this is to allow font visibility based on the background.
- Font weight changed to `300` for improved clarity and breathing space.

mturoci:
Changed accroding to design:

https://user-images.githubusercontent.com/64769322/116684833-c4250c80-a9b1-11eb-927a-0be5ad89ef8a.mov

* Close icon size increased to `18px` - XD seems to render icons weirdly, it should be font based, but it seems it has a container (black circle) and text with it.
* Nit: Maybe we could be consistent in border-radiuses - same for header and highlighted text.
* Changed UX of removing to only appear on hover.

shanaka-rajitha:
Looks good @mturoci 💯

- we can make the shapes of the tags and highlight section similar, refer attached image which shows first design with 4px radius and the second with no radius.
- the shadow on tags is it so show selected state.

Screenshot 2021-04-30 at 18 17 04

lo5:
Observations: 
- It's odd that the word-spacing increases when you highlight text.
- The selected-state for tags needs design @shanaka-rajitha (not sure if the drop shadow is the best mechanism here - any alternatives?)

mturoci:
> It's odd that the word-spacing increases when you highlight text.

caused by padding - the same what prodigy does to make a better distinction beween highlighted and non-highlighted text. We can go with just background color change ofc.

lo5:
> caused by padding

Padding around the highlighted section is fine. Padding around individual words within the section is not.

mturoci:
Padding fixed, border radius changed to `2px` - same as Fluent uses.

https://user-images.githubusercontent.com/64769322/116732919-b0969780-a9eb-11eb-836b-ab5dde325618.mov

shanaka-rajitha:
@lo5 I have shared 2 samples for selected-state of tags which is very subtle. Since the tag colors vary from each other having a color shade for active tag would not highlight the purpose. 

selected state of tags UI

lo5:
How about:
1. Gray out (or fade out?) other tags when a tag is active.
2. Make hover state = normal state for grayed-out tags.

shanaka-rajitha:
@lo5 fading out is a good approach

1. On active state other tags will fade in color or be greyed.
2. When hovered over a a faded tag it will show the default color.

Tag active state-v2
Tag active state-v1

lo5:
- Per design, height of tags [1] needs to be smaller, or match the height of the tagged text. Currently it's taller than a button [2].
- [3] [4] compare left vs right padding when you select text.
- [5] [6] looks like border-radius is missing?
- [7] Fade/dim needs to be more pronounced (see design).

mturoci:
Redesigned according the spec.
> Per design, height of tags [1] needs to be smaller, or match the height of the tagged text. Currently it's taller than a button [2].

Not sure where this came from. In the design spec, the tag height is `32px`, content line height `25px` and submit button `42px` (Fluent has button height `32px`). I have fixed it according to design spec with mentioned values with the exception of Fluent button.

> [3] [4] compare left vs right padding when you select text.

Fixed.

> [5] [6] looks like border-radius is missing?

Fixed.

> [7] Fade/dim needs to be more pronounced (see design).

IMO the proposed design with gray color is not ideal. It uses `neutralLight` color, which might not be a good fit with different theme (right now it looks good because it resembles disabled state, but that may not be the case with other themes). Secondly, even the current contrast between background and button text is pretty low, making it hard to read. I consider previous opacity reduction better.

Also some design spacing values are a bit strange:
* tagged text - padding: left `11px`, right `10px`, top `1px`, bottom `2px`
* tag padding: top `6px`, bottom `5px`

Result:
image

mturoci:
@shanaka-rajitha is this design still valid in respect to https://xd.adobe.com/view/0790f950-abbd-41fa-b372-332295fd876f-52c3/?x_product=cc-slack%2F1.5.1 ? If not, can you please update accordingly?

@lo5 I suggest renaming this component to `text_annotator` to allow for multiple types of annotators in future, e.g. `image_annotator`. wdyt?

lo5:
> I suggest renaming this component to text_annotator

Agree

shanaka-rajitha:
> is this design still valid in respect to https://xd.adobe.com/view/0790f950-abbd-41fa-b372-332295fd876f-52c3/?x_product=cc-slack%2F1.5.1 ? If not, can you please update accordingly?

@mturoci will update the following component with minor changes to match new design guidelines.

shihan007:
@mturoci https://xd.adobe.com/view/0790f950-abbd-41fa-b372-332295fd876f-52c3/screen/78995835-9119-4777-b4db-b7f5dd0e7bf6
Screenshot 2021-10-13 at 12 16 21

shihan007:
> leftright
> 
> If we go with the right-aligned 'x' button strategy, the 'x' will end up in the lower-middle [1] if the text wraps.

same issue occur if we have highlighted the top text. Btw I am okay with placing it right @mturoci @lo5 

cc: @shanaka-rajitha

PR #739, opened 2021-04-15
feat(table): checkbox visibility #713

### LINKED ISSUE #713: Add attribute to control checkbox visibility in ui.table()
Via #705: Add `checkbox_visibility` enum attribute based on [this](https://developer.microsoft.com/en-us/fluentui#/controls/web/detailslist#CheckboxVisibility).

mturoci:
Adds checkbox_visibility option.

Also fixed a minior centering bug for checkboxes.

Closes #713

### DIFF
--- a/py/h2o_wave/types.py
+++ b/py/h2o_wave/types.py
@@ -3297,12 +3297,12 @@ class TableRow:
         )
 
 
-_TableCheckboxVisibility = ['always', 'onHover', 'hidden']
+_TableCheckboxVisibility = ['always', 'on-hover', 'hidden']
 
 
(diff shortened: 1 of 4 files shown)


lo5 (changes requested):
The convention for all enum values in Wave is kebab-case. Can you please change `onHover` to `on-hover` and map it correctly to `Fluent.CheckboxVisibility`?

mturoci:
Thanks for the review @lo5! Changed to kebab-case as requested.

lo5 (approved):
Thanks @mturoci!

PR #749, opened 2021-04-19
feat: clipboard component #194

### LINKED ISSUE #194: Copy to clipboard button
via @tomkraljevic

mturoci:
Closes #194

### DIFF
--- a/py/examples/clipboard.py
+++ b/py/examples/clipboard.py
@@ -1,0 +1,14 @@
+
+# Clipboard
+# Use clipboard component to enable user to quickly copy paste text.
+# #clipboard
+# ---
+from h2o_wave import site, ui
+
+content = '''Wave is truly awesome.
+You should try all the features!'''
+page = site['/demo']
+
+page['hello'] = ui.form_card(box='1 1 2 2', items=[ui.clipboard(label='Clipboard', value=content, multiline=True)])
+
+page.save()
--- a/py/h2o_wave/types.py
+++ b/py/h2o_wave/types.py
@@ -5321,6 +5321,66 @@ class Inline:
         )
 
 
+class Clipboard:
+    """Create a copy to clipboard component.
+    Use this component when you want to enable your users to quickly copy paste sections of text.
+    """
+    def __init__(
+            self,
+            value: str,
+            label: str,
+            name: Optional[str] = None,
+            multiline: Optional[bool] = None,
+    ):
+        _guard_scalar('Clipboard.value', value, (str,), False, False, False)
+        _guard_scalar('Clipboard.label', label, (str,), False, False, False)
+        _guard_scalar('Clipboard.name', name, (str,), False, True, False)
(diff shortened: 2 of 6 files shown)


lo5:
Thanks for the PR.

I would prefer something more compact. See [this](https://blog.prototypr.io/3-ways-to-copy-to-clipboard-5077f5774b55) and [this](https://dribbble.com/shots/5640985-Daily-UI-Components-1-Copy-To-Clipboard) for ideas. @shanaka-rajitha might have better ideas.

> render small copy button on hover in case of markdown code

Agree, but only if `multiline` is set.

mturoci:
> Agree, but only if multiline is set

I meant to add this into our existing `ui.text` components by modifying the markdown renderer or do you think it should be only part of this one?

lo5:
@mturoci - better to have a separate component that does one thing well instead of blending it into another, so that we can control/enhance them independently. This PR would only need minor enhancements: if multiline is set, display a tall flavor with a copy button at the top-right, otherwise display a compact flavor with a copy button to the right.

And maybe rename `ui.clipboard()` to `ui.copyable_text()`.

mturoci:
Redesigned as suggested:
image
Current implementation supports only plaintext. It might make sense to support markdown for `multiline` as well.

lo5:
The multiline implementation looks odd - I was expecting the copy button to be inside the text's block to the top-left (see your screenshot above).

Markdown support is not required for now - plain-text is enough.

shanaka-rajitha:
@lo5 @mturoci Based on the last discussion on the comments I have drafted few samples. 
[Design URL](https://xd.adobe.com/view/1726549b-89f7-42d6-9e51-0f6308662114-4275/)

Copy to clipboard

lo5:
- Which multi-line design is recommended? [1] or [2]?
- Why are the empty spaces [3] [4] for?
- What is the purpose of the [5] icon?
- What is the recommendation for the single-line copy-to-clipboard component?

shanaka-rajitha:
@lo5 

- _Which multi-line design is recommended? [1] or [2]?_ **I would recommend design 2 which has a higher readability compared to 1.** 
- _What are the empty spaces [3] [4] for?_ **This section represents the tools available in an editor, similar to what we see when we comment on github.**
- _What is the purpose of the [5] icon?_. **This is the icon representing the resize text area, this is optional though.**
- _What is the recommendation for the single-line copy-to-clipboard component?_ **We can follow the same concept as design 2 but the layout would look similar to a textfield. Let me sketch that up.**
Screenshot 2021-04-30 at 12 00 04

lo5:
@shanaka-rajitha There's no editor here. It's a read only text area.

shanaka-rajitha:
> @shanaka-rajitha There's no editor here. It's a read only text area.

Yes @lo5, this is more of a presentational approach to separate the text from actions. Removing the separator will leave a large white space on the top. 

In this attachment I have reduced some of the extra white space to make it more compact.
Screenshot 2021-04-30 at 14 03 35

PR #753, opened 2021-04-20
Switch Tour to flex layout

### LINKED ISSUE #701: Switch tour app to responsive layout
The tour was authored before responsive layouts shipped. Needs a makeover, and an easier way to copy/paste code (#194?)

mturoci:
https://user-images.githubusercontent.com/64769322/115402327-9db1e500-a1eb-11eb-8e89-aa13883414fa.mov

Open questions:
* Do we want to make the tour even mobile friendly (to show flex layout capabilities) or is flex desktop (current state) enough?

Possible improvements:
* Split nav menu into subgroups (form / plots etc.) to make it easier for user when searching something.
* Dark mode?
* Implement global search to quickly go to the example that user wants. This could be achieved by using either `picker` (not so great) or by #303 (better).
* Clean up all the examples and make them use only the minimum necessary card space (box) so that it doesn't cause overflow within tour iframe even though content is short.
* Distant future - integrate our online IDE and make the tour examples editable with live changes.

Blocked by #749 
Closes #701

### DIFF
--- a/py/examples/tour.py
+++ b/py/examples/tour.py
@@ -161,57 +161,46 @@ app_title = 'H2O Wave Tour'
 
 
 async def setup_page(q: Q):
-    q.page['meta'] = ui.meta_card(
-        box='',
-        title=app_title
-    )
+    q.page['meta'] = ui.meta_card(box='', title=app_title, layouts=[
+        ui.layout(breakpoint='xs', zones=[
+            ui.zone('header'),
+            ui.zone('blurb'),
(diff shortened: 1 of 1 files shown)


lo5 (approved):
Thanks @mturoci!

> Do we want to make the tour even mobile friendly (to show flex layout capabilities) or is flex desktop (current state) enough?

For now this is enough. Your changes have made it better already. I would rather invest time in the Kitchen Sink instead of the tour, because we can put that up publicly online.

Your other ideas are useful - filed a separate issue for it: #776 - we can tackle those once we're through with the Kitchen Sink.

PR #762, opened 2021-04-26
Ashrith/blog r release

ashrith:
added  `R` release. This commit also checked in the local changes I had done for `keychain.go`. I have reverted that change and it is back to what it was. But you will see that change, and the unchange in commit history.


lo5 (approved):
Thanks @ashrith!

PR #779, opened 2021-05-03
feat: Tall info card

mturoci:
Tall info card implementation


shanaka-rajitha:
@mturoci looks good 💯 , two minor tweaks

- outer padding to be 16px
- text align to be left so its easy to read and less strain on the eyes.

lo5:
This card needs design.

The card is intended to be used at smaller widths, with less text than the example illustrates (say, a caption of not more than 15-20 words). Refer to illustrative examples in the design doc.

At that size, the default image height of 300px is too much. See:



Needs work: 
- Typography for [1] and [2]
- Spacing [3] and [4]
- Default `image_height` [5]



@shanaka-rajitha: 
- What's the reasoning behind 16px? All cards currently default to 15px.
- Is left-aligning important if we have 15-20 words split across 3-4 lines?

shanaka-rajitha:
@lo5 as a best practice and to build high-quality digital experiences I refer [MATERIAL DESIGN](https://material.io/design/introduction). 

- For more visually balanced layouts its recommend to use multiples of 4 in designs [source](https://material.io/design/layout/understanding-layout.html).
- Left aligning text is visually smooth to navigate. For the use of few lines centred text is good given the proper use of the typography .
- Typography used in the current design match with the wide infor card.

mturoci:
Reworked according to latest design in https://xd.adobe.com/view/119e73c2-d952-4760-8202-d36d3fc16016-44cd/specs/:
image
image
image

lo5:
I'm assuming this card needs to cover both "centered_tall_info_card / Tall_img_text" and "TallInfoCard" from XD. 

- The  "TallInfoCard" needs icon:text ratio adjustments as mentioned in the design review Keynote. The icon foreground/background colors need adjustment as well
  - in dark mode, the contrast between icon background and card background is too low.
  - in light mode, the icon background vs card background could be improved (seems dull).
- Needs to be clickable. See https://github.com/h2oai/wave/pull/1013#issuecomment-933654656
- The font size and color of the category needs work (it's especially poor in light mode - seems more like a subtitle than a tag/category).

mturoci:
> Needs to be clickable

Was already clickable, made `name` attr required. Also updated example.

https://user-images.githubusercontent.com/64769322/136000589-eca69b25-f359-4fdb-96c2-5510dfb6627b.mov

Waiting for design team to address the rest of the issues.

shanaka-rajitha:
@lo5  @mturoci 

> The "TallInfoCard" needs icon:text ratio adjustments as mentioned in the design review Keynote. The icon foreground/background colors need adjustment as well

Updated designs for dark theme on [1013](https://github.com/h2oai/wave/pull/1013#issuecomment-935446343)

> The font size and color of the category needs work (it's especially poor in light mode - seems more like a subtitle than a tag/category).

For categories, we used `theme primary` color in dark mode. @mturoci it seems that the `primary color` and `text color` in the light theme are both set to be a similar dark, can we change this with some contrast?. Or we might have to work around the colors for the light theme and introduce a new variable.

mturoci:
>  @mturoci it seems that the primary color and text color in the light theme are both set to be a similar dark, can we change this with some contrast?

No, we want to support 2 colors-only (primary and text being the same) themes as well I guess - cc @lo5

> Or we might have to work around the colors for the light theme and introduce a new variable.

We don't know in advance what colors will a user choose so we cannot special case light theme only. The colors have to work with any reasonable theme specified.

See https://github.com/h2oai/wave/pull/1013#issuecomment-935732063.

mturoci:
Redesigned according to new designs:
image
image
image

Same problem as https://github.com/h2oai/wave/pull/1013#issuecomment-935732063.

shanaka-rajitha:
>We don't know in advance what colors will a user choose so we cannot special case light theme only. The colors have to work with any reasonable theme specified. Introducing a new variable could work if based on active theme of course.

@mturoci since we don't know what color the user will choose it's not practical to have colors that will work across any theme specified. This has been the concern from the UX team from the very start, how accessibility/visibility issues will be looked into when full control is given to the user to select colors. 

Can we limit the UI elements the colors apply to from the user selections, as a solution @mturoci @lo5

sandruparo:
> No. We need to make sure the colors are always in sync. If you allow changing colors for a few components only, the whole UI will be inconsistent. The only exception for now are components using `success` | `error` | `warning` colors, e.g. message bar or toast in future.

im sure this is going to be a mess in the future. cause we know our users, as a design team i really don't think we should let them to use their own/preferred  colors in to detail levels, GIving them a theme to choose is fine, if you know what i meant. this will lead to lot of accessibility issues, cause i dont think user have that much wisdom and experience to choose colors which will pass the accessibility / contrast levels. so again agree with @shanaka-rajitha  on this. @shihan007

PR #786, opened 2021-05-06
Fix data type in Tour example

mtanco:
Picker example was failing, values needs to be a list, not a single string


lo5 (approved):
Thanks @mtanco!

PR #806, opened 2021-05-11
fix: UI unit tests

mturoci:
* fixes broken file upload test
* Puts back client side dialog close in addition to the server side one introduced in #803


lo5 (approved):
Thanks @mturoci!

PR #807, opened 2021-05-11
fix: Transpose plot when needed #801

### LINKED ISSUE #801: Horizontal bar plot doesn't work with theme
#### Wave SDK Version, OS
v0.15.0 in Chrome on Mac

#### Actual behaviour
I am unable to create a horizontal bar plot if I set a theme in `meta_card`.

App 1:
```python
from synth import FakeCategoricalSeries
from h2o_wave import main, Q, app, copy_expando, data, ui

@app('/demo')
async def serve(q: Q):
    n = 20
    f = FakeCategoricalSeries()
    v = q.page.add('example', ui.form_card(
        box='1 1 4 5',
        items=[
            ui.visualization(
                plot=ui.plot([ui.mark(type='interval', x='=price', y='=product')]),
                data=data(fields='product price', rows=[(c, x) for c, x, _ in [f.next() for _ in range(n)]], pack=True),
            )
        ],
    ))

    await q.page.save()
```

Output 1:
Screenshot 2021-05-08 at 10 35 35 PM

App 2:
```python
from synth import FakeCategoricalSeries
from h2o_wave import main, Q, app, copy_expando, data, ui

@app('/demo')
async def serve(q: Q):
    q.page['meta'] = ui.meta_card(box='', theme='light')

    n = 20
    f = FakeCategoricalSeries()
    v = q.page.add('example', ui.form_card(
        box='1 1 4 5',
        items=[
            ui.visualization(
                plot=ui.plot([ui.mark(type='interval', x='=price', y='=product')]),
                data=data(fields='product price', rows=[(c, x) for c, x, _ in [f.next() for _ in range(n)]], pack=True),
            )
        ],
    ))

    await q.page.save()
```

Output 2:
Screenshot 2021-05-08 at 10 35 51 PM

The only difference between App 1 and App 2 is `q.page['meta'] = ui.meta_card(box='', theme='light')` and I don't expect Output 2 to change from Output 1.

#### Expected behaviour
Should display horizontal bar plot in both examples (App 2 as well).

#### Steps To Reproduce
Run the two app codes.

#### Comments
I don't know if this is related to theme or something else altogether. But it definitely doesn't work while setting theme.

I also don't know if this is specific to bar plots or others too. I haven't checked.

mturoci:
Caused by: https://github.com/h2oai/wave/blob/7564195bdb5f9b5297ca294948672196edf971eb/ui/src/plot.tsx#L660

The problem was that transpose happens in place, mutating the original `model.plot` (due to obj reference) all the way up to page object stored in `contentB` in `App.tsx`. The issue happens with `form_card` only because currently, all the form items are recreated on every form render (#150). Flow:

1. Transpose plot marks.
2. Correctly return `coordinate` as [['transpose']] so that plot can be rendered correctly.
3. Change theme, which triggers React to update whole app.
4. Since form always recreates its children, recreate new plot, but this time with previously transposed data.
5. Render incorrect plot.

This PR solves this by making a deep copy of `model.plot` that is later worked with, which keeps the original reference (stored in `page`) from mutation.

@lo5 please let me know if this is viable solution since I am not familiar with all the reasoning behind transpose hack.

Closes #801

### DIFF
--- a/ui/src/plot.tsx
+++ b/ui/src/plot.tsx
@@ -786,7 +786,7 @@ export const
         if (el.clientHeight < 30) el.style.height = '300px'
         const
           raw_data = unpack<any[]>(model.data),
-          raw_plot = unpack<Plot>(model.plot),
+          raw_plot = unpack<Plot>(JSON.parse(JSON.stringify(model.plot))), //HACK: Plot may be transposed so always work with own copy.
           marks = raw_plot.marks.map(refactorMark),
           plot: Plot = { marks: marks },
           space = spaceTypeOf(raw_data, marks),


mturoci:
Closing in favor of #746

PR #839, opened 2021-05-26
fix: Drop dups and sort filenames from dir

geomodular:
PR will fix the examples list on Wave page.

There are 2 errors currently:

- Examples are added twice if listed in `tour.conf` and have the prefix `ml` or `db`.
- Examples read by `listdir()` are unsorted (seemingly random).


lo5:
Thanks for the PR @geomodular - I've made a simpler fix in `master` to read the additional examples from a separate web-only conf file. Closing.

PR #860, opened 2021-06-02
feat: Themeable plots #664

### LINKED ISSUE #664: Adapt plot colors to active theme
via @shivam5992 

image

Axis titles are hard to see on dark background.

mturoci:
This PR makes plots use theme colors by default. In order to achieve this, I had to update the `antv/g2` to the latest version which provides rich theming support. Check https://theme-set.antv.vision/  https://g2.antv.vision/en/docs/api/advanced/register-theme for reference.

This PR also changes the renderer used for plots from `canvas` to `svg` in order to enable usage of CSS variables. Another option is to update the plot manually every time the theme changes via JS if this solution wouldn't be ok.

@vopani also suggested making our plot API more flexible as for colors side (e.g. `x_title_color`) - depends on how much control we want to give the devs - wdyt @lo5 ? From technical side, this is a simple thing to do.

@shanaka-rajitha could you please have a look and see if you can come up with a better coloring than I did? I basically used theme text color for all the text related stuff + plot grid and primary for plots themselves. However, I am sure there is a way to improve this. Playing with colors should be straightforward, just edit values within `theme` config object on `plot.tsx`. The example to run is called `plot_theme.py`. Feel free to ping me in case you need help.

Demo:

https://user-images.githubusercontent.com/64769322/120463400-2c458480-c39c-11eb-9b38-1d768935efaf.mov

Side note: We could provide `min_width` & `max_width` to `ui.box` as this would save me from having 3 screen size breakpoints just to make the flex wrap work.

Closes #664


mturoci:
@lo5 wdyt about the demo? Is using `themePrimary` for plot colors sufficient or should I make any changes? E.g. put down lightness a bit or something. IMO the neon theme fits just fine, but the default light theme is somewhat weird.

mturoci:
Merging as is. We can iterate on this later when needed.

PR #870, opened 2021-06-09
WaveML DAI examples

vopani:
This PR adds examples of using WaveML with DAI.
Closes h2oai/wave-ml#43

- [x] **WaveML / DAI:** Build model using train data and predict on test data
- [x] **WaveML / DAI / Save:** Save a trained model and load to predict on new data
- [x] **WaveML / DAI / Categorical:** Select columns as categorical for training
- [x] **WaveML / DAI / Parameters:** Set hyper-parameters while training model
- [x] **WaveML / DAI / AutoDoc:** Download AutoDoc of a trained model
- [x] **WaveML / DAI / Instances:** Get user's DAI instances on cloud

Comments:

* These only work on H2O AI Hybrid Cloud. If run locally, it displays a sign-up message.
* The default setup takes about a minute to run and the outputs are mainly links to DAI and MLOps.
* It requires some `toml` changes to publish on cloud, will ensure those get updated.

### DIFF
--- a/py/examples/ml_dai.py
+++ b/py/examples/ml_dai.py
@@ -1,0 +1,117 @@
+# WaveML / DAI
+# Build Wave Models for training and prediction of classification or regression using Driverless AI.
+# ---
+import os
+
+from h2o_wave import main, app, Q, ui
+from h2o_wave_ml import build_model, ModelType
+from h2o_wave_ml.utils import list_dai_instances
+
+from sklearn.datasets import load_wine
+from sklearn.model_selection import train_test_split
+
+STEAM_URL = os.environ.get('STEAM_URL')
+MLOPS_URL = os.environ.get('MLOPS_URL')
+
+
+@app('/demo')
+async def serve(q: Q):
+    if 'H2_CLOUD_ENVIRONMENT' not in os.environ:
+        q.page['example'] = ui.form_card(
+            box='1 1 -1 -1',
+            items=[
+                ui.text('''This example requires access to Driverless AI running on
+                    <a href="https://h2oai.github.io/h2o-ai-cloud" target="_blank">H2O AI Hybrid Cloud</a> 
+                    and does not support standalone app instances.'''),
+                ui.text('''Sign up at <a href="https://h2o.ai/free" target="_blank">https://h2o.ai/free</a>
+                    to run apps on cloud.''')
+            ]
+        )
+    elif q.args.train:
+        # get DAI instance name
+        for dai_instance in q.client.dai_instances:
+            if dai_instance['id'] == int(q.client.dai_instance_id):
+                dai_instance_name = dai_instance['name']
+
+        # show training progress and details
+        q.page['example'].items[1].dropdown.value = q.args.dai_instance_id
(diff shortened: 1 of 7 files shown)


lo5:
Looks good, although from a readability/maintainability perspective I would just create a new `form_card` instead of using int indices to modify the existing form.

```py
        q.page['example'].items[3].buttons.visible = True
        q.page['example'].items[3].buttons.items[1].button.disabled = False
        q.page['example'].items[4].message_bar.type = 'success'
        q.page['example'].items[4].message_bar.text = 'Training successfully completed!'
        q.page['example'].items[5].progress.visible = False
        q.page['example'].items[6].text.content = model_details
```

Other than that we can merge if you're ready. I wasn't sure if @geomodular had to review or if there's pending work on this.

vopani:
Thanks. I will update the PR with a form-creation function and add all the other remaining WaveML / DAI examples.

geomodular (approved):
LGTM @vopani, I'm fine to be verbose for the sake of readability. It's hard to create a good example, maintaining balance in readability, simplicity, and good practices.

My thoughts:

The index version looks like a compact one but, from my perspective, has two tiny nits:
- The reader has to lookup up the indices to deeply understand the code (to change it perhaps). It's not a problem for a glance, however.
- The structure of the form reference might be confusing without further explanation: `q.page['example'].items[3].buttons.items[1].button.disabled = False`

But I'm fine with this approach as the example is small and showcasing *Wave ML*. On the other side, one might think this is a best practice.

The verbose version is clear.

I tried to come up with another solution, but it might be too much for an example:

```python3
def form_header(q):
    return [
        ui.text(content=DATASET_TEXT),
        ui.dropdown(
            name='dai_instance_id',
            label='Select Driverless AI instance',
            choices=q.client.choices_dai_instances,
            value=q.client.dai_instance_id,
            required=True),
    ]


def form_main(q, disable_training: bool):
    return [
        **form_header(q),
        ui.text(content=STEAM_TEXT, visible=disable_training),
        ui.buttons(items=[
            ui.button(name='train', label='Train', primary=True, disabled=disable_training),
            ui.button(name='predict', label='Predict', primary=True, disabled=True),
        ]),
        ui.message_bar(type='warning', text='Training will take a few seconds')
    ]
 
 def form_in_progress(q):
     ...
 def form_training_completed(q):
    ...
 def form_scoring_completed(q, preds):
    ...
```

The `form_header()` is included in every other form by `**form_header(q)` call. Every form has a name to suggest the defaults in the form.

The usage:
```python3
q.page['example'].items = form_in_progress(q)
```

or

```python3
    box='1 1 -1 -1',
    items=form_main(q, disable_training)
)
```

Thanks, @vopani, good job.

geomodular (approved):
LGTM @vopani and thanks!

vopani:
@geomodular I've added all the 6 examples based on your suggestion. They become longer but are more readable so I think it should be ok.

@lo5 This PR is now complete from my side unless any changes are required. Feel free to merge or add feedback.

Thanks a lot!

lo5 (approved):
Thanks @vopani!

PR #871, opened 2021-06-10
WaveML H2O-3 Algo example

vopani:
This PR adds an example of using WaveML with an Algo type.

 Closes https://github.com/h2oai/wave-ml/issues/62

Also added a minor improvement in categorical column choices based on this fix: https://github.com/h2oai/wave-ml/pull/59


geomodular (approved):
Thanks Rohan! LGTM.

lo5 (approved):
Thanks @vopani!

PR #931, opened 2021-07-23
feat: Update default Scopes and add `-oidc-scopes` cmd argument

### LINKED ISSUE #925: Update default OAuth scopes
Currently, the Wave app can use OIDC with *Keycloak* without a problem, but two problems with *Okta* have emerged:
1. The `q.auth.username` is empty.
2. The `q.auth.refresh_token` is empty as well.

This is because `Wave` server does not [send](https://github.com/h2oai/wave/blob/8321b7c5aafe951d250b5e12a5d1b7ebae4abf77/auth.go#L48) all *Scopes* that *Okta* requires. Two other *Scopes* are required:
1. `profile` to provide `preferred_username` assigned to `q.auth.username` as correctly stated [here](https://github.com/h2oai/wave/blob/8321b7c5aafe951d250b5e12a5d1b7ebae4abf77/auth.go#L47). This is not required in *Keycloak* as it has it automatically included in default scopes:

<kbd></kbd>

2. `offline_access` to enable `refresh_token`. It's not enough to just include this *Scope* but the **Refresh Token** needs to be checked in application settings as well:

<kbd></kbd>

This is automatically available in *Keycloak*.

## Considered solutions

**One quick** solution to `preferred_username` might have been to include `profile` as default scope in API settings:

<kbd></kbd>

... but according to [this](https://devforum.okta.com/t/enable-default-scope/2704) it's not doable.

**The other** is to include it as a default claim in the ID token. This is not doable either, because `preferred_username` is reserved word:

<kbd></kbd>

## Proposal

We can extend our *Scopes* with `profile` without a problem and we should be fine with both *Okta* and *Keycloak*. We can extend with `offline_access` as well but *Keycloak* needs to be updated of *Optional Client Scopes* under *Clients*:

<kbd></kbd>

The authorization will fail otherwise.

In addition to this, we can create a command-line argument, say `-oidc-scopes`, which will *override* default scopes - just in case.

cc @lo5 @gaduimovich @michal-raska @jagadeeshrajarajan 

Sorry for so many screenshots :)

geomodular:
If PR merged, the default scopes for OAuth request will be updated of `profile` and a new command-line argument `-oidc-scopes` will be introduced.

## Changes

The `profile` scope is added so `openid` and `profile` are now default. This change will ensure the `preferred_username` will be available in ID token making `q.auth.username` not to be empty while using e.g. *Okta*. Note that the `profile` scope is listed as a standard scope for OpenID and should always be available.

The `-oidc-scopes` is added so if `openid` and `profile` are not enough, it's possible to list values that override the defaults. Values are comma-separated:

```bash
./waved -oidc-scopes openid,profile,offline_access
```

## Notes
I was considering adding `offline_access` to default scopes as well to enable the `q.auth.refresh_token` in *Okta* but:
1. It actually not enables refresh token in *Okta* right away, there is one other step to allow it in UI as described in #925.
1. *Keycloak* will end up with an error and some setup in UI is necessary as well.
1. Refresh token is not needed for every Wave app.

cc @gaduimovich @jagadeeshrajarajan 

Resolves #925

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -31,6 +31,8 @@ import (
 
 const authCookieName = "oidcsession"
 
+var authDefaultScopes = []string{oidc.ScopeOpenID, "profile"}
+
 func connectToProvider(conf *AuthConf) (*oauth2.Config, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
@@ -38,14 +40,16 @@ func connectToProvider(conf *AuthConf) (*oauth2.Config, error) {
 	if err != nil {
 		return nil, err
 	}
+	scopes := authDefaultScopes
+	if len(conf.Scopes) > 0 && conf.Scopes[0] != "" {
+		scopes = conf.Scopes
+	}
 	return &oauth2.Config{
 		ClientID:     conf.ClientID,
 		ClientSecret: conf.ClientSecret,
 		Endpoint:     provider.Endpoint(),
 		RedirectURL:  conf.RedirectURL,
-		//TODO: make configurable
-		//TODO review: does this return preferred_username if Profile is not included in scope?
-		Scopes: []string{oidc.ScopeOpenID},
+		Scopes:       scopes,
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


lo5 (approved):
Awesome! Thanks @geomodular!

Issue #936, opened 2021-07-26
Port Collision


Derrick56007:
#### Wave SDK Version, OS

v0.17.0

#### Actual behavior

wave apps sharing a wave daemon on kubernetes results in port collision
all the apps use port 8000

#### Expected behavior

The port should increment properly or the user should be able to manually set the host and port variables

PR #937, opened 2021-07-26
check H2O_WAVE_APP_ADDRESS env var before scanning free port

Derrick56007:
fixes https://github.com/h2oai/wave/issues/936

### DIFF
--- a/py/h2o_wave/cli.py
+++ b/py/h2o_wave/cli.py
@@ -18,6 +18,7 @@ from contextlib import closing
 import uvicorn
 import click
 import os
+from urllib.parse import urlparse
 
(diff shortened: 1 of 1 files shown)


neuhausler:
Dear Wave Team .. this one will also get your documentation in line with the implementation. In your documentation https://wave.h2o.ai/docs/configuration you mention that one can use "H2O_WAVE_APP_ADDRESS" to overwrite the address .. but your code at https://github.com/h2oai/wave/blob/master/py/h2o_wave/cli.py#L64 doesn't even look at that variable and just overwrites it with `localhost` and open port found ..  `dos.environ['H2O_WAVE_APP_ADDRESS'] = addr`

PR #960, opened 2021-08-16
feat: Side panel. #898

### LINKED ISSUE #898: Side Panel for Configuration
It is common to use an app to configure based on information from the user. Today, this can be done by clicking a button and going to a new page, or by showing a dialog in the center of the page. Going to a new page can remove the end user mentally from the current flow they are working on, and the center dialog mostly blocks out the page the user is on.

Having a side panel type dialog which would fade the background but allow it to still be seen. 

#### Additional context
image

https://xd.adobe.com/view/a1f57a3b-dfdd-4d3c-ab5f-59d404c96593-f4a6/

mturoci:
Current demo:

https://user-images.githubusercontent.com/64769322/129548308-83eaab7d-58d5-4855-86f5-d300d2d5f3f3.mov

@lo5 
* I added `width` prop as well which I think might be handy - same as dialog.
* Added support for `q.events` on close - same as dialog.
* Needs discussion: It might be a good idea to introduce a `title` prop as well.

@sandruparo 
* As discussed, the designs need more work on colors. They are currently hardcoded and need to be themable instead.
* Current implementation doesn't use any offsets (top, right, bottom). The proposed design has them, but these seem like random values (top 223px, right 50px, bottom 244px). There are 2 options we have here:
1. Make the offset fixed - what is presented in the proposed design, but with reasonable values.
2. Use percentage value.
Please let me know which one to pick.

Closes #898

### DIFF
--- a/py/examples/meta_side_panel.py
+++ b/py/examples/meta_side_panel.py
@@ -14,7 +14,7 @@ async def serve(q: Q):
         q.client.initialized = True
     else:
         if q.args.show_side_panel:
-            q.page['meta'].side_panel = ui.side_panel(items=[
+            q.page['meta'].side_panel = ui.side_panel(title='Welcome to store', items=[
                 ui.text('Donuts cost $1.99. Proceed?'),
                 ui.buttons([ui.button(name='next_step', label='Next', primary=True)])
             ])
--- a/py/examples/meta_side_panel.py
+++ b/py/examples/meta_side_panel.py
@@ -1,0 +1,37 @@
+# Meta / SidePanel
+# Display a #sidePanel. #meta
+# ---
+from h2o_wave import main, app, Q, ui
+
+
+@app('/demo')
+async def serve(q: Q):
+    if not q.client.initialized:
+        q.page['meta'] = ui.meta_card(box='')
+        q.page['example'] = ui.form_card(box='1 1 11 10', items=[
+            ui.button(name='show_side_panel', label='Order donuts', primary=True)
+        ])
+        q.client.initialized = True
+    else:
+        if q.args.show_side_panel:
+            q.page['meta'].side_panel = ui.side_panel(items=[
+                ui.text('Donuts cost $1.99. Proceed?'),
+                ui.buttons([ui.button(name='next_step', label='Next', primary=True)])
+            ])
+        elif q.args.next_step:
+            q.page['meta'].side_panel.items = [
+                ui.text('You will be charged $1.99. Proceed?'),
+                ui.buttons([
+                    ui.button(name='cancel', label='Back to safety'),
+                    ui.button(name='submit', label='Place order', primary=True),
+                ])
+            ]
+        elif q.args.submit:
+            q.page['example'].items = [ui.message_bar('success', 'Order placed!')]
+            q.page['meta'].side_panel = None
+
+        elif q.args.cancel:
+            q.page['example'].items = [ui.message_bar('info', 'Order canceled!')]
+            q.page['meta'].side_panel = None
+
+    await q.page.save()
--- a/py/examples/tour.conf
+++ b/py/examples/tour.conf
@@ -172,6 +172,7 @@ meta_stylesheet.py
 meta_inline_stylesheet.py
 meta_dialog.py
 meta_dialog_closable.py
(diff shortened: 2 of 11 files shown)


lo5:
+1 to introducing a title prop.

> Current implementation doesn't use any offsets (top, right, bottom). The proposed design has them

I'd like to hear a strong reason for introducing offsets @sandruparo. I like the above implementation where the sidebar is flush with the window borders.

mtanco:
From @sandruparo -
The current implementation behaves similar to a navigation side panel and fully covers the header and title components, ideally the header and (optional) title components would be visible while using the side panel so that is is still very clear you are in the app and just doing an additional activity.

lo5:
@sandruparo Not a strong enough reason. The background fade makes it clear that one is still in the app. Also, the side panel is modal, so I would argue that *not* fading the header is confusing, and conveys the impression that it's still operable, which it's not.

@mturoci Let's go with the current implementation.

sandruparo:
@lo5 - I strongly feel that if we expand the side pane like the way Martin did, it may impact to the cognitive load of the screen, no space on the top of the side pane, i dont know, i kind of feel it's bit too much Prithivi. just check the attach designs and see how do you feel ?

sandruparo:
Screenshot 2021-08-31 at 21 55 28

lo5:
The second one not superior to the first one. 
The first one makes more optimal use of space.
The first one is common.
The second one is odd.

Material Design:
> Side sheets are anchored to the left or right edge of the screen. They are a fixed width and typically span the height of the screen. - https://material.io/components/sheets-side

iOS: spans the viewport height - https://designcode.io/ios-design-handbook-design-the-sidebar

More examples: https://dribbble.com/tags/drawer

sandruparo:
image

I'm fine with a design like this - material , they have kept a 50px of a space on the top. @lo5

lo5:
@sandruparo That is not modal. We're discussing modal drawers/sidebars here.

sandruparo:
image

What i have designed is similar to the 2nd example of this.

sandruparo:
image

In this it opens from the top - which is something i actually don't prefer. It's behavior looks like a mobile nav.

lo5:
@sandruparo See my comment above. 
[2] is non-modal
[1] is modal

Your MLOps screen design is clearly a modal overlay, but you're pointing at Material Design's [2], which is non-modal.

sandruparo:
Oh okay Prithivi, i will revisit to this and see how i can adjust this design to the way we need . sorry for the confusion. I should see the alternative ways. thanks. will update you tomorrow on this.

shihan007:
@lo5 @mturoci So we are updating our designs with full height side panel which we all of agreed to go with. But when it's comes to the overlay we recommend to go with modal side panel for these panels because this panel using to enter data and if a user click accidentally click on a background and side panel vanishing is not a good ux to this scenario. If we are only displaying data non modal panel would  be okay to use. Let me know what do you think?

cc: @sandruparo @mtanco

mturoci:
> we recommend to go with modal side panel for these panels because this panel using to enter data and if a user click accidentally click on a background and side panel vanishing is not a good ux to this scenario.

The current implementation is blocking, which means clicking outside the panel will not close it. We can add a flag to toggle this behavior in future if needed. Fluent even makes it easy to have a regular side panel which allows interacting with non-panel elements also, so this covers all the possible cases I guess. The only question is, whether we want to provide different visuals for each mode @lo5. IMO current blocking implementation is gonna be fine for most of the usecases.

shihan007:
@mturoci I am referring to very first example in this https://developer.microsoft.com/en-us/fluentui#/controls/web/panel. Is it possible to implement panel like this or if this takes time we can move with the non modal for now and then later add on this feature.

mturoci:
@shihan007 yes, that's the current implementation

shihan007:
@mturoci we are good to go with this

sandruparo:
> @mturoci I am referring to very first example in this https://developer.microsoft.com/en-us/fluentui#/controls/web/panel. Is it possible to implement panel like this or if this takes time we can move with the non modal for now and then later add on this feature.

looks Great ! @mturoci

PR #973, opened 2021-08-19
feat: Custom inline / external CSS support. #965

### LINKED ISSUE #965: Ability to add custom CSS files
via @mattdowle @lo5

## Proposed API

Add an api similar to `ui.script()` and `ui.inline_script()`, but for CSS.

mturoci:
cc @lo5 for API / examples review.

This PR introduces 2 ways of adding a CSS stylesheet:
* Simply inlining it.
* Loading external stylesheet.

Note that it might make sense to generalize the external stylesheet loading logic so that app devs would have full control over links and could [prefetch](https://developer.mozilla.org/en-US/docs/Web/HTTP/Link_prefetching_FAQ) or [preload](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/preload) the needed resources.

Closes #965


ui/src/meta.tsx
```diff
@@ -113,6 +123,10 @@ interface State {
   scripts?: Script[]
   /** Javascript code to execute on this page. */
   script?: InlineScript
+  /** CSS to be inlined into this page. */
+  style?: S
```
lo5, line 127:
```suggestion
  stylesheet?: InlineStylesheet
```

`InlineStylesheet` (consistent with InlineScript) can contain `content` and `media?` (for consistency with `Stylesheet.media`).

ui/src/meta.tsx
```diff
@@ -113,6 +123,10 @@ interface State {
...
+  /** External CSS files to load into the page. */
+  styles?: Style[]
```
lo5, line 129:
```suggestion
  stylesheets?: Stylesheet[]
```

ui/src/meta.tsx
```diff
@@ -113,6 +123,10 @@ interface State {
   scripts?: Script[]
   /** Javascript code to execute on this page. */
   script?: InlineScript
+  /** CSS to be inlined into this page. */
```
lo5, line 126:
```suggestion
  /** CSS stylesheet to be applied to this page. */
```

lo5 (changes requested):
Minor changes to API. Thanks @mturoci!

mturoci:
Thanks for the review @lo5! Comments addressed.

PR #1013, opened 2021-09-22
feat: Wide info card subtitle + redesign. #967

### LINKED ISSUE #967: Component: Wide Info Card
**Humid Design System:** wide_info_card
**Card Name:** `wide_info_card`

image

**Components:**
* Icon - proportionally-sized icon to the left that does not resize beyond a certain fixed width - required, uses a default like the `ui.header` object
* headline - optional
* title - Required
* subtitle - optional
* content - required

mturoci:
Closes #967

### DIFF
--- a/py/examples/wide_info.py
+++ b/py/examples/wide_info.py
@@ -8,12 +8,13 @@ page = site['/demo']
 page.add('example', ui.wide_info_card(
     box='1 1 4 5',
     title='Info Card',
+    subtitle='Subtitle',
     caption='''
-    Lorem ipsum dolor sit amet, consectetur adipisicing elit. Quia aliquam maxime quos facere
-    necessitatibus tempore eum odio, qui illum. Repellat modi dolor facilis odio ex possimus
-    ratione voluptate pariatur cupiditate, quidem quaerat sapiente exercitationem in omnis nulla
-    maiores consequatur dolores illo inventore quae obcaecati culpa totam corporis! Repudiandae, nostrum repellendus.
-    ''',
+Lorem ipsum dolor sit amet, consectetur adipisicing elit. Quia aliquam maxime quos facere
+necessitatibus tempore eum odio, qui illum. Repellat modi dolor facilis odio ex possimus
+ratione voluptate pariatur cupiditate, quidem quaerat sapiente exercitationem in omnis nulla
+maiores consequatur dolores illo inventore quae obcaecati culpa totam corporis! Repudiandae, nostrum repellendus.
+''',
     category='Category',
     image='https://images.pexels.com/photos/3225517/pexels-photo-3225517.jpeg?auto=compress&cs=tinysrgb&dpr=2&h=750&w=1260' # noqa
 ))
--- a/py/h2o_wave/types.py
+++ b/py/h2o_wave/types.py
@@ -6081,7 +6081,7 @@ class Component:
         self.image = image
         """Image"""
         self.persona = persona
-        """Persona"""
+        """Persona."""
 
     def dump(self) -> Dict:
         """Returns the contents of this object as a dict."""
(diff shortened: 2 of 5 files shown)


lo5:
Looks good:
- API changes
- Implementation with image

Needs design changes:
- When an icon is used, the ratio of icon:text needs to be higher. See Slide 3 in the design feedback Keynote.

lo5:
The intent of this card (and `tall_info_card`) is to "lead-in" into other content (e.g. summary-to-detail, call-to-action, etc.), so it needs to be clickable (either the card itself or via a button). 

This would mean that the card needs a mandatory `name`, with behavior matching `button.name`, and an optional `label`. If `label` is specified, render a button; if not, make the entire card clickable.

I'm assuming that the `wide_info_card` in this PR covers both the "WideInfoCard" and "centered_wide_info_card / Complex_wide_left_img_text" in XD. If so, the "centered_wide_info_card" already has design variations for card-with-button and card-without-button.

mturoci:
Added `label` and `align` props.

https://user-images.githubusercontent.com/64769322/136018578-10d8d1dc-e6df-46c7-910b-fbeb02dc8e39.mov

`align='right'`:

image


waiting for design team to address comment above

shanaka-rajitha:
@lo5 @mturoci card design for `WideInfoCard` and `TallInfoCard` card is updated with relation to ratio of icon:text.

shanaka-rajitha:
Hi @mturoci it's tricky to provide a solid deterministic approach for color selections as we don't know in advance what colors a user will choose.

What we can provide as a solution is to:
- identify UI elements that are contrasting/ less visible etc.
- introduce new color variables to these UI elements eg: icon color and icon foreground.

Using Fluent UI Theme Designer to generate colors would cover some parts of the UI with accessibility issues but not all aspects, so I don't think it's a solid approach.

mturoci:
> It seems foreground color is taking too much attention in light and yellow theme. can we use neutralLight for foreground?

The contrast in neon theme is not enough:
image

> it's tricky to provide a solid deterministic approach for color selections as we don't know in advance what colors a user will choose

Right, not easy, but doable.

>introduce new color variables to these UI elements eg: icon color and icon foreground.

Sounds good, just make sure they are based off the current theme, not hardcoded values.

shihan007:
@mturoci idea is to have attention to foreground like in neon theme. So NeutralLight would good to use as foreground color. We only need to highlight icon and that is contrast enough in every theme

shihan007:
@mturoci It is to make it a consistence UI element because icons width and height changes from icon to icon. Because of this circle it always take same space and consistence on each card

mturoci:
> width and height changes from icon to icon

These are minimal, maybe less than 1px differences as we use font-icons, but if the only issue is space, the wrapper can be transparent. Right now, it looks odd since it is there, but barely visible.

If you want to make the wrapper less prominent, what about going with very light text color (`text1`)? Would look like this:
image
image
image

mturoci:
@shihan007 is it good as it is?

mturoci:
@lo5 wdyt about `background` prop? I vote to remove it since it can break contrast with text easily.

mturoci:
No, I thought this is how you wanted based on designs - align with the bottom edge of the image. Should I put it right after the textual content?

mturoci:
A card can be of any height / width user wishes. That means having hardcoded dimensions (like in the example above) or let the content dictate the dimensions (flex layout).

> For tallinfo card with an icon the spacing should be 16px between content and button. Also, the height of the card should depend on the content.

This is implemented already, just the example uses hardcoded values.

> For tallinfo card with an image, you can align the button to the image at the very bottom. As we don't have control over the image height.

We don't honor image height, we fit the image to the card height.

The goal of all the cards in Wave is to provide a good experience experience for both fixed dimensions and content (flex) dimensions.

So both of the comments are already addressed.

PR #1031, opened 2021-09-30
feat: WidePlotCard. #1030

### LINKED ISSUE #1030: WidePlot card
design url: https://xd.adobe.com/view/119e73c2-d952-4760-8202-d36d3fc16016-44cd/specs/

image

## Proposed API:
image

mturoci:
image
image
image

## Proposed API

```ts
/** Create a wide plot card displaying a title, caption and a plot. */
interface State {
  /** The card's title. */
  title: S
  /** The card's caption, displayed below the title. */
  caption: S
  /** The card's plot. */
  plot: Plot
  /** The card's plot data. */
  data: Rec
}
```

possible to also add `events` if needed.

## Design questions

* What is the preferred text container width? I went with `200px` and let the plot take the rest of the space. We can also go for percentual ratio or something else.
* The space between text and plot seems to be too small (`8px` as per design)

Closes #1030

### DIFF
--- a/py/examples/tour.conf
+++ b/py/examples/tour.conf
@@ -4,6 +4,7 @@ wizard.py
 issue_tracker.py
 dashboard.py
 wide_info.py
+wide_plot.py
 stat_small.py
 stat_large.py
 stat_wide_gauge.py
--- a/py/examples/wide_plot.py
+++ b/py/examples/wide_plot.py
@@ -1,0 +1,23 @@
+# Plot / Wide
+# Create a wide information card displaying a plot with title and caption.
+# ---
+from h2o_wave import site, ui, data
+from synth import FakeCategoricalSeries
+
+page = site['/demo']
+
+f = FakeCategoricalSeries()
+n = 10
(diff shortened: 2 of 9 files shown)


shanaka-rajitha:
>What is the preferred text container width? I went with 200px and let the plot take the rest of the space. We can also go for percentual ratio or something else.

Preferred with is **274px**( 35% ) for text container and **454px**( 65% ) of the graph. The main container should be maximum **792px** as this is will add up to half the page width. 


>The space between text and plot seems to be too small (8px as per design)

Lets increase it to **16px** as the padding.

CC @mturoci

mturoci:
> Preferred with is **274px**( 35% ) for text container and **454px**( 65% ) of the graph. The main container should be maximum **792px** as this is will add up to half the page width. 

The card can be any width / height so will go with percents 35:65 if that's fine.

> Lets increase it to 16px as the padding.

Will do.

sandruparo:
cc - @dulajra @VijithaEkanayake

lo5 (approved):
Looks great!

PR #1039, opened 2021-10-04
feat/issue 1033

### LINKED ISSUE #1033: Profile card
image

design url: https://xd.adobe.com/view/acfb7ef1-5b99-4129-97b2-b30f8086c239-f9a8/

## Proposed API
image

mturoci:
Closes #1033


shihan007:
@mturoci in light theme name and profile pic space seems reduced, dark theme looks good. For pic border color which color variable did you use?

mturoci:
> in light theme name and profile pic space seems reduced

my bad, accidentally took a screenshot of wrong code

image
image

> For pic border color which color variable did you use?
neutralPrimary

shihan007:
@mturoci For border color lets go with neutralTertiary. Hard to find a variable that matches with each theme.
Fine with the pro pic size and font size.

mturoci:
> Hard to find a variable that matches with each theme.

Feel free to come up with your own color computation if you feel like the Fluent one is too hard to work with.

> Fine with the pro pic size and font size.

cool, thanks

What about the rest of the points?

shihan007:
@mturoci since we are using most icons in here we can go with 24px button as we did in the other cards.

lo5:
@mturoci - How about this:

```typescript
/** Create a profile card to display information about a user. */
interface State {
  /** The persona represented by this card. */
  persona: Persona
  /** 
   * The card’s background image, either a base64-encoded image, a path to an image hosted externally (starting with `https://` or `http://`)
   * or a path to an image hosted on the Wave daemon (starting with `/`).
. */
  image?: S
  /** Components displayed inside this card. */
  items?: Component[]
}
```

Note:
- Can rename to `persona_card` (since we've already shipped `ui.persona()`).
- `caption`, `size`, and `name` from `persona` can be ignored by this card's implementation.

mturoci:
@lo5 yes, this was my first go to, but then realized that `ui.persona` returns `Component`.

> Can rename to persona_card (since we've already shipped ui.persona())

Good point, will do.

lo5:
> yes, this was my first go to, but then realized that ui.persona returns Component,

Should be OK, since call sites are not affected:

```py
card = ui.persona_card(persona=ui.persona(...), image='...', items=[...])
```

mturoci:
> Should be OK, since call sites are not affected:

Not sure what you mean. This is what I get:

image

with code:

```py
from h2o_wave import site, ui

page = site['/demo']

image = 'https://images.pexels.com/photos/3225517/pexels-photo-3225517.jpeg?auto=compress&cs=tinysrgb&dpr=2&h=750&w=1260' # noqa
page.add('example', ui.profile_card(
    box='1 1 3 5',
    persona=ui.persona(title='John Doe', subtitle='Data Scientist', caption='Online', size='xs', image=image),
    commands=[
        ui.command(name='new', label='New', icon='Add', items=[
            ui.command(name='email', label='Email Message', icon='Mail'),
            ui.command(name='calendar', label='Calendar Event', icon='Calendar'),
        ]),
        ui.command(name='upload', label='Upload', icon='Upload'),
        ui.command(name='share', label='Share', icon='Share'),
        ui.command(name='download', label='Download', icon='Download'),
    ],
    items=[
        ui.inline(justify='center', items=[
            ui.button(name='btn1', label='Button 1'),
            ui.button(name='btn2', label='Button 2'),
            ui.button(name='btn3', label='Button 3'),
        ]),
    ]
))

page.save()
```

lo5:
By "call sites are not affected", I meant that it makes no difference to the following line whether `ui.persona()` returns a `Component` or a `Persona`. 

```
card = ui.persona_card(persona=ui.persona(...), image='...', items=[...])
```

This is not special to `persona_card` or `persona`. All the existing components work this way (e.g. a `button` inside `buttons`). 

> Maybe we could adjust wavegen in a way that it would decide whether to return wrapped persona within component or just a simple persona based on some kind of Python reflection that would search the stacktrace for form_card occurence?

It's not urgent right now, and the persona card will work just fine without any modifications to wavegen. 

A Persona is modeled as a variant of Component, not a subclass (Component is a union type). I understand it's possible to pass in the wrong kind of component, but Python doesn't provide safety at runtime anyway. `Union[Component, Persona]` wouldn't be a proper solution to this, because it doesn't solve the primary problem of preventing the wrong kind of component from being passed in.

So I would recommend proceeding without changes to wavegen.

PR #1042, opened 2021-10-05
feat: WidePieStat card. #1032

### LINKED ISSUE #1032: WidePieStat card
image

design url: https://xd.adobe.com/view/acfb7ef1-5b99-4129-97b2-b30f8086c239-f9a8/

## Proposed API
image

mturoci:
Closes #1032


shihan007:
@mturoci can we match the bg color to stroke color

mturoci:
Yes, we can use whatever color we like. My question is about full vs inner-only stroke though.

shanaka-rajitha:
@mturoci let's keep it only to have the inner-only stroke for consistency.

mturoci:
@shanaka-rajitha @shihan007 please read my question. I am asking whether the inner-stroke is important / required. If so, why.

shihan007:
@mturoci when there are multiple shades in same color easy to separate and identify each section. That why we include that

mturoci:
Having multiple shades of the same color defeats the purpose of whole pie chart as it makes legend unusable.

shihan007:
@mturoci If we are using different color yeah we can use the stroke

mturoci:
This is how the stroke would look like with card background color.

image

wdyt?

shihan007:
@mturoci yeah let's go with no stroke

mturoci:
cc @lo5 for design review before merge.

Issue #59, opened 2020-08-25
Allow prefixes to URL


codyharris-h2o-ai:
Currently, when an app gets started, we expect that users will access the app via a path like /app/<uuid>/ .
We’re rewriting the index path to update the references for the JS & CSS resources (ie, everything that comes from /static in Qd) to point to /app/<uuid>/static, and we proxy that directory to Qd.  For the JS resources, we’re rewriting the websocket path from /_s to /app/<uuid>/ws.  We’re also translating the websocket messages to rewrite /app/<uuid> to the route the app is currently listening on.  We’re planning on adding a field to the toml file that allows developers to specify the route they’re using so we can do the correct proxy.  We also rewrite /_f references to /app/<uuid>/_f.
It’s not ideal for us, particularly since the string sub can translate innocent code that breaks the page (we're already seeing this in some situations).  We were wondering if it may be possible to have Qd know that it’s being loaded from a different path (using something like https://muffinman.io/react-router-subfolder-on-server/) that would allow us to avoid doing the translation.
Another option would be to serve apps from a subdomain, ie, <uuid>.q8s.h2o.ai, and then it’s like it’s running from local.  The concern is that this makes deployment on prem and local testing more difficult (but not impossible)

What we'd like to be able to do:
Allow the browser to fetch `/app/<uuid>`, and have the app function, and all we need to do is proxy `/app/<uuid>` to the container without any rewrites.

The HTML base page for the react app has a hardcoded path for `/static` where it loads the JS resources from, and the JS resources contain a hardcoded path to the websocket (`/_s`).  Once the websocket connection is established, the browser sends its location to Qd.  We can pass this path ahead of time as an environment variable when we launch the container so that the Qd knows to expect it and allow the app to load on the different path.

Apps might need to know about the prefix as well (ie, if they create a link to somewhere else in the app, they may need to know), so an API may be needed for this purpose.

We're open to other suggestions as well

mturoci:
A few points that arose after discussion with @keznikl:

* The auth paths (and their redirects) need to respect base URL as well, e.g. `/base-url/auth`. It currently returns 404.
![image](https://user-images.githubusercontent.com/64769322/144418584-91728fd7-800b-44ff-ad52-0b0f36ba4f9b.png)
* Minor: Assume leading `/` if not specified by the user, otherwise results in a python error:
![image](https://user-images.githubusercontent.com/64769322/144418807-e148afca-2ee8-4100-9b7a-18a2a29cbc6c.png)
[...]

PR #1092, opened 2021-10-27
feat: make all asset paths relative #59

mturoci:
All resources are loaded relative to the root document URL in order to allow path prefixes.
These are the necessary FE changes.

## Assumptions

It's assumed that only https://github.com/h2oai/wave/blob/424c77ab770d062ca99e5544b02a63e4d8cf791b/server.go#L109 will be affected by the prefix (let me know if I am wrong), so [file upload](https://github.com/h2oai/wave/blob/424c77ab770d062ca99e5544b02a63e4d8cf791b/server.go#L78) and [socket server](https://github.com/h2oai/wave/blob/424c77ab770d062ca99e5544b02a63e4d8cf791b/server.go#L75) can remain the same.

## Needs discussion

The aim of this PR is to allow path prefixes, this will break [initial websocket slug payload](https://github.com/h2oai/wave/blob/424c77ab770d062ca99e5544b02a63e4d8cf791b/ts/index.ts#L817) since `pathname` can now consist of `subpath` + `site/app` route. I can think of 2 solutions:
1. Constrain site route annotation to not include `/`.
2. Parse route payload within Go server and update (strip prefix) accordingly since the server will be able to tell the boundary between subpath and site/app path.

Second option seems better IMO.

### DIFF
--- a/ui/config/paths.js
+++ b/ui/config/paths.js
@@ -1,25 +1,12 @@
-'use strict';
+'use strict'
 
-const path = require('path');
-const fs = require('fs');
-const getPublicUrlOrPath = require('react-dev-utils/getPublicUrlOrPath');
+const path = require('path')
+const fs = require('fs')
 
 // Make sure any symlinks in the project folder are resolved:
 // https://github.com/facebook/create-react-app/issues/637
-const appDirectory = fs.realpathSync(process.cwd());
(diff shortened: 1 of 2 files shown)


mturoci:
Closing since this PR was used as a base for #1104.

PR #1104, opened 2021-11-03
Add base URL support

### LINKED ISSUE #59: Allow prefixes to URL

lo5:
Launch `waved` with `-base-url` to set a base URL. Or, launch `waved` with `$H2O_WAVE_BASE_URL` set.

Setting base url to `/foo/` will allow serving an app with route `/` at `/foo/`, or an app with route `/bar` at `/foo/bar/`.

Accessing `/foo/`'s parent(s) from browser/API will 404.

If a base url is set, an app must be launched with `$H2O_WAVE_BASE_URL` set to the same base url value.

Closes #59


web_server.go
```diff
@@ -42,14 +47,31 @@ func newWebServer(
...
+
+	fs := handleStatic([]byte(mungeIndexPage(baseURL, string(indexPage))), http.StripPrefix(baseURL, http.FileServer(http.Dir(webDir))), header)
 	if auth != nil {
 		fs = auth.wrap(fs)
 	}
-	return &WebServer{site, broker, fs, keychain, maxRequestSize}
+	return &WebServer{site, broker, fs, keychain, maxRequestSize, baseURL}, nil
+}
+
+func mungeIndexPage(baseURL, html string) string {
```
mturoci, line 68: My original intention was to use relative frontend paths and just rewrite / strip path prefixes server side. So for example ws connection (same for file API endpoint, static resources) would connect to `_s` instead of absolute `/_s`. Assuming browser url would be `http:localhost:10101/prefix/app-route`, the final ws connection string would be `http:localhost:10101/prefix/app-route/_s`. Server would be then able to strip the prefix as it is aware of `base-url` and `app-route` and correctly route it to `/_s` handler.

This would save us from munging `index.html` and keep frontend part simple.

wdyt @lo5 ?

lo5: The go routing mechanism expects rooted paths, though. So a page accessed at /a/b/c would attempt to connect to /a/b/c/_s, which will not hit the websocket handler.

lo5: But your strategy could be made to work if we bypass go's routing altogether and implement a custom routing mechanism.

PR #1122, opened 2021-11-11
feat: Footer improvements #1119

### LINKED ISSUE #1119: Footer improvements
image

[design url](https://xd.adobe.com/view/acfb7ef1-5b99-4129-97b2-b30f8086c239-f9a8/)

## Proposed API

```ts
/** Create a footer link leading to other internal / external pages. */
interface FooterLink {
  /** The text to be displayed. If blank, the `path` is used as the label. */
  label?: S
  /** The path or URL to link to. */
  path?: S
  /** Where to display the link. Setting this to an empty string or `'_blank'` opens the link in a new tab or window. */
  target?: S
  /** An identifying name for this component. Use this for Wave related routing. Mutually exclusive with path. */
  name?: S
  /** An icon to display next to the label. */
  icon?: S
}

/** Create a collection of footer links. */
interface FooterLinkGroup {
  /** The name of the link group. */
  label: S
  /** The link groups contained in the header. */
  items: FooterLink[]
}

/**
 * Render a page footer displaying a caption.
 * Footer cards are typically displayed at the bottom of a page.
 */
interface State {
  /** The caption. Supports markdown. **/
  caption: S
  /** The card’s image, displayed in the top left corner. */
  image?: S
  /** The navigation groups displayed to the right of the image. */
  items: FooterLinkGroup[]
}
```

mturoci:
image
image
image

## Notable changes

* Added a new `items` prop to allow customizing rhs of the footer.
* If `items` is not specified, the markdown caption is centered by default - most of the people try to center it anyway - cc @mtanco.
* Images in markdown don't have width/height constrained, but rather their intrinsic size is used - this way we force people to use properly sized images and avoid potential perf issues.
* Added `inline` prop that allows rendering links horizontally - useful for a header for example. Using `ui.inline` would work too, but the gap between particular elements is not enough (links too close).

## Final API

```ts

/** Create a collection of links. */
export interface Links {
  /** The links contained in this group. */
  items: Component[]
  /** The name of the link group. */
  label?: S
  /** Render links horizontally. Defaults to 'false'. */
  inline?: B
}

/**
 * Render a page footer displaying a caption.
 * Footer cards are typically displayed at the bottom of a page.
 */
interface State {
  /** The caption. Supports markdown. **/
  caption: S
  /** The components displayed to the right of the caption. */
  items?: Component[]
}
```

@lo5 for final review ^^

Closes #1119


lo5:
API and changes look good.

The link columns would look neater if they had the same widths. Is there a way to do that? Adding a `width=` to `Links` is one option, but wondering if there are better ways.

mturoci:
>The link columns would look neater if they had the same widths. Is there a way to do that? Adding a width= to Links is one option, but wondering if there are better ways.

Agree. With Flexbox, it's possible to distribute the remaining free space to its children like this:

image

However, what we want here is that all children have the width of the widest child.

Another option is to use the existing `width` attr of any link in `ui.links` and that will automatically expand the whole links section but is not clear APIwise.

IMO introducing a `width` prop is a good idea.

PR #1123, opened 2021-11-11
fix: correctly display selected day in datepicker #1098

### LINKED ISSUE #1098: Date Picker UI off by 1 Value
#### Wave SDK Version, OS
Wave 0.18.0, tested running in OS X and Linux (AI Cloud), safari & chrome

#### Actual behavior
When selecting a date using the `ui.date_picker` component, the UI shows the day _before_ whatever is selected. The value inside of `q.args.date_picker_name` is correct meaning the end user can work around this by remaking the card with the selected value.

#### Expected behavior
The date clicked is the date showed in the UI

#### Steps To Reproduce
Run this app with the 0.18.0 Server and API. Then, select any date (say October 5th), note that the day before (October 4th) is shown. Click the button and see that the expected date is then shown (as October 5th, not 4th, is actually saved, the wrong date is just being shown).

```python


@app('/')
async def serve(q: Q):

    if q.args.save:
        
        q.page["form"] = ui.form_card(
            box="4 1 4 2",
            items=[
                ui.date_picker(name='start_date', label='Start Date', value=q.args.start_date),
                ui.button(name='save', label='Save Date', primary=True),
            ]
        )

    else:
        q.page["form"] = ui.form_card(
            box="4 1 4 2",
            items=[
                ui.date_picker(name='start_date', label='Start Date'),
                ui.button(name='save', label='Save Date', primary=True),
            ]
        )
    await q.page.save()
```

aalencar:
* Fixes timezone bug when displaying the day in the Datepicker component

closes #1098

### DIFF
--- a/ui/src/date_picker.tsx
+++ b/ui/src/date_picker.tsx
@@ -68,7 +68,7 @@ export const
       onSelectDate = (d: Date | null | undefined) => {
         const val = (d === null || d === undefined) ? defaultVal : formatDate(d)
         wave.args[m.name] = val
-        setValue(val ? new Date(val) : undefined)
+        setValue(val ? new Date(`${val} 00:00:00`) : undefined)
         if (m.trigger) wave.push()
       }
 


ui/src/date_picker.tsx
```diff
@@ -68,7 +68,7 @@ export const
       onSelectDate = (d: Date | null | undefined) => {
         const val = (d === null || d === undefined) ? defaultVal : formatDate(d)
```
mturoci, line 69: nit: I know this is not part of the issue, but can you please refactor this? Thanks.
```suggestion
        const val = d ?? defaultVal : formatDate(d)
```

mturoci (approved):
Thanks @aalencar, looks good!

PR #1129, opened 2021-11-15
feat: Sidebar improvements

### LINKED ISSUE #1118: Sidebar improvements

Wave currently supports only header where title, subtitle, and menu can be specified. However, some apps may want to use side nav menu only, but still keep the branding.

Allow people to use a layout like:
image
image

#### Proposed API

There are 2 options, either go with a new `ui.side_nav` card or enhance the existing `ui.nav` card.

```ts
/** Create a card containing a navigation pane. */
export interface State {
  /** The navigation groups contained in this pane. */
  items: NavGroup[]
  /** The name of the active (highlighted) navigation item. */
  value?: S
  /** Items that should be displayed at the bottom of the card. */
  secondary_items?: Component[]
  /** The card's title. */
  title?: S
  /** The card's subtitle. */
  subtitle?: S
  /** The logo displayed at the top. **/
  image?: S
  /** The user avatar displayed at the top. Mutually exclusive with image, title and subtitle. **/
  persona?: Component
  /** Card background color. Defaults to 'card'. */
  color?: 'card' | 'primary
}
```
Extending the existing card makes probably more sense to me, wdyt @lo5?

mturoci:
image
image
image
image

## Proposed API

```ts
/** Create a card containing a navigation pane. */
export interface State {
  /** The navigation groups contained in this pane. */
  items: NavGroup[]
  /** The name of the active (highlighted) navigation item. */
  value?: S
  /** The card's title. */
  title?: S
  /** The card's subtitle. */
  subtitle?: S
  /** The icon, displayed to the left. **/
  icon?: S
  /** The icon's color. **/
  icon_color?: S
  /** The logo displayed at the top. **/
  image?: S
  /** The user avatar displayed at the top. Mutually exclusive with image, title and subtitle. **/
  persona?: Component
  /** Items that should be displayed at the bottom of the card if items are not empty, otherwise displayed under subtitle. */
  secondary_items?: Component[]
  /** Card background color. Defaults to 'card'. */
  color?: 'card' | 'primary'
}
```

cc @lo5 for design review

Closes #1118

### DIFF
--- a/py/examples/header.py
+++ b/py/examples/header.py
@@ -8,7 +8,6 @@ commands = [
     ui.command(name='profile', label='Profile', icon='Contact'),
     ui.command(name='preferences', label='Preferences', icon='Settings'),
     ui.command(name='logout', label='Logout', icon='SignOut'),
-
 ]
 page = site['/demo']
 page['header1'] = ui.header_card(
--- a/py/examples/nav.py
+++ b/py/examples/nav.py
@@ -4,6 +4,9 @@
 from h2o_wave import main, app, Q, ui
 
 
+persona = 'https://images.pexels.com/photos/220453/pexels-photo-220453.jpeg?auto=compress&h=750&w=1260'
+
+
 @app('/demo')
 async def serve(q: Q):
     if '#' in q.args and not q.args.show_nav:
@@ -14,7 +17,7 @@ async def serve(q: Q):
             ui.button(name='show_nav', label='Back', primary=True),
         ])
     else:
-        q.page['meta'] = ui.meta_card(box='', redirect='#')
+        q.page['meta'] = ui.meta_card(box='', redirect='#', theme='h2o-dark')
         q.page['nav1'] = ui.nav_card(
             box='1 1 2 -1',
             value='#menu/spam',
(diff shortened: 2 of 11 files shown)


shihan007:
Screenshot 2021-11-15 at 20 17 00i
@mturoc This layout is better. but logo takes too much space. it would be better if we can reduce the logo size so that we can show more menu items. 
If we can reduce the logo size and left align with title and subtitle would be great.

@sandruparo @shanaka-rajitha wdyt?

lo5:
API and redesign looks great! 

Agree with @shihan007's comment above.

shihan007:
@mturoci @lo5 btw it would be great if we can display logo, title and logged person in top navigation. Like we have added in the design. So we have space left on side navigation for menu items.
Screenshot 2021-11-16 at 10 28 30

cc: @sandruparo @shanaka-rajitha

mturoci:
>@mturoc This layout is better. but the logo takes too much space. it would be better if we can reduce the logo size so that we can show more menu items.

Yes, this depends on the size of the logo you upload, you can upload a smaller logo if you need more space. I only added top boundaries - `max-height: 150px`, but you are right, it can be too much, will go with `100px` instead.

> reduce the logo size and left align with title and subtitle would be great.

Not sure if this looks better
image
than
image

mturoci:
>btw it would be great if we can display logo, title and logged person in top navigation

Already done, see https://github.com/h2oai/wave/pull/378.

shihan007:
@mturoci can set the logo size to 64px and middle align title with it ?  
If we are displaying  logo and profile in top nav do we need to display again in side nav too? 

cc: @lo5

mturoci:
>can set the logo size to 64px

You can set it to whatever size you want.

> middle align title with it 

So back to the original implementation, right?

>If we are displaying logo and profile in top nav do we need to display again in side nav too?

No, it's unnecessary to display it in both side and top, better to pick just one.

shihan007:
> You can set it to whatever size you want.

From system lets set the size to 40px ( I think 40px is enough) .  And align text like this. So every logo have a standard size of 40px
Screenshot 2021-11-16 at 17 05 26

mturoci:
>From system lets set the size to 40px ( I think 40px is enough) . And align text like this. So every logo have a standard size of 40px

If an app dev wants it this way, topnav can be used, no need to force it for sidenav as well.

shihan007:
@mturoci It's better to use same sizing for both top nav and side nav logos in consistency wise. By doing that we can save up space in the side nav to display more menu item which is the primary goal

mturoci:
Why would I display the same logo twice in a single app?

mturoci:
> By doing that we can save up space in the side nav to display more menu items which is the primary goal

* If you have plenty of nav items - use top nav with logo + sidebar without logo only with links.
* If you have just a few links, use just one. Either sidenav or top nav.

Design in this PR is suited for low content apps (majority of apps) so it's better to use the available space rather than have a lot of whitespace which doesn't look good.

shihan007:
> Why would I display the same logo twice in a single app?

No not the same logo, I meant sizing

> * If you have plenty of nav items - use top nav with logo + sidebar without logo only with links.
> * If you have just a few links, use just one. Either sidenav or top nav.
> 
> Design in this PR is suited for low content apps (majority of apps) so it's better to use the available space rather than have a lot of whitespace which doesn't look good.

Apps like mlops have more menu items. Even though app have one or two many item it doesnt matter normally those menu items aligned to top and we can keep rest white space.

PR #1130, opened 2021-11-18
feat: add preview card #1071

### LINKED ISSUE #1071: Preview card
image
image
image
image
image

@lo5 all these cards are marked as `preview_card`, do you want to go with a single one or introduce some kind of `align` attr? I vote for going with a single one that makes the most sense, but open to other ideas.

aalencar:
* Create a preview card displaying an image with shadow overlay, title, social icons, caption, and button.

### Design

I implemented only the first variation of the preview card

image

### Implemented

h2o-dark
preview-card

Neon
preview-card

Default (broken)

preview-card

@mturoci any idea on what we should do for the default theme? Maybe a light overlay shadow?

### API

Here is the proposed API
```typescript
/** Create a preview card displaying an image with shadow overlay, title, social icons, caption, and button. */
interface State {
  /** The card’s image. */
  image: S
  /** The card's title */
  title?: S
  /** Social icons displayed at the top-right corner */
  social?: Component[]
  /** The card's caption, displayed bellow the title. */
  caption?: S
  /** The card's buttons, displayed at the bottom. */
  items?: Component[]
}
```

 closes #1071


mturoci:
>any idea on what we should do for the default theme? Maybe a light overlay shadow?

Yes, the overlay has to be themable as well, not hardcoded.

The design needs work
* it's odd to have partial overlays like this 
image
* the title looks almost identical to caption

Please update the PR description with a proposed API so that we don't need to browse your changes.

aalencar:
@mturoci, I applied the suggestions. That's how it's looking now:

Default (the icons might need some work)


H2O-dark


Neon

mturoci:
Looks better already @aalencar!

* The social icons need contrast fixed.
* Rename `social` to `secondary_items` (tip: running `make generate` outputs all the type names already used so ideally you always pick one of those)
* make sure you test all the edge cases. For example, does the overlay expand if the caption will be longer (takes up the whole card)?

aalencar:
> The social icons need contrast fixed

What does it mean to have them with contrast fixed? Do I have to do something about it or leave it as is?

Working on the other suggestions.

mturoci:
>What does it mean to have them with contrast fixed?

The contrast is not enough and needs to be corrected (fixed).

ui/src/preview.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { Component, XComponents } from './form'
+import { cards } from './layout'
+import { clas, cssVar, cssVarValue } from './theme'
+import { bond, wave } from './ui'
+import * as Fluent from '@fluentui/react'
+
+const css = stylesheet({
+  card: {
+    borderRadius: 4,
+  },
```
mturoci, line 27:
```suggestion
```

ui/src/preview.tsx
```diff
@@ -0,0 +1,117 @@
...
+      wave.push()
+    },
+    getGradient = () => {
+      const { r, g, b } = Fluent.getColorFromString(cssVarValue('$card')) || { r: 255, g: 255, b: 255 }
+      return `linear-gradient(rgba(${r}, ${g}, ${b}, .8) 50%, rgba(${r}, ${g}, ${b}, 0) 100%)`
+    }
+
+    return (
+      <div data-test={name} className={css.card}>
+        <div className={css.img} style={{ backgroundImage: `url('${image}')` }}>
```
mturoci, line 100: Also rename css.img to css.card.
```suggestion
      <div data-test={name} className={css.img} style={{ backgroundImage: `url('${image}')` }}>
```

ui/src/preview.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+const css = stylesheet({
+  card: {
+    borderRadius: 4,
+  },
+  img: {
+    display: 'flex',
+    flexDirection: 'column',
+    justifyContent: 'space-between',
+    height: '100%',
```
mturoci, line 32: Is this needed?

ui/src/preview.tsx
```diff
@@ -0,0 +1,117 @@
...
+    boxSizing: 'border-box',
```
mturoci, line 33: Is this needed?

ui/src/preview.tsx
```diff
@@ -0,0 +1,117 @@
...
+    backgroundSize: 'cover',
+    backgroundRepeat: 'no-repeat',
+    backgroundPosition: 'center',
+    borderRadius: 4,
```
mturoci, line 37: Wave currently doesn't use rounded corners for cards.
```suggestion
```

mturoci (changes requested):
Thanks @aalencar.

* When I specify longer text, a scrollbar appears despite the fact there is plenty of free space.
image
* The examples should use the default theme.
* The code in examples should be formatted. Use either Pycharm or https://marketplace.visualstudio.com/items?itemName=ms-python.python if you want to stay within VSCode.
* Prefer text colors for text - https://github.com/h2oai/wave/blob/master/ui/src/index.scss#L161-L163
* Follow the same clickability rules as in https://github.com/h2oai/wave/pull/1131 to keep consistency.

ui/src/preview.tsx
aalencar: Not after removing the `div` for the image

ui/src/preview.tsx
aalencar: Not after removing the div for the image

ui/src/preview.tsx
```diff
@@ -0,0 +1,117 @@
...
+        window.location.hash = stateName.substr(1)
+        return
+      }
+      wave.args[stateName] = stateName
+      wave.push()
+    },
+    getGradient = () => {
+      const { r, g, b } = Fluent.getColorFromString(cssVarValue('$card')) || { r: 255, g: 255, b: 255 }
+      return `linear-gradient(rgba(${r}, ${g}, ${b}, .8) 50%, rgba(${r}, ${g}, ${b}, 0) 100%)`
+    }
```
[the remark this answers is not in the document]
aalencar: I had moved it to `theme.ts` but after some changes and adding parameters I noticed that they were becoming too different.

This is how it would look like if I would use:

`tall_articles_preview`'s


The above gradient in line 95 (with 2 color stops, transparency `0.8` for the first color stop)


In order to avoid regression, I decided to have one for Preview Card.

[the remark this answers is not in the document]
aalencar: Open to suggestions

aalencar:
@mturoci, thanks for the thorough review! I applied all suggestions but the one to move the gradient function.

aalencar:
Removed the the linear gradient and added box shadow to create a gradient-like effect. Thanks @mturoci for the suggestion.

ui/src/preview.tsx
mturoci: Not resolved.

ui/src/preview.tsx
```diff
@@ -0,0 +1,119 @@
...
+    cursor: 'pointer'
+  },
+  header: {
+    display: 'flex',
+    justifyContent: 'space-between',
+    alignItems: 'center',
+    marginBottom: 8,
+  },
+  content: {
+    padding: 24,
```
mturoci, line 44: The bottom padding should be 0 so that there is no extra overlay.

ui/src/preview.tsx
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    return (
+      <div
+        data-test={name}
+        onClick={stateName && !label ? onClick : undefined}
+        className={clas(css.card, stateName && !label ? css.clickable : '')}
+        style={{ backgroundImage: `url('${image}')` }}
+      >
+        <div className={css.content} style={{ background: `${getOverlay()}`, boxShadow: getBoxShadow(), }}>
```
mturoci, line 104:
```suggestion
        <div className={css.content} style={{ background: cssVar('$card7'), boxShadow: `0px 6px 30px 30px ${cssVar('$card7')}`  }}>
```

ui/src/preview.tsx
```diff
@@ -0,0 +1,119 @@
...
+  header: {
+    display: 'flex',
+    justifyContent: 'space-between',
+    alignItems: 'center',
+    marginBottom: 8,
+  },
+  content: {
+    padding: 24,
+    color: cssVar('$text'),
+    overflowY: 'hidden',
```
mturoci, line 46: What is this for?

mturoci (changes requested):
Thanks @aalencar! A few more comments.

Also please hit the **re-request the review** button when you are finished so that I am notified for the next review round.

ui/src/preview.tsx
aalencar: I forgot about it. Good catch!

ui/src/preview.tsx
aalencar: This is unnecessary. Removed.

ui/src/preview.tsx
aalencar: My mistake, now removed on the last pushed commit.

aalencar:
I didn't know about the "re-request review" feature. I will be using from now on.

aalencar:
Pushed again fixing compilation error and missing fixes

mturoci (approved):
Thanks @aalencar! LGTM.

PR #1131, opened 2021-11-18
fix: make tall / wide info cards clickable based on name parameter #1117

### LINKED ISSUE #1117: Allow wide / tall info card to be non-clickable
via @5675sp

Currently, both cards are clickable by default if `label` is not specified.

The goal of this issue is to make the whole card clickable only if the name is not empty.

aalencar:
closes #1117

### DIFF
--- a/ui/src/tall_info.test.tsx
+++ b/ui/src/tall_info.test.tsx
@@ -82,7 +82,7 @@ describe('TallInfo.tsx', () => {
     expect(wave.args[name]).toBe(name)
   })
 
-  it('Makes card unclickable if name was not provided', () => {
+  it('Makes card unclickable if name is empty', () => {
     tallInfoProps.state.name = ''
     const { getByTestId } = render(<View {...tallInfoProps} />)
     fireEvent.click(getByTestId(name))
--- a/ui/src/tall_info.test.tsx
+++ b/ui/src/tall_info.test.tsx
@@ -17,11 +17,12 @@ import { wave } from './ui'
 import React from 'react'
 import { View } from './tall_info'
 import { box, Model } from 'h2o-wave'
+import { State } from './tall_info'
 
 const
   name = 'tall_info',
   pushMock = jest.fn()
-let tallInfoProps: Model<any>
+let tallInfoProps: Model<State>
 
 describe('TallInfo.tsx', () => {
   beforeAll(() => wave.push = pushMock)
@@ -30,7 +31,7 @@ describe('TallInfo.tsx', () => {
     wave.args = [] as any
     tallInfoProps = {
       name,
-      state: { title: name },
+      state: { title: name, caption: '', name },
       changed: box(false)
     }
   })
@@ -51,9 +52,9 @@ describe('TallInfo.tsx', () => {
 
   it('Submits data to server if label specified and name without #', () => {
     tallInfoProps.state.name = name
-    tallInfoProps.state.label = 'label'
+    tallInfoProps.state.label = 'Click me'
     const { getByTestId } = render(<View {...tallInfoProps} />)
-    fireEvent.click(getByTestId(name))
+    fireEvent.click(getByTestId('Click me'))
     expect(pushMock).toHaveBeenCalled()
     expect(wave.args[name]).toBe(name)
   })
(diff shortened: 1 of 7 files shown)


mturoci (changes requested):
Thanks @aalencar!

* When the name is not specified, the card shouldn't be clickable, but I still get the cursor: pointer on hover.
* If the label is specified, only the button should be clickable, not the whole card. 
* Add some unit tests to make sure every click scenario is covered.

aalencar:
Fixed and tests added

ui/src/tall_info.tsx
```diff
@@ -113,7 +113,7 @@ export const View = bond(({ name, state, changed }: Model<State>) => {
             {category && <div className={clas('wave-s14 wave-w4 wave-t5', css.category)}>{category}</div>}
           </div>
           {caption && <div className='wave-s14 wave-w4 wave-t7'><Markdown source={caption} /></div>}
-          {label && <Fluent.DefaultButton data-test={name} text={label} styles={{ root: { marginTop: 16 } }} onClick={onClick} />}
+          {label && <Fluent.DefaultButton data-test={label} text={label} styles={{ root: { marginTop: 16 } }} onClick={onClick} />}
```
mturoci, line 116: Let's apply `data-test` consistently, on card wrapper only.

```suggestion
          {label && <Fluent.DefaultButton text={label} styles={{ root: { marginTop: 16 } }} onClick={onClick} />}
```

mturoci (changes requested):
Thanks @aalencar, please also fix wide info card in the same way as well.

ui/src/tall_info.tsx
aalencar: Good catch, I forgot about the other component.

ui/src/tall_info.test.tsx
```diff
@@ -73,4 +74,31 @@ describe('TallInfo.tsx', () => {
...
   })
+
+  it('Makes card clickable only if name was specified but label was not', () => {
+    const { getByTestId } = render(<View {...tallInfoProps} />)
+    fireEvent.click(getByTestId(name))
+    expect(pushMock).toHaveBeenCalled()
+    expect(wave.args[name]).toBe(name)
+  })
+
+  it('Makes card unclickable if name was not provided', () => {
```
mturoci, line 85: Name is required, so it always has to be provided.
```suggestion
  it('Makes card unclickable if name is empty', () => {
```

ui/src/wide_info.test.tsx
```diff
@@ -73,4 +73,31 @@ describe('WideInfo.tsx', () => {
...
   })
+
+  it('Makes card clickable only if name was specified but label was not', () => {
+    const { getByTestId } = render(<View {...wideInfoProps} />)
+    fireEvent.click(getByTestId(name))
+    expect(pushMock).toHaveBeenCalled()
+    expect(wave.args[name]).toBe(name)
+  })
+
+  it('Makes card unclickable if name was not provided', () => {
```
mturoci, line 84:
```suggestion
  it('Makes card unclickable if name is empty', () => {
```

mturoci (dismissed):
Thanks @aalencar, looks good.

Please also describe this new feature in the docstring so that it is discoverable. Don't forget to run `make generate` then.

aalencar:
>Please also describe this new feature in the docstring so that it is discoverable

Are the following comments the docstrings you are talking about? If so, I believe it's already documenting the feature. See `name` and `label`:

```typescript
/** Create a tall information card displaying a title, caption and either an icon or image. */
export interface State {
  /** An identifying name for this card. Makes the card clickable, similar to a button. */
  name: S
  /** The card's title. */
  title: S
  /** The card's caption, displayed below the title. Supports markdown. */
  caption: S
  /** Label of a button rendered at the bottom of the card. If specified, whole card is not clickable anymore. */
  label?: S
  /** The card's icon. */
  icon?: S
  /** The card’s image. */
  image?: S
  /** The card’s image height in px. Defaults to '150px'. */
  image_height?: S
  /** The card's category, displayed below the title. */
  category?: S
}
```

mturoci:
I don't see any mention of the card not being clickable when an empty `name` is provided which is the main point of this issue.

aalencar:
I changed to  `/** An identifying name for this card. Makes the card clickable only if name is not empty and label is empty */`

mturoci:
>Don't forget to run make generate then

aalencar:
> Don't forget to run make generate then

My bad 😅

mturoci (approved):
Thanks @aalencar!

PR #1141, opened 2021-12-01
feat: add value prop to facepile #1133

### LINKED ISSUE #1133: Add value attribute to Facepile component ( Improvement )
When Facepile component is rendered as a 'Add button', it does not have the ability to provide a value.

aalencar:
* If value is provided for facepile, submit that value when add button is clicked. If not, submit true

resolves #1133

### DIFF
--- a/py/examples/facepile.py
+++ b/py/examples/facepile.py
@@ -16,7 +16,7 @@ async def serve(q: Q):
     else:
         image = 'https://images.pexels.com/photos/220453/pexels-photo-220453.jpeg?auto=compress&h=750&w=1260'
         q.page['example'] = ui.form_card(box='1 1 2 2', items=[
-            ui.facepile(name='facepile', max=4, items=[
+            ui.facepile(name='facepile', value='submitted value', max=4, items=[
                 ui.persona(title='John Doe', image=image),
                 ui.persona(title='John Doe', image=image),
                 ui.persona(title='John Doe'),
--- a/py/examples/facepile.py
+++ b/py/examples/facepile.py
@@ -13,10 +13,24 @@ async def serve(q: Q):
             ui.text_m(f'q.args.facepile={q.args.facepile}'),
             ui.button(name='back', label='Back', primary=True),
         ]
+    elif q.args.facepile_value:
+        q.page['example'].items = [
+            ui.text_m(f'q.args.facepile={q.args.facepile_value}'),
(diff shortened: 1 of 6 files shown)


py/examples/facepile.py
```diff
@@ -16,7 +16,7 @@ async def serve(q: Q):
     else:
         image = 'https://images.pexels.com/photos/220453/pexels-photo-220453.jpeg?auto=compress&h=750&w=1260'
         q.page['example'] = ui.form_card(box='1 1 2 2', items=[
-            ui.facepile(name='facepile', max=4, items=[
+            ui.facepile(name='facepile', value='submitted value', max=4, items=[
```
mturoci, line 19: Please add another facepile component to the example instead of munging all together so that one submits original `True` and other one submits the specified `value`. Similar to https://github.com/h2oai/wave/blob/master/py/examples/links.py.

ui/src/facepile.tsx
```diff
@@ -16,18 +16,20 @@ export interface Facepile {
...
-    { name, max, items } = model,
+    { name, max, items, value } = model,
     onClick = () => {
       if (name) {
         if (name.startsWith('#')) {
           window.location.hash = name.substr(1)
           return
         }
-        wave.args[name] = true
+        wave.args[name] = value || true
```
mturoci, line 32: This will fail if `value` is specified as an empty string. Better go with
```suggestion
        wave.args[name] = value === undefined || value
```

mturoci (changes requested):
Thanks @aalencar! Looking good, just a few small nits.

ui/src/facepile.tsx
aalencar: I don't see the point of letting the user send an empty string but If that's the case I will change.

py/examples/facepile.py
```diff
@@ -13,16 +13,30 @@ async def serve(q: Q):
             ui.text_m(f'q.args.facepile={q.args.facepile}'),
             ui.button(name='back', label='Back', primary=True),
         ]
+    elif q.args.facepile_value:
+        q.page['example'].items = [
+            ui.text_m(f'q.args.facepile={q.args.facepile_value}'),
```
mturoci, line 18:
```suggestion
            ui.text_m(f'q.args.facepile_value={q.args.facepile_value}'),
```

mturoci (approved):
Thanks @aalencar!

PR #1142, opened 2021-12-02
feat: Add dialog prop to dropdown #1116

### LINKED ISSUE #1116: Add popup attribute to dropdown to force popup dialog
Add `popup: bool` to `ui.dropdown()`, default to `false`. If set to `true`, then always display the popup picker regardless of the number of items.

Via https://github.com/h2oai/wave/discussions/1100#discussioncomment-1588518

cc @dipplestix

aalencar:
* Define when dialog should be displayed. If not provided, display only when choices > 100

API: `dialog?: 'auto' <- (default) | 'always' | 'never'`

resolves #1116

### DIFF
--- a/py/examples/dropdown.py
+++ b/py/examples/dropdown.py
@@ -33,7 +33,7 @@ async def serve(q: Q):
                         disabled=True),
             ui.dropdown(name='dropdown_dialog', label='Pick multiple in dialog (>100 choices)', values=['1'],
                         required=True, choices=choices_dialog),
-            ui.dropdown(name='dropdown_dialog_always', label='Always show dialog even when choices < 100', value='A',
+            ui.dropdown(name='dropdown_popup_always', label='Always show popup even when choices < 100', value='A',
                         required=True, choices=choices, popup='always'),
             ui.dropdown(name='dropdown_dialog_never', label='Never show dialog even when choices > 100', value='1',
                         required=True, choices=choices_dialog, popup='never'),
--- a/py/examples/dropdown.py
+++ b/py/examples/dropdown.py
@@ -33,6 +33,10 @@ async def serve(q: Q):
                         disabled=True),
             ui.dropdown(name='dropdown_dialog', label='Pick multiple in dialog (>100 choices)', values=['1'],
                         required=True, choices=choices_dialog),
+            ui.dropdown(name='dropdown_dialog_always', label='Always show dialog even when choices < 100', value='A',
+                        required=True, choices=choices, popup='always'),
+            ui.dropdown(name='dropdown_dialog_never', label='Never show dialog even when choices > 100', value='1',
+                        required=True, choices=choices_dialog, popup='never'),
             ui.button(name='show_inputs', label='Submit', primary=True),
         ])
     await q.page.save()
--- a/py/examples/dropdown.py
+++ b/py/examples/dropdown.py
@@ -22,6 +22,8 @@ async def serve(q: Q):
             ui.text(f'dropdown_multi={q.args.dropdown_multi}'),
             ui.text(f'dropdown_disabled={q.args.dropdown_disabled}'),
             ui.text(f'dropdown_dialog={q.args.dropdown_dialog}'),
+            ui.text(f'dropdown_popup_always={q.args.dropdown_popup_always}'),
+            ui.text(f'dropdown_popup_never={q.args.dropdown_popup_never}'),
             ui.button(name='show_form', label='Back', primary=True),
         ]
     else:
--- a/py/examples/dropdown.py
+++ b/py/examples/dropdown.py
@@ -35,7 +35,7 @@ async def serve(q: Q):
                         required=True, choices=choices_dialog),
             ui.dropdown(name='dropdown_popup_always', label='Always show popup even when choices < 100', value='A',
                         required=True, choices=choices, popup='always'),
-            ui.dropdown(name='dropdown_dialog_never', label='Never show dialog even when choices > 100', value='1',
+            ui.dropdown(name='dropdown_popup_never', label='Never show popup even when choices > 100', value='1',
                         required=True, choices=choices_dialog, popup='never'),
             ui.button(name='show_inputs', label='Submit', primary=True),
         ])
--- a/ui/src/dropdown.test.tsx
+++ b/ui/src/dropdown.test.tsx
@@ -377,38 +377,38 @@ describe('Dropdown.tsx', () => {
     })
 
     it(`Displays dialog when choices > 100 and 'popup' prop is not provided`, () => {
-      const { getByTestId, getAllByRole  } = render(<XDropdown model={dialogProps} />)
+      const { getByTestId, getByRole  } = render(<XDropdown model={dialogProps} />)
 
       fireEvent.click(getByTestId(name))
 
-      expect(getAllByRole('dialog')).toBeDefined()
+      expect(getByRole('dialog')).toBeInTheDocument()
     })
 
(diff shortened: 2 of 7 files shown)


ui/src/dropdown.tsx
```diff
@@ -59,6 +59,8 @@ export interface Dropdown {
   visible?: B
   /** An optional tooltip message displayed when a user clicks the help icon to the right of the component. */
   tooltip?: S
+  /** Define when the dialog will be displayed. If no value is provided, defaults to 'auto' which will display the dialog only when there are more than 100 choices. */
+  dialog?: 'auto' | 'always' | 'never'
```
lo5, line 63:
```suggestion
  popup?: 'auto' | 'always' | 'never'
```

ui/src/dropdown.tsx
```diff
@@ -59,6 +59,8 @@ export interface Dropdown {
   visible?: B
   /** An optional tooltip message displayed when a user clicks the help icon to the right of the component. */
   tooltip?: S
+  /** Define when the dialog will be displayed. If no value is provided, defaults to 'auto' which will display the dialog only when there are more than 100 choices. */
```
lo5, line 62:
```suggestion
  /** Whether to present the choices using a pop-up dialog. Defaults to `auto`, which pops up a dialog only when there are more than 100 choices. */
```

ui/src/dropdown.tsx
```diff
@@ -246,11 +248,22 @@ const
...
     )
+  },
+  getCompBasedOnDialogProp = (m: Dropdown) => {
+    m.dialog = m.dialog ?? 'auto'
+    switch (m.dialog) {
+      case 'always':
+        return <DialogDropdown {...m} />
+      case 'never':
+        return <BaseDropdown {...m} />
+      case 'auto':
```
lo5, line 259:
```suggestion
      default: // auto
```

ui/src/dropdown.tsx
```diff
@@ -246,11 +248,22 @@ const
         </Fluent.Dialog>
       </>
     )
+  },
+  getCompBasedOnDialogProp = (m: Dropdown) => {
+    m.dialog = m.dialog ?? 'auto'
```
lo5, line 253:
```suggestion

```
No point in mutating this field unnecessarily.

ui/src/dropdown.tsx
```diff
@@ -246,11 +248,22 @@ const
...
+    }
   }
 
 export const XDropdown = ({ model: m }: { model: Dropdown }) => {
   // eslint-disable-next-line react-hooks/exhaustive-deps
   React.useEffect(() => { wave.args[m.name] = m.values ? (m.values || []) : (m.value || null) }, [])
-
-  return (m.choices?.length || 0) > 100 ? <DialogDropdown {...m} /> : <BaseDropdown {...m} />
+  
+  return getCompBasedOnDialogProp(m)
```
lo5, line 268: I would just inline this like before instead of a separate function.

```
return m.dialog === 'always' 
    ? <DialogDropdown {...m} /> 
    : m.dialog === 'never' 
        ? <BaseDropdown {...m} /> 
        : (m.choices?.length || 0) > 100 
            ? <DialogDropdown {...m} /> 
            : <BaseDropdown {...m} />
```

lo5 (changes requested):
Thanks @aalencar - minor changes to API - rename `dialog` (noun) to `popup` (verb).

ui/src/dropdown.tsx
aalencar: I like it. Inlined.

ui/src/dropdown.tsx
aalencar: Better wording. Done.

ui/src/dropdown.tsx
aalencar: I thought of popup as in "the popup window", that's why I used dialog instead. Thinking about it as a verb makes more sense.

py/examples/dropdown.py
```diff
@@ -33,6 +33,10 @@ async def serve(q: Q):
                         disabled=True),
             ui.dropdown(name='dropdown_dialog', label='Pick multiple in dialog (>100 choices)', values=['1'],
                         required=True, choices=choices_dialog),
+            ui.dropdown(name='dropdown_dialog_always', label='Always show dialog even when choices < 100', value='A',
```
lo5, line 36:
```suggestion
            ui.dropdown(name='dropdown_popup_always', label='Always show popup even when choices < 100', value='A',
```

py/examples/dropdown.py
```diff
@@ -33,6 +33,10 @@ async def serve(q: Q):
...
+                        required=True, choices=choices, popup='always'),
+            ui.dropdown(name='dropdown_dialog_never', label='Never show dialog even when choices > 100', value='1',
```
lo5, line 38:
```suggestion
            ui.dropdown(name='dropdown_popup_never', label='Never show popup even when choices > 100', value='1',
```

lo5 (changes requested):
Thanks @aalencar - minor changes to the examples to make them consistent with the `popup` attribute. If possible, can you also change the label on the existing `popup=auto` example?

ui/src/dropdown.test.tsx
```diff
@@ -375,5 +375,40 @@ describe('Dropdown.tsx', () => {
       fireEvent.click(getByTestId(name))
       expect(getAllByRole('listitem')).toHaveLength(40) // Fluent Detaillist uses virtualization, so only first 40 listitems are rendered.
     })
+
+    it(`Displays dialog when choices > 100 and 'popup' prop is not provided`, () => {
+      const { getByTestId, getAllByRole  } = render(<XDropdown model={dialogProps} />)
+
+      fireEvent.click(getByTestId(name))
+
+      expect(getAllByRole('dialog')).toBeDefined()
```
mturoci, line 384: No need for `getAll` since there should always be just a single dialog. Also `toBeDefined` is not that great of an assertion, more readable would be `toBeVisible` or `toBeInTheDocument`. Apply the same to the rest of test cases as well.

```suggestion
const { getByTestId, queryByRole } = render(<XDropdown model={dialogProps} />)

      expect(queryByRole('dialog')).not.toBeInTheDocument()
      fireEvent.click(getByTestId(name))
      expect(queryByRole('dialog')).toBeInTheDocument()
```

aalencar: Done

ui/src/dropdown.test.tsx
```diff
@@ -375,5 +375,40 @@ describe('Dropdown.tsx', () => {
...
+      dialogProps.popup = 'always'
+      dialogProps.choices = [{ name: 'A' }]
+      const { getByTestId, getByRole  } = render(<XDropdown model={dialogProps} />)
+
+      fireEvent.click(getByTestId(name))
+
+      expect(getByRole('dialog')).toBeInTheDocument()
+    })
+
+    it(`Displays dialog when choices > 100 and 'popup' prop is set as 'never'`, () => {
```
mturoci, line 405: If `never` is specified, the dialog shouldn't be displayed.

aalencar: Fixed.

mturoci:
https://github.com/h2oai/wave/pull/1152#issuecomment-989903414

aalencar:
> [#1152 (comment)](https://github.com/h2oai/wave/pull/1152#issuecomment-989903414)

done

mturoci (approved):
Great, thanks @aalencar!

Issue #699, opened 2021-03-27
Improve Showcase


lo5:
Make the showcase app in `py/demo` similar in principle to the [Bootstrap](https://getbootstrap.com/docs/5.0/components/alerts/) docs - an app that serves three purposes:

1. An extensive showcase of what Wave is capable of.
2. A reference developers can use in addition to the the currrent "Tour" app ("How do I do that?").
3. (For contributors) - a development-time tool manually/visually test new cards/components or changes to existing ones. If we build up a comprehensive collection of every card/component variation, the next step would be to automate visual verification (based on #531).

**Tasks**
- [ ] Create one page per card (with all possible variations)
- [ ] Create one page per component (with all possible variations)
- [ ] Guide level docs
- [ ] :muscle: Compile and display code snippet into the page for copypasta convenience

PR #1147, opened 2021-12-06
chore: remove static examples from Tour #31

### LINKED ISSUE #31: Hard-code synthetic data in examples

mturoci:
Since all the static examples moved to our online docs (in #699), there is no need for duplication in Tour.

The proposal (originally by @lo5 quite a while ago) was to use our docs for showcase/static examples and the Tour app for interactive ones.

However, removing `synth` may not be the best idea since I find it useful for cases when one wants to generate real-looking mock data, for example in a table or dashboard examples without bloating the code, so I didn't remove it completely.

@lo5 wdyt?

Closes #31


mtanco:
Feedback that a lot of users have been very confused by Synth. Can we use synth to generate a dataset and then that data comes with Tour?

mturoci:
@mtanco would https://github.com/h2oai/wave/blob/master/py/demo/synthetic_data.py be a viable replacement for Synth?

lo5:
There's an existing PR https://github.com/h2oai/wave/pull/443/files meant to remove `synth.py` (probably super-outdated by now to be of any use). 

For the examples to be more readable, it's better to embed the data directly in each example, like in the linked PR, especially  with columns/rows that represent some kind of real world data (instead of the random values generated by synth.py).

mturoci:
Okay, what about the duplicates (basically all non-interactive examples)? Shall they be removed (like in this PR) or kept with hardcoded data?

mturoci:
Based on yesterday's discussion, having duplicated examples across docs/tour is fine, the only thing needed to be done is to replace synth with reasonable data.

PR #1149, opened 2021-12-07
feat: Add trigger to combobox #1136

### LINKED ISSUE #1136: Add trigger to combobox
via @azim-b

aalencar:
* Add prop trigger to combobox allowing the choice to be sent to the server right away when it is selected.

resolves #1136

### DIFF
--- a/py/examples/combobox.py
+++ b/py/examples/combobox.py
@@ -25,8 +25,6 @@ async def serve(q: Q):
                         choices=combobox_choices, disabled=True),
             ui.combobox(name='combobox_error', label='Enter or choose a color', placeholder='Color...', value='Blue',
                         choices=combobox_choices, error='This combobox has an error!'),
-            ui.combobox(name='combobox_trigger', trigger=True, label='Enter or choose a color (submit color when changed)', placeholder='Color...', value='Blue',
-                        choices=combobox_choices),
             ui.button(name='show_inputs', label='Submit', primary=True),
         ])
     await q.page.save()
--- a/py/examples/combobox.py
+++ b/py/examples/combobox.py
@@ -14,7 +14,6 @@ async def serve(q: Q):
             ui.text(f'combobox={q.args.combobox}'),
             ui.text(f'combobox_disabled={q.args.combobox_disabled}'),
             ui.text(f'combobox_error={q.args.combobox_error}'),
-            ui.text(f'combobox_trigger={q.args.combobox_trigger}'),
             ui.button(name='show_form', label='Back', primary=True),
         ]
     else:
--- a/py/examples/combobox.py
+++ b/py/examples/combobox.py
@@ -25,7 +25,7 @@ async def serve(q: Q):
                         choices=combobox_choices, disabled=True),
             ui.combobox(name='combobox_error', label='Enter or choose a color', placeholder='Color...', value='Blue',
                         choices=combobox_choices, error='This combobox has an error!'),
-            ui.combobox(name='combobox_trigger', trigger=True, label='Enter or choose a color (trigger on selection)', placeholder='Color...', value=q.args.combobox_trigger,
+            ui.combobox(name='combobox_trigger', trigger=True, label='Enter or choose a color (submit color when changed)', placeholder='Color...', value='Blue',
                         choices=combobox_choices),
             ui.button(name='show_inputs', label='Submit', primary=True),
         ])
--- a/py/examples/combobox.py
+++ b/py/examples/combobox.py
@@ -14,6 +14,7 @@ async def serve(q: Q):
             ui.text(f'combobox={q.args.combobox}'),
             ui.text(f'combobox_disabled={q.args.combobox_disabled}'),
             ui.text(f'combobox_error={q.args.combobox_error}'),
+            ui.text(f'combobox_trigger={q.args.combobox_trigger}'),
             ui.button(name='show_form', label='Back', primary=True),
         ]
     else:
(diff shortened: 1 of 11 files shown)


py/examples/combobox.py
```diff
@@ -24,6 +25,8 @@ async def serve(q: Q):
                         choices=combobox_choices, disabled=True),
             ui.combobox(name='combobox_error', label='Enter or choose a color', placeholder='Color...', value='Blue',
                         choices=combobox_choices, error='This combobox has an error!'),
+            ui.combobox(name='combobox_trigger', trigger=True, label='Enter or choose a color (trigger on selection)', placeholder='Color...', value=q.args.combobox_trigger,
```
aalencar, line 28: @mturoci is there something wrong with using the value that was recently send to the server? And how do I conditionally pass arguments in Python like in JS `value || fallback`?

py/examples/picker.py
```diff
@@ -14,7 +14,7 @@ async def serve(q: Q):
         ]
     else:
         q.page['example'] = ui.form_card(box='1 1 4 10', items=[
-            ui.picker(name='picker', label='Place an order (try Spam, Eggs or Ham):', choices=[
+            ui.picker(name='picker', trigger=True, label='Place an order (try Spam, Eggs or Ham):', choices=[
```
lo5, line 17: This example is unrelated to the combobox. Is this an intentional change?

py/examples/combobox.py
lo5: It's confusing to read. It's enough if the example line demonstrates how to make a combobox submit its value when changed. The `value=q.args.combobox_trigger` is unnecessary.

lo5, line 28:
```suggestion
            ui.combobox(name='combobox_trigger', trigger=True, label='Enter or choose a color (submit color when changed)', placeholder='Color...', value='Blue',
```

lo5 (changes requested):

py/examples/combobox.py
mturoci: >is there something wrong with using the value that was recently send to the server

`value` is meant to be "initial" value - all wave components should be uncontrolled. That means the `value` would work only if you recreate the card.

>how do I conditionally pass arguments in Python 

`value or fallback`

mturoci: Anyway it's better to create a separate example for `trigger`, like we have for `textbox` for example. The reason is that triggering will submit also filled values from other comboboxes.

py/examples/picker.py
aalencar: I was checking the picker behavior on trigger and commited it by mistake. Reverted, thanks.

py/examples/combobox.py
aalencar: I added a new example like we have for textbox

py/examples/combobox.py
```diff
@@ -14,6 +14,7 @@ async def serve(q: Q):
             ui.text(f'combobox={q.args.combobox}'),
             ui.text(f'combobox_disabled={q.args.combobox_disabled}'),
             ui.text(f'combobox_error={q.args.combobox_error}'),
+            ui.text(f'combobox_trigger={q.args.combobox_trigger}'),
```
mturoci, line 17: out

ui/src/combobox.tsx
```diff
@@ -51,6 +51,8 @@ export interface Combobox {
   visible?: B
   /** An optional tooltip message displayed when a user clicks the help icon to the right of the component. */
   tooltip?: S
+  /** True if the choice should be submitted when the dropdown value changes. */
```
mturoci, line 54: Not just dropdown.

mturoci (changes requested):
Thanks @aalencar. Don't forget that combobox also allows free text (not just dropdown) which needs to be triggered as well. The free text needs to be debounced so that we do not submit every keystroke. See `textbox` for an implementation reference.

py/examples/combobox.py
aalencar: Removed.

ui/src/combobox.tsx
aalencar: Improved.

aalencar:
> The free text needs to be debounced so that we do not submit every keystroke

I implemented the debounce but I noticed that `onChange` is only triggered on blur, pressing enter,  or selecting a choice.
Based on that, there is no need to debounce if we prevent the user from sending undefined choices. I added a guard to prevent that.

mturoci:
>I implemented the debounce but I noticed that onChange is only triggered on blur, pressing enter, or selecting a choice.

Good to know.

>Based on that, there is no need to debounce if we prevent the user from sending undefined choices. I added a guard to prevent that.

I don't understand this one. The value will never be undefined thanks to `const v = option?.text || value || ''`, right?

aalencar:
> > I implemented the debounce but I noticed that onChange is only triggered on blur, pressing enter, or selecting a choice.
> 
> Good to know.
> 
> > Based on that, there is no need to debounce if we prevent the user from sending undefined choices. I added a guard to prevent that.
> 
> I don't understand this one. The value will never be undefined thanks to `const v = option?.text || value || ''`, right?

By undefined I mean some choice that is not present in choices / dropdown (i.e. not defined by the user).

Example: `choices = ['red', 'yellow', 'green']`, if user types `blue` or `yello` it won't send it to the server

 I made it more confusing by using the word 'undefined' in a javascript context. I hope it's more clear now.

mturoci:
But this is not the correct logic, combobox allows free text as well which should be sent to server also. Otherwise, there would be no difference between combobox and dropdown basically. Please revert the condition.

aalencar:
> But this is not the correct logic, combobox allows free text as well which should be sent to server also. Otherwise, there would be no difference between combobox and dropdown basically. Please revert the condition.

Reverted.

mturoci (approved):
Thanks @aalencar!

Issue #1145, opened 2021-12-03
Allow inline for ui.choice_group()


mturoci:
### Discussed in https://github.com/h2oai/wave/discussions/1144

<div type='discussions-op-text'>

<sup>Originally posted by **azim-b** December  3, 2021</sup>
Is there an option to make the choices of a `ui.choice_group()` inline/horizontal?</div>

## Implementation notes

Same as https://github.com/h2oai/wave/blob/b0f066a506d5ecf4fd152b522bda4d0eda5123b9/ui/src/checklist.tsx#L40

mturoci:
Closed by #1152

PR #1152, opened 2021-12-09
feat: Add inline option for choice group #1145

aalencar:
image

### DIFF
--- a/py/examples/choice_group_inline.py
+++ b/py/examples/choice_group_inline.py
@@ -1,0 +1,32 @@
+# Form / Choice Group / Inline
+# Use #choice groups to let users select one option from two or more choices.
+# #form #choice_group #inline
+# ---
+from h2o_wave import main, app, Q, ui
+
+choices = [
+    ui.choice('A', 'Option A'),
+    ui.choice('B', 'Option B'),
+    ui.choice('C', 'Option C', disabled=True),
+    ui.choice('D', 'Option D'),
+    ui.choice('E', 'Option E'),
+    ui.choice('F', 'Option F'),
+    ui.choice('G', 'Option H'),
+    ui.choice('I', 'Option I'),
+]
+
+
+@app('/demo')
+async def serve(q: Q):
+    if q.args.show_inputs:
+        q.page['example'].items = [
+            ui.text(f'selected={q.args.choice_group}'),
+            ui.button(name='show_form', label='Back', primary=True),
+        ]
+    else:
+        q.page['example'] = ui.form_card(box='1 1 4 10', items=[
+            ui.choice_group(name='choice_group', inline=True, label='Pick one',
+                            value='B', required=True, choices=choices),
+            ui.button(name='show_inputs', label='Submit', primary=True),
+        ])
+    await q.page.save()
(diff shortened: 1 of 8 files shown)


py/examples/choice_group_inline.py
```diff
@@ -0,0 +1,32 @@
+# Form / Choice Group / Inline
+# Use #choice groups to let users select one option from two or more choices.
```
mturoci, line 2: Don't forget to update docstrings when copy-pasting.

ui/src/choice_group.tsx
```diff
@@ -62,6 +65,21 @@ export interface ChoiceGroup {
...
 }
 
+const
+  css = stylesheet({
+    inline: {
+      $nest: {
+        '.ms-ChoiceFieldGroup-flexContainer': {
+          display: 'flex',
+          flexWrap: 'wrap',
+        },
```
mturoci, line 75: Prefer to use Fluent way of restyling (using `styles` prop on component) when possible. If Fluent team decides to change the CSS class name, we will not know, but if they change the TS API our build will fail so that we can fix it ourselves.

mturoci (changes requested):
Thanks @aalencar! Just a few minor changes.

mturoci:
@aalencar also please include the newly added features/cards in the docs as well. This one to the https://github.com/h2oai/wave/blob/master/website/components/form/choice_group.md.

ui/src/choice_group.tsx
aalencar: Thanks for the tip. I couldn't add the margin for the subcomponent though as the parent doesn't expose its style props, so I used the `selectors` approach. Let me know if there is a better way of handling this.

I followed this guide https://github.com/microsoft/fluentui/wiki/Component-Styling#-styles--prop

aalencar:
> @aalencar also please include the newly added features/cards in the docs as well. This one to the https://github.com/h2oai/wave/blob/master/website/components/form/choice_group.md.

I didn't know about these docs. I will update them from now on.

ui/src/choice_group.tsx
```diff
@@ -69,13 +71,25 @@ export const
...
+          }
+        },
       }
 
     // eslint-disable-next-line react-hooks/exhaustive-deps
     React.useEffect(() => { wave.args[m.name] = m.value || null }, [])
 
     return (
       <Fluent.ChoiceGroup
+        styles={styles}
```
mturoci, line 92:
```suggestion
        styles={{ flexContainer: { display: 'flex', flexWrap: 'wrap' } }}
```

ui/src/choice_group.tsx
```diff
@@ -69,13 +71,25 @@ export const
...
         if (m.trigger) wave.push()
+      },
+      styles: Fluent.IChoiceGroupStyles = {
+        flexContainer: {
+          display: 'flex',
+          flexWrap: 'wrap',
+          selectors: {
+            '.ms-ChoiceField': {
+              marginRight: 15,
+            }
```
mturoci, line 82: For this one you need to update styles of each particular option (L70) to ` styles: { choiceFieldWrapper: { marginRight: 15 } }`

mturoci (changes requested):

ui/src/choice_group.tsx
aalencar: done

mturoci (approved):
Thanks @aalencar!

PR #1153, opened 2021-12-13
feat: Package Wave daemon within python wheel. #250

### LINKED ISSUE #250: Package server binaries with PyPI package

mturoci:
For testing purposes one can use `pip install -i https://test.pypi.org/simple/ h2o-wave-test` which should download the correct OS-specific distribution. This PR also starts a wave daemon together with `wave run app`.

* Verified that pip downloads correct distribution also for higher mac versions than `10_9` - thanks @geomodular!
* Verified that `wave run app` works on Windows also - thanks @Far0n !
* Had to also bundle UI build, not just a standalone server binary.

## Possible improvements

* @geomodular suggested adding an env var to skip spinning up wave daemon in case anyone wanted to go by the "old way" - separate daemon and separate app
* python virtual env dir structure is different on Windows (`venv/lib/site-packages/h2o_wave`) vs unix-based systems (`venv/lib/python3.7/site-packages/h2o_wave`) - note the missing `python3.7` dir. This is currently solved with an ugly workaround. However, a more viable option could to recursively find the `waved` binary starting from the current file (`cli.py`) with a max depth specified to avoid scanning full FS. This ofc comes for the price of worse perf.

@lo5 please have a look at the PR and let me know wyt. Once the review is passed, I will fill in the docs. 

Closes #250


lo5 (approved):
Awesome @mturoci! This is a significant devX improvement. Thanks for your help, @geomodular, @Far0n!

py/h2o_wave/cli.py
```diff
@@ -77,9 +79,21 @@ def run(app: str, no_reload: bool):
     if ext.lower() == '.py':
         app = app_path.replace(os.path.sep, '.')
 
-    uvicorn.run(f'{app}:main', host=_localhost, port=port, reload=not no_reload)
+    # Try to start Wave daemon.
+    try:
```
geomodular, line 83: Just my observation.

`try`/`finally` block feels unnecessary, but maybe it's on purpose:
1. The only function you want to `try` seems to be `Popen()`. You can move the `try` keyword right before. 🤔 
2. `finally` is executed regardless of exception. This might be what you want to run the `wave` app even if the try block fails.

geomodular (approved):
Good job @mturoci!

py/h2o_wave/cli.py
mturoci: Agree, good point, thanks!

mturoci:
>  an env var to skip spinning up wave daemon in case anyone wanted to go by the "old way" - separate daemon and separate app

Leaving that out for now, can be implemented once requested.

Issue #1132, opened 2021-11-19
WideArticlePreview card


mturoci:
![image](https://user-images.githubusercontent.com/64769322/142599492-9d09ee03-9453-400b-85e4-36b94aaf19fa.png)

design url - https://xd.adobe.com/view/acfb7ef1-5b99-4129-97b2-b30f8086c239-f9a8/

mturoci:
Closed by #1155

PR #1155, opened 2021-12-15
feat: add Wide Article Preview component #1132

aalencar:
### Design
image

### Themes
Default
image

H2O-dark
image

Neon
image

Proposed API: 

```ts
/** Create a wide article preview card displaying a persona, image, title, caption, and optional buttons. */
export interface State {
  /** The card's user avatar, 'size' prop is restricted to 'xs'. */
  persona: Component
  /** The card’s image displayed on the left-hand side. */
  image: S
  /** The card's title on the righ-hand side */
  title: S
  /** An identifying name for this card. Makes the card clickable, similar to a button. */
  name?: S
  /** The card's auxiliary text, displayed on the right-hand side of the header. */
  aux_value?: S
  /** The card's caption, displayed bellow the title on the right-hand side. */
  caption?: S
  /** The card's buttons, displayed at the bottom-right corner. */
  items?: Component[]
}
```

### DIFF
--- a/py/examples/wide_article_preview.py
+++ b/py/examples/wide_article_preview.py
@@ -6,36 +6,45 @@ from h2o_wave import main, app, Q, ui
 
 @app('/demo')
 async def serve(q: Q):
-    persona_pic = 'https://images.pexels.com/photos/415829/pexels-photo-415829.jpeg?auto=compress&cs=tinysrgb&dpr=2&h=100&w=100'
-    q.page['example'] = ui.wide_article_preview_card(
-        box='1 1 6 5',
-        persona=ui.persona(title='Jasmine Grand', subtitle='Marketing Executive',
-                           image=persona_pic, caption='caption'),
-        commands=[
-            ui.command(name='new', label='New', icon='Add', items=[
-                ui.command(name='email', label='Email Message', icon='Mail'),
-                ui.command(name='calendar', label='Calendar Event', icon='Calendar'),
-            ]),
-            ui.command(name='upload', label='Upload', icon='Upload'),
-            ui.command(name='share', label='Share', icon='Share'),
-            ui.command(name='download', label='Download', icon='Download'),
-        ],
-        aux_value='2h ago',
-        image='https://images.pexels.com/photos/1269968/pexels-photo-1269968.jpeg?auto=compress',
-        title='Jasmine Grand',
-        caption='''
-            Lorem ipsum dolor sit amet consectetur adipisicing elit. Ipsum sed odio porro veniam dolorum velit doloremque neque aliquam quisquam rem officiis, eius facilis iste quam, minus repellat magni iure eaque!
-            Veniam pariatur itaque nisi, a nam consequuntur. Aliquam nulla sequi, nihil soluta quaerat vitae inventore magni vero voluptates officiis dolorem alias incidunt iure in sapiente doloribus, quos distinctio? Illo, ullam?
-        ''',
-        items=[
-            ui.inline(justify='end', items=[
-                ui.mini_buttons([
-                    ui.mini_button(name='like', label='4', icon='Heart'),
-                    ui.mini_button(name='comment', label='2', icon='Comment'),
-                    ui.mini_button(name='share', label='1', icon='Share'),
+    if q.args.wide_article_preview:
+        q.page['example'] = ui.form_card(box='1 1 4 6', items=[
+            ui.button(name='back', label='Go back', primary=True),
+        ])
+    else:
+        q.page['meta'] = ui.meta_card(box='', theme='neon')
+        persona_pic = 'https://images.pexels.com/photos/415829/pexels-photo-415829.jpeg?auto=compress&cs=tinysrgb&dpr=2&h=100&w=100'
+        q.page['example'] = ui.wide_article_preview_card(
+            box='1 1 6 5',
+            name='wide_article_preview',
+            persona=ui.persona(title='Jasmine Grand', subtitle='Marketing Executive',
+                               image=persona_pic, caption='caption'),
+            commands=[
+                ui.command(name='new', label='New', icon='Add', items=[
+                    ui.command(name='email', label='Email Message', icon='Mail'),
+                    ui.command(name='calendar', label='Calendar Event', icon='Calendar'),
                 ]),
-            ])
-        ],
-    )
+                ui.command(name='upload', label='Upload', icon='Upload'),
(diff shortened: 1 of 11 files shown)


ui/src/wide_article_preview.tsx
```diff
@@ -0,0 +1,144 @@
...
+        height: '40px !important',
+        width: '40px !important',
+      },
+      '.ms-Persona-imageArea': {
+        height: '40px !important',
+        width: '40px !important',
+      }
+    }
+  },
+  content: {
```
mturoci, line 59: Probably needs some bottom spacing for cases when content is so long it reaches the items.

ui/src/wide_article_preview.tsx
```diff
@@ -0,0 +1,144 @@
...
+        lineHeight: px(22),
+        fontWeight: 500,
+      },
+      '.ms-Persona-secondaryText': {
+        fontSize: 14,
+        lineHeight: px(20),
+        color: cssVar('$text7')
+      },
+      '.ms-Persona-image': {
+        height: '40px !important',
```
mturoci, line 50: nit: You can use a util func
```suggestion
        height: important('40px'),
```

mturoci (changes requested):
Thanks @aalencar. I see improvements in the code quality compared to the past, good job!

Apart from a few minor comments:
* There also seems to be a missing spacing between title and caption compared to the original design. 
* Also please align the title to the top image edge.

image

* Please align `aux_value` with persona subtitle
image

ui/src/wide_article_preview.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+import React from 'react'
+import { wave } from './ui'
+import { View } from './wide_article_preview'
+
+const
+  name = 'wide_article_preview',
+  pushMock = jest.fn()
+let wideArticlePreviewProps: Model<any>
+
+describe('WideArticlePreview.tsx', () => {
```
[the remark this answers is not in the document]
aalencar: added

ui/src/wide_article_preview.tsx
aalencar: added

py/examples/wide_article_preview.py
```diff
@@ -0,0 +1,41 @@
+# Wide Article Preview
+# Create a Wide Article Preview card displaying a persona, image, title, caption, and buttons.
+# ---
+from h2o_wave import main, app, Q, ui
+
+
+@app('/demo')
+async def serve(q: Q):
+    persona_pic = 'https://images.pexels.com/photos/415829/pexels-photo-415829.jpeg?auto=compress&cs=tinysrgb&dpr=2&h=100&w=100'
```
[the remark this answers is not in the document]
aalencar: changed

aalencar:
@mturoci when I address your comments, do you prefer that I split the fixes / improvements into separate commits or a single one? Whatever makes it easier for you to review.

mturoci:
Thanks for asking @aalencar, but I don't mind. I look at branch comparison against master and squash all the commits during merge anyway so whatever suits you.

py/examples/wide_article_preview.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+@app('/demo')
+async def serve(q: Q):
+    if q.args.wide_article_preview:
+        q.page['example'] = ui.form_card(box='1 1 4 6', items=[
+            ui.button(name='back', label='Go back', primary=True),
+        ])
+    else:
+        q.page['meta'] = ui.meta_card(box='', theme='neon')
```
mturoci, line 14: out

ui/src/wide_article_preview.tsx
```diff
@@ -0,0 +1,154 @@
...
+        color: cssVar('$text'),
+      }
+    }
+  },
+  title: {
+    marginBottom: 8,
+    lineHeight: 0.6,
+  },
+  caption: {
+    marginBottom: 24,
```
mturoci, line 92: This needs to be better thought through. Think of cases when there are no `items` specified. The code above would result in extra unwanted whitespace. The same goes the other way around - if `caption` is not specified, but `items` are.

website/components/content/article.md
```diff
@@ -70,3 +70,43 @@ q.page['example'] = ui.tall_article_preview_card(
...
+
+## Wide article preview
+
+Horizontal variant.
+
+Check the full API at [ui.wide_article_preview_card](/docs/api/ui#wide_article_preview_card).
+
+```py
+    persona_pic = 'https://images.pexels.com/photos/415829/pexels-photo-415829.jpeg?auto=compress&cs=tinysrgb&dpr=2&h=100&w=100'
+    q.page['example'] = ui.wide_article_preview_card(
```
mturoci, line 82: Look at how other examples are aligned. There is no need to start with a tab. For verification purposes, you can
```sh
cd tools/showcase
make generate
```
^^ This will generate all the screenshots and fails in case of error in python example code. To then view it in action:
```sh
cd website
npm start
```
This will start a dev mode for the docs where you can see if the example is rendered according to the expectations and is well-aligned with the rest.

ui/src/wide_article_preview.tsx
```diff
@@ -0,0 +1,154 @@
...
+        lineHeight: px(22),
+        fontWeight: 500,
+      },
+      '.ms-Persona-secondaryText': {
+        fontSize: 14,
+        lineHeight: px(20),
+        color: cssVar('$text7')
+      },
+      '.ms-Persona-image': {
+        height: important('40px'),
```
mturoci, line 50: Be consistent, either use the util function for the whole for or not at all.

mturoci (changes requested):
Don't forget to run `make generate` when changing docstrings so that py/R API is always in sync.

website/components/content/article.md
aalencar: I had to use Playwright 1.17 because 1.10 is not working on macOS Monterey. I got an error which also happens on `master` branch.

image

ui/src/wide_article_preview.tsx
aalencar: I added the logic to conditionally add space to `caption` based on whether `items` was specified. `title` is a required prop so I didn't have to handle the case where only `items` is provided on the right-hand side.

website/components/content/article.md
mturoci: Thanks for the info, will update the playwright version then.

mturoci (approved):
LGTM!

Issue #1099, opened 2021-10-29
Visually show a color was clicked


mtanco:
#### Is your feature request related to a problem? Please describe
When a user clicks a value in the `color_picker` there is no UI change like we see for `date_picker` or `dropdown` or other components. This makes the user think the click didn't work. However, after clicking a button and updating the value of the color_picker we can see it did work. 

```python

@app('/')
async def serve(q: Q):

    if q.args.save:
        q.page["form"] = ui.form_card(
            box="4 1 4 2",
            items=[
                ui.color_picker(name='color', label='Color', choices=["red", "yellow", "green"], value=q.args.color),
                ui.button(name='save', label='Save Date', primary=True),
            ]
        )

    else:
        q.page["form"] = ui.form_card(
            box="4 1 4 2",
            items=[
                ui.color_picker(name='color', label='Color', choices=["red", "yellow", "green"]),
                ui.button(name='save', label='Save Color', primary=True),
            ]
        )
    await q.page.save()
```

#### Describe the solution you'd like
The browser handles "selecting" the color after it has been clicked. 

#### Describe alternatives you've considered

(A clear and concise description of any alternative solutions or features you've considered.)

#### Additional context
Note that when we use the trigger in this example we end up with 2 `color_picker`s that we would expect to be identical but the original is always going to be the default value

```python
from h2o_wave import main, app, Q, ui


@app('/')
async def serve(q: Q):

    if q.args.color:
        q.page['form_2'] = ui.form_card(
            box="4 4 4 2",
            items=[
                ui.color_picker(name='color', label='Color', choices=["red", "yellow", "green"], value=q.args.color),
            ]
        )

    else:
        q.page["form"] = ui.form_card(
            box="4 1 4 2",
            items=[
                ui.color_picker(name='color', label='Color', choices=["red", "yellow", "green"], trigger=True),
            ]
        )
    await q.page.save()
```

![image](https://user-images.githubusercontent.com/4793561/139476759-1f1553c2-05d1-43c6-a279-1a0d5a84ca7a.png)

PR #1165, opened 2021-12-29
fix: Select color on swatch color picker #1099

aalencar:
* Visually indicate which color is currently selected on color picker


mturoci (commented):
Thanks @aalencar! The solution is correct for Fluent v7, however, let's wait for #1169 and use `defaultSelectedId` without local state instead of `selectedId` with local state.

PR #1180, opened 2022-01-10
fix: Prevent race condition during wave run / showcase by retrying

mturoci:


lo5 (dismissed):
Looks good (haven't tested). Thanks, @mturoci!

mturoci:
Got good feedback from @geomodular about `os.environ.get('H2O_WAVE_NO_AUTOSTART', None) is None` just checking the presence, but not handling the actual true/false correctly. Will fix it tomorrow.

mturoci:
Documented the variable, added support for enabling/disabling.

py/h2o_wave/cli.py
```diff
@@ -80,14 +81,28 @@ def run(app: str, no_reload: bool):
...
     try:
-        server_port = int(os.environ.get('H2O_WAVE_LISTEN', 10101))
-        server_not_running = _scan_free_port(server_port) == server_port
-        if os.environ.get('H2O_WAVE_NO_AUTOSTART', None) is None and server_not_running:
-            subprocess.Popen(['waved.exe' if 'Windows' in platform.system() else './waved'],
-                             cwd=sys.exec_prefix, env=os.environ.copy(), shell=True)
+        waved = 'waved.exe' if 'Windows' in platform.system() else './waved'
+        # OS agnostic wheels do not have waved - needed for HAC.
+        is_waved_present = os.path.isfile(os.path.join(sys.exec_prefix, waved))
+        is_autostart_off = os.environ.get('H2O_WAVE_NO_AUTOSTART', 'false') in ['false', '0', 'f']
```
geomodular, line 90: Not mandatory to do `lower()` but `False` value will not work. Up to you.

geomodular (approved):
Thanks Martin!

Issue #602, opened 2021-03-08
Add Select / Unselect All for Table Filter


mtanco:
#### Is your feature request related to a problem? Please describe
When looking at data in the table it's common to want to see all rows _except_ with a specific value. Today, you have to check mark every option except the one you'd like to filter. In this case I only had 12ish unique values, but it would be arduous with 100s 

![image](https://user-images.githubusercontent.com/4793561/110273923-804df200-7f82-11eb-8c27-5aaf7f6ea4e3.png)

#### Describe the solution you'd like

It would be nice to have a `select all` option to quickly unselect one or two options to filter by

mturoci:
Implementation note: Make the UX and style the same as select / unselect all in `ui.checklist`.

PR #1187, opened 2022-01-14
feat: Add select / deselect all to table filter menu #602

aalencar:
Add Select and Deselect all functionality to filter menu

### DIFF
--- a/ui/src/table.tsx
+++ b/ui/src/table.tsx
@@ -119,7 +119,7 @@ type WaveColumn = Fluent.IColumn & {
 
 type DataTable = {
   model: Table
-  onFilterChange: (filterKey: S, filterVal: S, checked?: boolean) => void
+  onFilterChange: (filterKey: S, filterVal: S, checked?: B) => void
   sort: (col: WaveColumn) => void,
   reset: () => void
   filteredItems: any[]
@@ -132,8 +132,8 @@ type DataTable = {
   setFiltersInBulk: (colKey: S, filters: S[]) => void
 }
 
-type ContextualMenuState = {
-  onFilterChange: (filterKey: S, filterVal: S, checked?: boolean) => void
+type ContextualMenuProps = {
+  onFilterChange: (filterKey: S, filterVal: S, checked?: B) => void
   col: WaveColumn
   listProps: Fluent.IContextualMenuListProps
   selectedFiltersRef: React.MutableRefObject<Dict<S[]> | null>
@@ -217,26 +217,26 @@ const
     const line = JSON.stringify(row)
     return line.substr(1, line.length - 2)
   }).join('\n'),
-  ContextualMenu = ({ onFilterChange, col, listProps, selectedFiltersRef, setFiltersInBulk }: ContextualMenuState) => {
+  ContextualMenu = ({ onFilterChange, col, listProps, selectedFiltersRef, setFiltersInBulk }: ContextualMenuProps) => {
     const
       isFilterChecked = (data: S, key: S) => !!selectedFiltersRef.current && selectedFiltersRef.current[data]?.includes(key),
-      [filters, setFilters] = React.useState(col.cellType?.tag
+      [menuFilters, setMenuFilters] = React.useState(col.cellType?.tag
         ? Array.from(listProps.items.reduce((_filters, { key, text, data }) => {
-          key.split(',').forEach(key => _filters.set(key, { key, text, data }))
+          key.split(',').forEach(key => _filters.set(key, { key, text, data, checked: isFilterChecked(data, key) }))
           return _filters
         }, new Map<S, Fluent.IContextualMenuItem>()).values())
         : listProps.items.map(i => ({ ...i, checked: isFilterChecked(i.data, i.key) }))
       ),
       selectAll = () => {
-        setFilters(filters.map(i => ({ ...i, checked: true })))
-        setFiltersInBulk(col.key, filters.map(f => f.key))
+        setMenuFilters(menuFilters.map(i => ({ ...i, checked: true })))
+        setFiltersInBulk(col.key, menuFilters.map(f => f.key))
       },
       deselectAll = () => {
-        setFilters(filters.map(i => ({ ...i, checked: false })))
+        setMenuFilters(menuFilters.map(i => ({ ...i, checked: false })))
         setFiltersInBulk(col.key, [])
       },
-      getOnFilterChangeHandler = (data: S, key: S) => (_ev?: React.FormEvent<HTMLInputElement | HTMLElement>, checked?: boolean) => {
-        setFilters(filters => filters.map(f => f.key === key ? ({ ...f, checked }) : f))
+      getOnFilterChangeHandler = (data: S, key: S) => (_ev?: React.FormEvent<HTMLInputElement | HTMLElement>, checked?: B) => {
+        setMenuFilters(filters => filters.map(f => f.key === key ? ({ ...f, checked }) : f))
         onFilterChange(data, key, checked)
       }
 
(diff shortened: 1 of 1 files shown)


mturoci:
Thanks @aalencar. Please have a look at `feat/alternative-issue-602` branch for an alternative, non-hacky, solution and let me know what you think

aalencar:
Thanks for providing an alternative, @mturoci !

I like the idea of having a new component for Contextual Menu and not using `setTimeout` to trigger the changes.

But I noticed that the current solution is broken when we have 2 filterable columns. The bug is in `onFilterChange`. Working on a fix.

aalencar:
@mturoci I fixed the issue above, and added other changes to `feat/alternative-issue-602`.

One of the missing fixes was persisting the checkbox status after the contextual menu was closed. For that we needed to set the `checked` property of each item in the `filters` state upon initialisation. This raises a problem: by setting `checked` we now have a controlled component and we need to set ContextualMenu state when the checkboxes are clicked. To achieve that I passed a callback to `onFilterChange` that has the access to ContextualMenu `filters` used to manage the checkbox state.

Let me know what you think of the solution.

aalencar:
I just realized that I might have overlooked the logic for tags. Taking a look at it.

mturoci:
@aalencar if you decided to take the approach proposed in `feat/alternative-issue-602`, can you please also switch branch for this PR? The review will be easier that way.

aalencar:
@mturoci done.

mturoci:
> To achieve that I passed a callback to onFilterChange that has the access to ContextualMenu filters used to manage the checkbox state.

This seems to be way more complex than necessary. IIUC, what you do basically is:

* checkbox is clicked
* call parent `onFilterChange`
* from within top-level component (`onFilterChange`) call the callback to update context menu state

I would suggest a simpler approach, which doesn't require any callback:

* checkbox is clicked
* update menu state right away, then call original `onFilterChange` callback

Or maybe I am missing something?

aalencar:
> update menu state right away, then call original onFilterChange callback

`onFilterChange()` is not the event handler but instead creates a custom handler for each filter.

We can't update right away because we have to update the state from within the handler which is returned by `onFilterChange`, and that's why we also pass a callback to `onFilterChange`.

mturoci:
> We can't update right away because we have to update the state from within the handler which is returned by onFilterChange, and that's why we also pass a callback to onFilterChange.

I still don't understand, why would we have to update the state within the handler and not directly? You can refactor the code, no need to stick with what is there currently.

aalencar:
> why would we have to update the state within the handler and not directly

Because we want to update the ContextualMenu's state on click and that has to be done from within the handler.

aalencar:
@mturoci applied your suggestions. Ready to be reviewed again.

ui/src/table.tsx
```diff
@@ -119,7 +119,7 @@ type WaveColumn = Fluent.IColumn & {
 
 type DataTable = {
   model: Table
-  onFilterChange: (filterKey: S, filterVal: S) => (e?: React.FormEvent<HTMLInputElement | HTMLElement>, checked?: B) => void
+  onFilterChange: (filterKey: S, filterVal: S, checked?: boolean) => void
```
mturoci, line 122: nit
```suggestion
  onFilterChange: (filterKey: S, filterVal: S, checked?: B) => void
```

ui/src/table.tsx
```diff
@@ -208,7 +217,51 @@ const
...
     return line.substr(1, line.length - 2)
   }).join('\n'),
-  DataTable = ({ model: m, onFilterChange, items, filteredItems, selection, selectedFilters, isMultiple, isSearchable, groups, sort, reset }: DataTable) => {
+  ContextualMenu = ({ onFilterChange, col, listProps, selectedFiltersRef, setFiltersInBulk }: ContextualMenuState) => {
+    const
+      isFilterChecked = (data: S, key: S) => !!selectedFiltersRef.current && selectedFiltersRef.current[data]?.includes(key),
+      [filters, setFilters] = React.useState(col.cellType?.tag
+        ? Array.from(listProps.items.reduce((_filters, { key, text, data }) => {
+          key.split(',').forEach(key => _filters.set(key, { key, text, data }))
+          return _filters
```
mturoci, line 226: Didn't check, but it seems like it's missing handling `checked`.

ui/src/table.tsx
```diff
@@ -208,7 +217,51 @@ const
...
+      ),
+      selectAll = () => {
+        setFilters(filters.map(i => ({ ...i, checked: true })))
+        setFiltersInBulk(col.key, filters.map(f => f.key))
+      },
+      deselectAll = () => {
+        setFilters(filters.map(i => ({ ...i, checked: false })))
+        setFiltersInBulk(col.key, [])
+      },
+      getOnFilterChangeHandler = (data: S, key: S) => (_ev?: React.FormEvent<HTMLInputElement | HTMLElement>, checked?: boolean) => {
```
mturoci, line 238: Odd naming, handlers usually start with on rather than get.

nit
```suggestion
      getOnFilterChangeHandler = (data: S, key: S) => (_ev?: React.FormEvent<HTMLInputElement | HTMLElement>, checked?: B) => {
```

ui/src/table.tsx
```diff
@@ -543,26 +579,25 @@ export const
         setGroupByKey(option.key as S)
         initGroups()
       },
-      onFilterChange = React.useCallback((filterKey: S, filterVal: S) => (_e?: React.FormEvent<HTMLInputElement | HTMLElement>, checked?: B) => {
+      onFilterChange = React.useCallback((filterKey: S, filterVal: S, checked) => {
```
mturoci, line 582: Missing type for checked

ui/src/table.tsx
```diff
@@ -208,7 +217,51 @@ const
     const line = JSON.stringify(row)
     return line.substr(1, line.length - 2)
   }).join('\n'),
-  DataTable = ({ model: m, onFilterChange, items, filteredItems, selection, selectedFilters, isMultiple, isSearchable, groups, sort, reset }: DataTable) => {
+  ContextualMenu = ({ onFilterChange, col, listProps, selectedFiltersRef, setFiltersInBulk }: ContextualMenuState) => {
+    const
+      isFilterChecked = (data: S, key: S) => !!selectedFiltersRef.current && selectedFiltersRef.current[data]?.includes(key),
+      [filters, setFilters] = React.useState(col.cellType?.tag
```
mturoci, line 223: Would go for `menuFilters` or something like that so that reader is sure which ones are meant.

mturoci (changes requested):
Thanks @aalencar. A few minor comments. I didn't try your solution yet, will do tmr.

ui/src/table.tsx
aalencar: Not a handler, it's a high-order function that creates a handler, hence the name

ui/src/table.tsx
aalencar: Good catch.

ui/src/table.tsx
aalencar: Indeed, surely more meaningful.

ui/src/table.tsx
mturoci: Agree, fair point.

ui/src/table.tsx
```diff
@@ -543,26 +579,25 @@ export const
         setGroupByKey(option.key as S)
         initGroups()
       },
-      onFilterChange = React.useCallback((filterKey: S, filterVal: S) => (_e?: React.FormEvent<HTMLInputElement | HTMLElement>, checked?: B) => {
+      onFilterChange = React.useCallback((filterKey: S, filterVal: S, checked?: B) => {
         setSelectedFilters(selectedFilters => {
-          const filters = selectedFilters || {}
-
+          let filters: Dict<S[]> | null = null
+          filters = selectedFilters || {}
```
mturoci, line 585: What's the reason for this change?

mturoci (approved):
LGTM @aalencar! Just one minor comment from me. Also, let's merge this after `0.20` is released.

ui/src/table.tsx
aalencar: It was probably a mistake during the refactoring. Reverted.

PR #1197, opened 2022-01-19
Icons replaced with fluentui/react-icons-mdl2

marek-mihok:
This PR addresses the replacement of Fluent UI default `font-icons-mdl2` icons with `@fluentui/react-icons-mdl2`.

It also brings Icons section in Wave docs pages.

https://user-images.githubusercontent.com/23740173/154069154-25d21eb1-d303-4fc7-acb2-50f3ebf843e4.mov


ui/src/index.tsx
```diff
@@ -12,21 +12,37 @@
...
+
+const iconsToRegister = (reactIcons as IconMapping[]).reduce(
+  (acc: { [key: string]: React.FunctionComponentElement<React.FC> }, cur) => {
+    acc[cur.substring(0, cur.length - 4)] = React.createElement<React.FC>(Icons[cur] as React.FC)
+    return acc
+  }, {})
+
+registerIcons({
+  icons: iconsToRegister
+})
```
mturoci, line 45: This should be cleaner. Can you verify this works as expected @marek-mihok? Also note that `icons` prop in `registerIcons` accepts `string | JSX.Element` so that your original typing was a bit misleading (using React.FC instead).
 
```suggestion
const icons = Object.entries(Icons).reduce((acc, [iconName, iconComponent]) => {
  if (iconComponent.name === 'Component') acc[iconName.slice(0, -4)] = React.createElement(iconComponent as React.FC)
  return acc
}, {} as { [key: string]: JSX.Element })
registerIcons({ icons })
```

ui/src/index.tsx
```diff
@@ -12,21 +12,37 @@
...
+import { loadTheme, registerIcons } from '@fluentui/react'
 import React from 'react'
 import ReactDOM from 'react-dom'
 import './cards'
 import './index.scss'
 import Router from './router'
 import * as serviceWorker from './serviceWorker'
 import { defaultTheme } from './theme'
+import * as Icons from '@fluentui/react-icons-mdl2'
+
```
mturoci, line 24:
```suggestion
```

mturoci (changes requested):
Thanks @marek-mihok! 

* Please also update documentation links (https://uifabricicons.azurewebsites.net/) to point to the newly added icon subset (non-branded) instead of the old one. (Better to wait for #1206 to save yourself from plenty of merge conflicts)
* Update https://github.com/h2oai/wave/blob/78c3da6941d3d168cf4c8cfbbac9a9de5370c9a5/tools/intellij-plugin/src/main/java/org/h2o/utils/UtilsResources.java#L21 with the newest icons.

mturoci (dismissed):
Thanks @marek-mihok!

For the sake of time, I committed the requested changes right away instead of proper code review since this PR is blocking #1242.

My changes include:

* general refactor
* improving sluggish searchbox rendering performance (laggy typing) caused by filtering and rendering all the icons on every single key stroke. This is now fixed using debounce and proper React component for icons that rerenders only when filteredIcons list changes.
* fixed ui package-lock that was out of sync with package.json causing `npm ci` to fail
* added fuzzy search for better UX

Please check out my changes and verify I didn't break anything. Once that is done, this PR is ready to be merged.

mturoci (approved):
Thanks @marek-mihok! Great job!

PR #1207, opened 2022-01-25
feat: Add overflow attr to table row.

### LINKED ISSUE #515: Allow text wrapping in string table cell
via @mtanco 

@lo5 The only thing I am wondering about is where should we put the switch within API. Directly on `TableColumn` as `wrap`? This would however apply to text types only. Another option would be to wrap by default and swap it with current elipsis.

marek-mihok:
Adds optional `overflow` property to `ui.table_column` which defines how to handle the long text that does not fit into the cell. 

Possible values are:

- `tooltip`
_Shows a tooltip in case text does not fit into the cell_

https://user-images.githubusercontent.com/23740173/150987809-f6066045-58f5-465c-8739-417ea0f2b254.mov

- `wrap`
_Wraps text into multiple lines in case text does not fit into the cell_

image

Closes #515


ui/src/table.tsx
```diff
@@ -392,6 +397,22 @@ const
...
+            '::after': {
+              content: '',
+              display: 'block',
+            }
+          },
+        }
+
+        const TooltipWrapper = ({ children }: { children: string }) => {
+          if (col.overflow === 'tooltip') return <TooltipHost styles={hostStyles} id={item.key as string} content={children} overflowMode={TooltipOverflowMode.Parent} title={children}>{children}</TooltipHost>
+          if (col.overflow === 'wrap') return <span style={{ whiteSpace: col.overflow === 'wrap' ? 'normal' : undefined }}>{children}</span>
```
mturoci, line 412: We already know that overflow === 'wrap' so:
```suggestion
          if (col.overflow === 'wrap') return <span style={{ whiteSpace: 'normal' }}>{children}</span>
```

mturoci (changes requested):
Thanks @marek-mihok. Please also add a docs section about the table row overflow. Use `wrap` as an example there since there is currently no way to simulate hover during screenshot creation.

mturoci (dismissed):
Thanks, @marek-mihok! LGTM.

@lo5 could you have a quick look at the proposed API, please?

```ts
 /** Defines how to handle the long text that does not fit into the cell. */
overflow?: 'tooltip' | 'wrap'
```

ui/src/table.tsx
```diff
@@ -54,6 +54,8 @@ interface TableColumn {
   data_type?: 'string' | 'number' | 'time'
   /** Defines how to render each cell in this column. Renders as plain text by default. */
   cell_type?: TableCellType
+  /** Defines how to handle the long text that does not fit into the cell. */
```
lo5, line 57:
```suggestion
  /** Defines what to do with a cell's contents in case it does not fit inside the cell. */
```

ui/src/table.tsx
```diff
@@ -54,6 +54,8 @@ interface TableColumn {
...
+  overflow?: 'tooltip' | 'wrap'
```
lo5, line 58:
```suggestion
  cell_overflow?: 'tooltip' | 'wrap'
```

lo5:
@marek-mihok - sorry, hit the approve button by mistake - needs changes.

mturoci (changes requested):
Since there are no dependency changes in `package.json`, there should not be any changes in `package-lock.json` either. Remove the changes and the PR is good to go.

PR #1256, opened 2022-02-21
fix: Kill hanging waved process when app fails to start.

mturoci:
* Kill waved server process created during `wave run` if the app exits with failure.
* Had to update uvicorn to `0.16` due to https://github.com/encode/uvicorn/pull/1278 - uvicorn previously returned 0 exit code no matter what meaning there was no way of distinguishing when the app failed.
* Minor: Remove unnecessary `certifi` dep.


lo5 (approved):
Thanks @mturoci!

PR #1262, opened 2022-02-22
feat: Table custom groups with ability to collapse/expand row

### LINKED ISSUE #961: Collapsible table
via @sandruparo 

## Design
image
[design URL](https://xd.adobe.com/view/eebb600c-026b-4ca1-bb6b-0a439235452b-bbe8/specs/)

## Features
* Ability to collapse / expand a particular row
* Ability to collapse all
* ~~Tags column support~~
* ~~Text color column support~~

Needs discussion: How important is the row collapse / expand icon with the lines design (img below) as opposed to default [Fluent one](https://developer.microsoft.com/en-us/fluentui#/controls/web/groupedlist)? Asking since it is not impossible, but can take more time.

@lo5 I feel like this could be just an extension to our existing `ui.table` component. Wdyt? This way, we could leverage search / sort / filter and even maybe download if that makes sense. The table would display normal / grouped mode based on data provided - either `rows` or a new prop `groups` that would contain `rows` per each group.

marek-mihok:
This PR comes with a new TableGroup API for creating custom groups:
```ts
interface TableGroup {
  /** The title of the group. */
  label: S
  /** The rows in this group. */
  rows: TableRow[]
  /** Indicates whether the table group should be collapsed by default. Defaults to True. */
  collapsed?: B
}
```

Also there is a new `groups` prop in Table API for passing the array of groups created with a new TableGroup API:
```ts
export interface Table {
  name: Id
  columns: TableColumn[]
  rows?: TableRow[]
  /** True to allow multiple rows to be selected. */
  multiple?: B
  /** True to allow group by feature. Ignored if `groups` are specified. */
  groupable?: B
  /** Indicates whether the contents of this table can be downloaded and saved as a CSV file. Defaults to False. */
  downloadable?: B
  resettable?: B
  height?: S
  width?: S
  values?: S[]
  visible?: B
  tooltip?: S
  groups?: TableGroup[]
}
```
_Note:_ `rows` are now mutually exclusive with `groups` and `groupable` is ignored when `groups` are specified.

image

Closes #961


ui/src/table.tsx
```diff
@@ -453,17 +497,24 @@ export const
...
           groupedByKeys = Object.keys(groupedBy),
-          groups: Fluent.IGroup[] = groupedByKeys.map((key, i) => {
-            if (i !== 0) {
-              const prevKey = groupedByKeys[i - 1]
-              prevSum += groupedBy[prevKey].length
-            }
+          groups: Fluent.IGroup[] = m.groups
+            ? filteredItems.reduce((acc, { group }, idx) => {
+              const lastG = acc[acc.length - 1]
+              if (lastG?.key === group) lastG.count = lastG.count + 1
```
mturoci, line 503:
```suggestion
              if (lastG?.key === group) lastG.count++
```

ui/src/table.tsx
```diff
@@ -416,27 +431,56 @@ const
...
+              return { ...getItem(r), group: label }
+            }))
+            acc.push(...item)
+            return acc
+          }, [] as (Fluent.IObjectWithKey & Dict<any> & { group?: string })[])
+          : (m.rows || []).map(r => {
+            return getItem(r)
+          })
+        , [m.rows, m.groups, getItem]),
+      customGroups = React.useMemo(() => {
```
mturoci, line 469: What's the difference between groups vs customGroups?

ui/src/table.tsx
```diff
@@ -416,27 +431,56 @@ const
...
+        return m.groups ? m.groups.reduce((acc, { rows, label, collapsed = true }, idx) => {
```
mturoci, line 470: Use implicit return.

mturoci (changes requested):
Thanks @marek-mihok! We will have another round of review another this one.

py/examples/table_groups.py
```diff
@@ -0,0 +1,62 @@
...
+        text=fake.sentence(),
+        status=('Closed' if i % 2 == 0 else 'Open'),
+        created=fake.iso8601()) for i in range(30)
+]
+
+# Create columns for our issue table.
+columns = [
+    ui.table_column(name='text', label='Issues reported by', sortable=True, searchable=True, max_width='300'),
+    ui.table_column(name='status', label='Status', filterable=True),
+    ui.table_column(name='created', label='Created', sortable=True, data_type='time'),
```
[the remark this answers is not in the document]
marek-mihok: The purpose of extra columns was to demonstrate how filtering and sorting works with custom groups.

mturoci (changes requested):
A few warnings that need to be fixed:

image

* Add your `table_groups.py` example to Tour app via `tour.conf`.
* Add an example screenshot to `webiste/docs/examples/assets`. (make sure `make website` passes without errors)
* Update PR description with the new API that needs to be reviewed.

ui/src/table.test.tsx
```diff
@@ -642,9 +642,10 @@ describe('Table.tsx', () => {
       fireEvent.click(getAllByText('Col2')[1]!)
       fireEvent.click(container.querySelector('.ms-GroupHeader-expand')!)
 
-      expect(getAllByRole('row')).toHaveLength(tableProps.rows.length + headerRow + groupByRow)
+      expect(getAllByRole('row')).toHaveLength(tableProps.rows!.length + headerRow + groupByRow)
       fireEvent.change(getByTestId('search'), { target: { value: cell21 } })
-      expect(getAllByRole('row')).toHaveLength(headerRow + groupByRow)
+      const filteredItems = tableProps.rows!.filter(row => row.cells.find(cell => cell === cell21)).length
```
mturoci, line 647: No need, we know there is just a single row with that value, you can hardcode it.

mturoci (changes requested):
A few more comments

* data download is not working when custom groups are set
* it would be good to also add unit tests to make sure search/filter/sort etc. work properly even when custom groups are specified.

ui/src/table.test.tsx
marek-mihok: Do you mean like this `expect(getAllByRole('row')).toHaveLength(headerRow + groupByRow + 1)`or should I declare a constant in the top of the file similar to `headerRow` or `groupByRow`so that someone in a future can easily find what   does value 1 stand for?

ui/src/table.tsx
```diff
@@ -678,23 +745,24 @@ export const
...
+          {groupable && <Fluent.Dropdown data-test='groupby' label='Group by' selectedKey={groupByKey} onChange={onGroupByChange} options={groupByOptions} styles={{ root: { width: 300 } }} />}
           {!!searchableKeys.length && <Fluent.SearchBox data-test='search' placeholder='Search' onChange={onSearchChange} value={searchStr} styles={{ root: { width: '50%', maxWidth: 500 } }} />}
         </Fluent.Stack>
         <Fluent.ScrollablePane
           scrollbarVisibility={Fluent.ScrollbarVisibility.auto}
           styles={{
-            root: { top: m.groupable || searchableKeys.length ? 80 : 0, bottom: shouldShowFooter ? 46 : 0 },
-            stickyAbove: { right: important('0px'), border: border(2, 'transparent') },
+            root: { top: groupable || searchableKeys.length ? 80 : 0, bottom: shouldShowFooter ? 46 : 0 },
+            stickyAbove: { right: important('0px'), border: border(2, 'transparent'), zIndex: 2 },
```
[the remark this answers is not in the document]
mturoci: Got it, makes sense.

ui/src/table.tsx
```diff
@@ -448,43 +513,69 @@ export const
...
+                return acc
+              }, {} as { [key: S]: B })
+            }
+            const prevGroup = acc[acc.length - 1]
+            prevGroup?.key === group
+              ? prevGroup.count++
+              : acc.push({ key: group, name: group, startIndex: idx, count: 1, isCollapsed: getIsCollapsed(group) })
+            return acc
+          }, [] as Fluent.IGroup[])
+          if (collapsedRefs.current === undefined) collapsedRefs.current = null
```
mturoci, line 540: Why is this here?

mturoci (changes requested):
Some more comments from me. @marek-mihok prior to addressing the review comments, please also add unit tests that check if table is in correct collapsed/expanded state during sort/search etc. Once you have these tests in place, you will be able to do the refactor and general development much faster as you will no longer need to manually test out stuff.

mturoci (approved):
Thanks @marek-mihok, LGTM!

cc @lo5 for the final API review

mturoci:
@marek-mihok please resolve conflicts

lo5 (approved):
API looks good - thanks @marek-mihok!

PR #1281, opened 2022-03-04
feat: Server-side table

### LINKED ISSUE #730: Add pagination support to table component
Would be a cool feature to have pagination within a table. My use case is specifically pandas, but could see how any table would benefit. I apologize if I this exists and I have missed it, didn't see it in the docs anywhere! Please do let me know if this exists and I am over looking it. Thanks!

mturoci:
## Demo

https://user-images.githubusercontent.com/64769322/156760855-b86176fe-e699-4010-92a7-4727ea69af67.mov

## Features

* Server side pagination, sort, filter, search, download, reset.
* All of the above are firing an event that needs to be handled by developer and send the new data.

Once the general ideas are discussed and approved, I will add more fine-grained usage examples/docs.

## Needs discussion

* what should we advise people to display when group by is clicked?
* How important is having infinite-scroll? See [this comment](https://github.com/h2oai/wave/issues/730#issuecomment-1059072043). It's probably doable, just not so easily and would take more time than originally expected.
* Currently, if pagination is set, sort/search/filter etc. events are fired automatically. They do not require manual registration compared to the rest of Wave components that support `events=[]` attr. The reason is simple, if you want to have these features enabled, you have no choice but register these events anyway. The question though is if that's fine regarding consistency.
* The example is the simplest I could think of, but maybe we could use a more real-world example, e.g. pd.dataframe etc.

## API review

```ts
interface TableColumn {
   ...
  /** List of values to allow filtering by, needed when pagination is set. Only applicable to filterable columns. */
  filters?: S[]
}

/** Creates a paginated table. Useful for large amounts of data (1M+ rows). */
interface TablePagination {
  /** Renders a progress arc with a percentage value in the middle. */
  total_rows: U
  /** Renders a progress arc with a percentage value in the middle. */
  rows_per_page: U
}
```

## API in action

```py
from copy import deepcopy
import csv


# Create a dummy data blueprint.
class Issue:
    def __init__(self, text: str, status: str):
        self.text = text
        self.status = status


all_issues = [Issue(text=i + 1, status=('Closed' if i % 2 == 0 else 'Open')) for i in range(100)]
rows_per_page = 10
total_rows = len(all_issues)


def get_table_rows(q: Q):
    # Make a deep copy in order to not mutate the original `all_issues` which serves as our baseline.
    rows = deepcopy(all_issues)
    if q.client.sort:
        # Sort by multiple columns.
        for col, reverse in q.client.sort.items():
            rows.sort(key=lambda i: getattr(i, col), reverse=reverse)
    # Filter out all rows that do not contain searched string in `text` cell.
    # You can adjust searched cells as you see fit.
    if q.client.search:
        rows = [i for i in rows if q.client.search in str(i.text)]
    # Filter out rows that do not contain filtered column value.
    if q.client.filters:
        for col, filters in q.client.filters.items():
            rows = [row for row in rows if not filters or any(f in getattr(row, col) for f in filters)]

    # Update table pagination according to the new row count.
    if q.client.search is not None or q.client.filters:
        q.page['form'].items[0].table.pagination = ui.table_pagination(len(rows), rows_per_page)

    offset = q.client.page_offset or 0
    rows = rows[offset:offset + rows_per_page]
    return [ui.table_row(name=str(i.text), cells=[str(i.text), i.status]) for i in rows]


@app('/demo')
async def serve(q: Q):
    if not q.app.initialized:
        # Create and upload a CSV file for downloads.
        with open('data_download.csv', 'w') as csvfile:
            csv_writer = csv.writer(csvfile, delimiter=',')
            for i in all_issues:
                csv_writer.writerow([i.text, i.status])
        q.app.data_download, = await q.site.upload(['data_download.csv'])
        q.app.initialized = True

    if not q.client.initialized:
        q.page['form'] = ui.form_card(box='1 1 -1 -1', items=[
            ui.table(
                name='table',
                columns=[
                    ui.table_column(name='text', label='Text', sortable=True, searchable=True),
                    ui.table_column(name='status', label='Status', filterable=True),
                ],
                rows=get_table_rows(q),
                resettable=True,
                downloadable=True,
                pagination=ui.table_pagination(total_rows=len(all_issues), rows_per_page=rows_per_page),
                # Make sure to register the necessary events for the feature you want to support, e.g. sorting.
                # All the registered events have to be handled by the developer.
                # `page_change` event is required to be handled for pagination to work.
                events=['sort', 'filter', 'search', 'page_change', 'download', 'reset']
            )
        ])
        q.client.initialized = True

    # Check if user triggered any table action and save it to local state for allowing multiple
    # actions to be performed on the data at the same time, e.g. sort the filtered data etc.
    if q.events.table:
        if q.events.table.sort:
            q.client.sort = q.events.table.sort
            q.client.page_offset = 0
        if q.events.table.filter:
            q.client.filters = q.events.table.filter
            q.client.page_offset = 0
        if q.events.table.search is not None:
            q.client.search = q.events.table.search
            q.client.page_offset = 0
        if q.events.table.page_change:
            q.client.page_offset = q.events.table.page_change.get('offset', 0)
        if q.events.table.reset:
            q.client.search = None
            q.client.sort = None
            q.client.filters = None
            q.client.page_offset = 0
            q.page['form'].items[0].table.pagination = ui.table_pagination(total_rows, rows_per_page)
        if q.events.table.download:
            q.page['meta'].script = ui.inline_script(f'window.open("{q.app.data_download}")')

        q.page['form'].items[0].table.rows = get_table_rows(q)

    await q.page.save()
```

cc @lo5 @mtanco 

Closes #730


lo5 (commented):
> what should we advise people to display when group by is clicked?

Maybe it's better to disable group-by if pagination is enabled (to avoid confusing behavior)?

> How important is having infinite-scroll? See https://github.com/h2oai/wave/issues/730#issuecomment-1059072043. It's probably doable, just not so easily and would take more time than originally expected.

Not important. The implementation in this PR is good enough for now.

> Currently, if pagination is set, sort/search/filter etc. events are fired automatically. They do not require manual registration compared to the rest of Wave components that support events=[] attr. The reason is simple, if you want to have these features enabled, you have no choice but register these events anyway. The question though is if that's fine regarding consistency.

I agree that the author must handle those events, but it's better to be explicit and raise events only for registered events (as is the case with other components). 

If a component raises non-registered events, then it can lead to head-scratching moments where an event are not trapped by any existing  `if-elif-else` case. Secondly, if we introduce another table event (say, group-by) in a future release and automatically raise it, it might break existing code or cause unintended effects.

It would be good to add a note to the table's docs indicating that the events *must* be handled.

mturoci:
>I agree that the author must handle those events, but it's better to be explicit and raise events only for registered events (as is the case with other components).

Agree, will do.

mturoci:
Disabled group by when pagination is set, added explicit events registration, added examples, updated widgets docs. Ready for the final review cc @lo5 @mtanco

py/examples/table_pagination_pandas.py
```diff
@@ -0,0 +1,98 @@
...
+        q.app.initialized = True
+
+    if not q.client.initialized:
+        q.page['meta'] = ui.meta_card(box='')
+        q.page['form'] = ui.form_card(box='1 1 -1 -1', items=[
+            ui.table(
+                name='table',
+                columns=[
+                    ui.table_column(name='text', label='Text', sortable=True, searchable=True),
+                    ui.table_column(name='status', label='Status', filterable=True),
```
mtanco, line 59: Running this example, filter is not working. The arrow shows for this column but clicking it does not do anything.

py/examples/table_pagination_pandas.py
```diff
@@ -0,0 +1,98 @@
...
+        if q.events.table.page_change:
+            q.client.page_offset = q.events.table.page_change.get('offset', 0)
+        if q.events.table.reset:
+            q.client.search = None
+            q.client.sort = None
+            q.client.filters = None
+            q.client.page_offset = 0
+            q.page['form'].items[0].table.pagination = ui.table_pagination(total_rows, rows_per_page)
+        if q.events.table.download:
+            q.page['meta'].script = ui.inline_script(f'window.open("{q.app.data_download}")')
```
mtanco, line 94: I expect users to want to download the table how it is sorted right now and with any filters

py/examples/table_pagination_pandas.py
```diff
@@ -0,0 +1,98 @@
...
+        q.app.data_download, = await q.site.upload(['data_download.csv'])
+        q.app.initialized = True
+
+    if not q.client.initialized:
+        q.page['meta'] = ui.meta_card(box='')
+        q.page['form'] = ui.form_card(box='1 1 -1 -1', items=[
+            ui.table(
+                name='table',
+                columns=[
+                    ui.table_column(name='text', label='Text', sortable=True, searchable=True),
```
mtanco, line 58: In the example let's make `link=False` so this column is not clickable OR handle clicking the row in the app

py/examples/table_pagination_pandas.py
```diff
@@ -0,0 +1,98 @@
...
+
+all_issues_df = pd.DataFrame(
+    [[i + 1, 'Closed' if i % 2 == 0 else 'Open'] for i in range(100)],
+    columns=['text', 'status']
+)
+rows_per_page = 10
+total_rows = len(all_issues_df)
+
+
+def get_table_rows(q: Q):
```
mtanco, line 18: It would be great to have this a function that would work for any `ui.table`

mtanco (commented):
This looks great! A couple usability thoughts

py/examples/table_pagination_pandas.py
mturoci: Agree, done.

py/examples/table_pagination_pandas.py
mturoci: Good catch, already fixed in another PR, need to rebase this one and should be ok.

py/examples/table_pagination_pandas.py
mturoci: Agree, done.

py/examples/table_pagination_pandas.py
mturoci: Agree, done.

mturoci:
@mwysokin yep, this feature will basically only tell you when the sort/filter/search/download/reset/change_page was triggered and with what values (in case of sort - asc/desc direction). After that, you are free to do whatever you wish as long as you return new data rows to be displayed. The sorting/fitlering logic is 100% up to the dev. We will provide some boilerplate to use that we consider good default, but you are free to code your own solution.

mturoci:
Will be part of this PR, planned for 0.21. We would like to make a 0.21-rc by the end of this week. Stay tuned!

mtanco:
@mturoci from what I can tell, the `filter` option only shows the unique options for the `n` rows that the pagination knows about. Should there be some way to give all unique values for a column? Or am I just missing something?

mtanco:
I am noticing from logging that every search is called twice - let me know if you need a demo / repro

mturoci:
> Should there be some way to give all unique values for a column?

Good call, I need to think about it.

> I am noticing from logging that every search is called twice 

Hm.. I don't see this behavior, demo would be nice.

mtanco:
> Hm.. I don't see this behavior, 
With the latest wheel, me neither, this seems to have been implicitly fixed. I made not code change (other than to support that `q.client.search` is now a dictionary instead of a string

lo5 (dismissed):
Looks great! Thanks @mturoci!

mturoci:
@marek-mihok can you please do a bit of a QA on this PR once you are done with #1287 and make sure every table example works as expected (especially your custom group related stuff as I had to merge it)? I went through all of them and couldn't find anything, but I would like to have this gone through with a second pair of eyes.

ui/src/table.tsx
```diff
@@ -483,6 +531,96 @@ const
         {colContextMenuList && <Fluent.ContextualMenu {...colContextMenuList} />}
       </>
     )
+  }),
+  Pagination = ({ currentPage, onPageChange, pagination }: PaginationProps) => {
+    const
+      { total_rows, rows_per_page } = pagination,
+      lastPage = Math.ceil(total_rows / rows_per_page),
+      btnStyles: Fluent.IButtonStyles = { rootDisabled: { background: 'transparent' }, root: { marginLeft: -8 } },
+      isLastPage = currentPage === lastPage
```
marek-mihok, line 540:
```suggestion
      isLastPage = currentPage === lastPage || lastPage === 0
```
Pagination controls are not disabled in case no search results are found:
https://user-images.githubusercontent.com/23740173/161576060-e95cf06f-171a-478d-9013-e0d13b6ba9e4.mov

ui/src/table.tsx
```diff
@@ -692,29 +797,39 @@ export const
...
-            if (groups) initGroups()
-            return groups
-          })
+          if (m.pagination && m.events?.includes('filter')) {
+            wave.emit(m.name, 'filter', filters)
+            setCurrentPage(1)
+          } else {
+            filter(filters)
+            search()
+            setGroups(groups => {
```
marek-mihok, line 806: Is user able to use custom groups with custom filter implementation if `setGroups` is inside `else` part of a condition?

ui/src/table.tsx
```diff
@@ -744,22 +861,28 @@ export const
...
-            return groups
-          })
+          if (m.pagination && m.events?.includes('filter')) {
+            wave.emit(m.name, 'filter', newFilters)
+            setCurrentPage(1)
+          }
+          else {
+            filter(newFilters)
+            search()
+            setGroups(groups => {
```
marek-mihok, line 878: The same here.

marek-mihok (changes requested):
Thanks @mturoci! All examples look good except one issue that is necessary to be fixed.

ui/src/table.tsx
mturoci: Good catch!

ui/src/table.tsx
mturoci: Yes, groups have to be provided server side. This piece of code is for client-side, which does not matter in this context.

PR #1283, opened 2022-03-07
fix: make table filter icon clickable

aalencar:
Fixes the click handler for Chevron Down icon on Table component. Didn't spot the same incorrect behavior on other icons.

### DIFF
--- a/ui/src/table.tsx
+++ b/ui/src/table.tsx
@@ -263,7 +263,7 @@ const
       [colContextMenuList, setColContextMenuList] = React.useState<Fluent.IContextualMenuProps | null>(null),
       selectedFiltersRef = React.useRef(selectedFilters),
       onColumnClick = (e: React.MouseEvent<HTMLElement>, column: WaveColumn) => {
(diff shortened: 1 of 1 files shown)


ui/src/table.tsx
```diff
@@ -263,7 +263,7 @@ const
       [colContextMenuList, setColContextMenuList] = React.useState<Fluent.IContextualMenuProps | null>(null),
       selectedFiltersRef = React.useRef(selectedFilters),
       onColumnClick = (e: React.MouseEvent<HTMLElement>, column: WaveColumn) => {
-        const isMenuClicked = (e.target as HTMLElement).getAttribute('data-icon-name') === 'ChevronDown'
+        const isMenuClicked = (e.target as HTMLElement).closest('[data-icon-name=\'ChevronDown\']')
```
mturoci, line 266: Nit: No need for escaping.
```suggestion
        const isMenuClicked = (e.target as HTMLElement).closest('[data-icon-name="ChevronDown"]')
```

mturoci (approved):
Thanks @aalencar! Just a single small comment.

mturoci (approved):

PR #1293, opened 2022-03-11
feat: Allow gzip compression for static assets.

mturoci:


mturoci:
Thanks for review @geomodular and pointing me to a more idiomatic Go pattern!

gzip.go
```diff
@@ -0,0 +1,52 @@
...
+	New: func() interface{} {
+		return gzip.NewWriter(ioutil.Discard)
+	},
+}
+
+func (w *gzipResponseWriter) Write(b []byte) (int, error) {
+	return w.Writer.Write(b)
+}
+
+func GzipHandler(h http.Handler) http.Handler {
```
geomodular, line 40: nit: I'd make a verb from the function name and start it with a small letter.

geomodular (approved):
Thanks, @mturoci, looks nice.

gzip.go
mturoci: But starting it with a small letter would prevent exporting it, wouldn't it?

geomodular: It depends on the package. The package `wave` is the same for `gzip.go` as for `web_server.go` with no intention to use it elsewhere.

lo5 (approved):
Thanks @mturoci! Welcome to Go ;-)

PR #1297, opened 2022-03-15
fix: set the dropdown input to "width" prop

### LINKED ISSUE #1164: Dropdown width not respected when tooltip is set.
When I have a multichoice ui element, and select many values, the visual element exceeds the page limits:

image

I have set a maximum layout width, and also tried to control the width of `ui.dropdown`.
Maybe it is also just me doing something wrong here.

aalencar:
@mturoci 
Looks like the `width` prop on Dropdown was not being used. 

`Fluent.Dropdown` has a property called `dropdownWidth` which is used to control the size of the dropdown. I used `width` prop to control the size of the input instead. 

Example using `ui.dropdown` with tooltip and width = 200px:

With tooltip and not specifying the width:

@psinger is this the intended behavior?

Closes #1164

### DIFF
--- a/ui/src/tooltip.tsx
+++ b/ui/src/tooltip.tsx
@@ -17,6 +17,7 @@ import { B, S } from 'h2o-wave'
 import React from 'react'
 import { stylesheet } from 'typestyle'
 import { Markdown } from './markdown'
+import { cssVar } from './theme'
 
 const
   css = stylesheet({
@@ -29,7 +30,7 @@ const
       flexGrow: 1
     },
     icon: {
-      color: '#323130',
+      color: cssVar('$text'),
       fontWeight: 400,
       fontSize: '14px',
       userSelect: 'none',
@@ -44,7 +45,7 @@ const
   })
 
 export const
-  XToolTip = ({ children, content, showIcon, expand }: {
+  XToolTip = ({ children, content, showIcon = true, expand = true }: {
     children: React.ReactChild,
     content?: S,
     showIcon?: B,
@@ -52,17 +53,14 @@ export const
   }) => {
     if (!content) return <>{children}</>
 
-    const
-      tooltipProps: Fluent.ITooltipProps = {
-        onRenderContent: () => (<div><Markdown source={content} /></div>)
-      }
+    const tooltipProps: Fluent.ITooltipProps = { onRenderContent: () => <div><Markdown source={content} /></div> }
     return (
       <div className={css.container} data-test='tooltip'>
         {
-          showIcon === undefined || showIcon
+          showIcon
             ? (
               <>
-                <div className={expand === undefined || expand ? css.element : css.preventOverflow }>{children}</div>
+                <div className={expand ? css.element : css.preventOverflow}>{children}</div>
                 <Fluent.TooltipHost tooltipProps={tooltipProps}>
                   <Fluent.FontIcon className={css.icon} iconName='Info' />
                 </Fluent.TooltipHost>
--- a/ui/src/tooltip.tsx
+++ b/ui/src/tooltip.tsx
@@ -36,6 +36,10 @@ const
       textAlign: 'left',
       marginLeft: '0.5em',
       cursor: 'pointer'
+    },
+    preventOverflow: {
+      minWidth: 0,
+      flex: '1 1 0'
     }
   })
 
@@ -58,7 +62,7 @@ export const
           showIcon === undefined || showIcon
             ? (
               <>
-                <div className={expand === undefined || expand ? css.element : ''}>{children}</div>
+                <div className={expand === undefined || expand ? css.element : css.preventOverflow }>{children}</div>
                 <Fluent.TooltipHost tooltipProps={tooltipProps}>
                   <Fluent.FontIcon className={css.icon} iconName='Info' />
                 </Fluent.TooltipHost>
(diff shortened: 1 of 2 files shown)


mturoci:
> Looks like the width prop on Dropdown was not being used.

It's used [here](https://github.com/h2oai/wave/blob/5c113e91122396881288a287964169e8dcf31853/ui/src/form.tsx#L220). See https://github.com/h2oai/wave/commit/96b3898beee0e8fdd8fa3d4507b8759654d51be2. The reason it's handled in `form.tsx` is [here](https://github.com/h2oai/wave/blob/master/ui/src/theme.ts#L49-L51).

> With tooltip and not specifying the width:

The dropdown for this case should behave regardless of tooltip presence - not causing the overflow, expanding the whole width of the card.

psinger:
Yes, the dropdown should always fill the width of the card.

aalencar:
> > Looks like the width prop on Dropdown was not being used.
> 
> It's used [here](https://github.com/h2oai/wave/blob/5c113e91122396881288a287964169e8dcf31853/ui/src/form.tsx#L220). See [96b3898](https://github.com/h2oai/wave/commit/96b3898beee0e8fdd8fa3d4507b8759654d51be2). The reason it's handled in `form.tsx` is [here](https://github.com/h2oai/wave/blob/master/ui/src/theme.ts#L49-L51).
> 
> > With tooltip and not specifying the width:
> 
> The dropdown for this case should behave regardless of tooltip presence - not causing the overflow, expanding the whole width of the card.

Got it, I will make the changes.

aalencar:
File used to test:

```python


choices = [ui.choice(str(i), f'Option {i}') for i in range(1, 102)]


@app('/demo')
async def serve(q: Q):
    if not q.client.initialized:
        q.page['meta'] = ui.meta_card(box='', layouts=[
            ui.layout(
                # If the viewport width >= 0:
                breakpoint='xs',
                zones=[
                    ui.zone('form')
                ]
            )
        ])
        q.client.initialized = True
    if q.args.show_inputs:
        q.page['example'].items = [
            ui.text(f'dropdown_multi={q.args.dropdown_multi}'),
        ]
    else:
        q.page['example'] = ui.form_card(box='form', items=[
            ui.dropdown(name='dropdown_multi', label='No width and no Tooltip', values=['B', 'D'], required=True,
                        choices=choices, popup='never'),
            ui.dropdown(name='dropdown_multi', label='With tooltip', tooltip='some text', values=['B', 'D'], required=True,
                        choices=choices, popup='never'),
            ui.dropdown(name='dropdown_multi', width='200px', label='With width', values=['B', 'D'], required=True,
                        choices=choices, popup='never'),
            ui.dropdown(name='dropdown_multi', width='200px', tooltip='some text', label='With width and tooltip', values=['B', 'D'], required=True,
                        choices=choices, popup='never'),
        ])
    await q.page.save()

```

https://user-images.githubusercontent.com/15820796/160148033-09599da6-2740-43a3-8a58-55c8e80f902d.mov

psinger:
I assume not if the `width `is manually provided.

mturoci:
>Even when the width is provided

If width is provided, it should be respected, expand otherwise.

ui/src/tooltip.tsx
```diff
@@ -29,36 +30,37 @@ const
...
       fontWeight: 400,
       fontSize: '14px',
       userSelect: 'none',
       textAlign: 'left',
       marginLeft: '0.5em',
       cursor: 'pointer'
+    },
+    preventOverflow: {
+      minWidth: 0,
+      flex: '1 1 0'
```
mturoci, line 43: I am wondering what is the role of `expand` now, since both CSS classes add `flex-grow` now.

aalencar: I kept it to avoid regression but it seems that we can have `flex-grow: 1` in every component. `min-width` is what prevents overflow. I'm on it.

ui/src/tooltip.tsx
```diff
@@ -29,36 +30,36 @@ const
...
+    const tooltipProps: Fluent.ITooltipProps = { onRenderContent: () => <div><Markdown source={content} /></div> }
     return (
       <div className={css.container} data-test='tooltip'>
         {
-          showIcon === undefined || showIcon
+          showIcon
             ? (
               <>
-                <div className={expand === undefined || expand ? css.element : ''}>{children}</div>
+                <div className={clas(css.element, expand ? '' : css.preventOverflow )}>{children}</div>
```
mturoci, line 62: Shouldn't it be 
```suggestion
                <div className={expand ? css.element : css.preventOverflow}>{children}</div>
```
according to the previous logic?

Are all the other form components checked for possible regressions? Both with and without tooltip.

aalencar: > Shouldn't it be according to the previous logic?

Reverted.

> Are all the other form components checked for possible regressions? Both with and without tooltip.

Only the ones that use `expand=false`.

aalencar:
Reverted the last commit as I was testing with a really long word which causes it not to wrap.

mturoci:
Thanks @aalencar. I changed your solution a bit in order to not expand the element every time (`flex-grow: 1`) regardless of the `expanded` value. The difference between expanded=true and expanded=false is that the tooltip is either rendered at the end of the actual row (horizontal space) regardless of actual component width (expanded=true) or at the end of the component even if it doesn't fill all the width - like text for example (expanded=false).

PR #1301, opened 2022-03-21
docs: Make plots look normal in docs/widgets section.

### LINKED ISSUE #805: Axis title overlaps labels when labels are angled
#### Wave SDK Version, OS
v0.15.0 in Chrome on Mac

#### Actual behaviour
The axis title overlaps the labels when the labels are angled.

Screenshot 2021-05-11 at 1 18 05 PM

#### Expected behaviour
The axis title should be `below` the labels without any overlapping.

#### Steps To Reproduce
Any bar plot using longer label names and smaller card size.

mturoci:
This PR aims to better fit sample data for plot widgets in order to get natural-looking plot shapes.

https://user-images.githubusercontent.com/64769322/159261642-ea2dfb86-edca-497c-97c0-fc10233243e9.mov

## Needs discussion

As part of this PR, I also found a few subtle bugs that are already fixed in the most recent `g2` version. Here are the most notable ones (left - current, right - upgrade):

Seems like angling is no longer supported. (Can fix #805)
image

Polar coordinate plots do not render text ellipsis, but correct axis values instead.
image

Smooth path curves have changed a bit, but nothing serious.
image

https://github.com/h2oai/wave/discussions/1041 - This might be a problem as well although I suspect that an incorrect plot type was chosen for the underlying data rather than a bug.

@lo5 lmk if we are good with upgrading.

### DIFF
--- a/py/examples/wizard.py
+++ b/py/examples/wizard.py
@@ -1,47 +1,32 @@
 # Wizard
 # Create a multi-step #wizard using #form cards.
 # ---
-from h2o_wave import Q, ui, main, app, cypress, Cypress
+from h2o_wave import Q, ui, main, app, cypress, Cypress, data
 
 
 @app('/demo')
 async def serve(q: Q):
-    if not q.client.initialized:  # First visit, create an empty form card for our wizard
-        q.page['wizard'] = ui.form_card(box='1 1 2 4', items=[])
-        q.client.initialized = True
-
-    ui.button(name='default'),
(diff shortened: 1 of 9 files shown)


lo5 (approved):
New docs look good! Thanks @mturoci. I'm fine with upgrading g2.

mturoci:
Okay, will make the update in another PR so that we can easily revert in case of trouble.

PR #1311, opened 2022-03-29
Create table_pagination_h2o3.py

mtanco:
Based off of the `table_pagination_pandas.py` example but in theory this should work for any h2o-3 DataFrame with very minimal changes to the code. All changes end-users may want to make are marked with a `TODO`.

I am not sure if other files need to be updated when you make a new example.

### DIFF
--- a/py/examples/table_pagination_h2o3.py
+++ b/py/examples/table_pagination_h2o3.py
@@ -1,0 +1,225 @@
+# Table / Pagination / H2O-3 Dataframe
+# Use a paginated #table to display large (100m+ rows) tabular data using a H2O-3 dataframe.
+# #form #table #pagination #h2o3
+# ---
+
+import os
+from time import time
+
+import h2o
+from h2o_wave import Q, app, main, ui
+from loguru import logger
+
+
+@app("/")
+async def serve(q: Q):
+    logger.info(q.args)
+    logger.info(q.events)
+
+    if not q.app.initialized:
+        # This is called the first time our app runs
+        # Variables created here will be the same of all users of the app
+        # Save a direct link to our H2O Dataframe for all users to use throughout the app
+        h2o.connect(url="http://127.0.0.1:54321")
+        q.app.h2o_df = h2o.get_frame("py_6_sid_aff3")
+        
+        # EXAMPLE OF CREATING A LARGE DATAFRAME
+        # h2o_df = h2o.create_frame(
+        #     rows=1000000,
+        #     cols=5,
+        #     categorical_fraction=0.6,
+        #     integer_fraction=0,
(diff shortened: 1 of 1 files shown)


py/examples/table_pagination_h2o3.py
```diff
@@ -0,0 +1,225 @@
...
+
+        if q.events.h2o_table.reset:
+            logger.info("table reset")
+            q.client.search = None
+            q.client.sort = None
+            q.client.filters = None
+            q.client.page_offset = 0
+            q.client.total_rows = len(q.app.h2o_df)
+
+        # Update the rows on in our UI
```
mturoci, line 141:
```suggestion
        # Update the rows in our UI
```

py/examples/table_pagination_h2o3.py
```diff
@@ -0,0 +1,225 @@
...
+    logger.info(
+        f"Creating new table for rows: {q.client.page_offset} to {q.client.page_offset + q.app.rows_per_page}"
+    )
+
+    working_frame = prepare_h2o_data(q)
+
+    # Bring our limited UI rows locally to pandas to prepare for our ui.table
+    local_df = working_frame[
+        q.client.page_offset : q.client.page_offset + q.app.rows_per_page, :
+    ].as_data_frame()
```
mturoci, line 159: Why is conversion to pandas needed? Isn't it possible to iterate over `working_frame` and map it to `ui.table_row` directly?

mturoci (commented):
Looking good. A few comments from me.

* It would be good to add a link to h2o3 docs for newcomers.
* This example will not be runnable out of the box as it depends on running h2o3. We should make it clear via a comment at the beginning of the file.

mturoci:
Merging this so that I could merge the server-side table PR. We can deal with minor improvements later as needed.

py/examples/table_pagination_h2o3.py
```diff
@@ -0,0 +1,225 @@
...
+            ]
+
+    if q.client.search is not None:
+        # We check if our search term is in any of the searchable columns
+        # Start with and index of 0s and then filter to only keep rows with index > 0
+
+        index = h2o.create_frame(
+            rows=len(working_frame), cols=1, integer_fraction=1, integer_range=1
+        )
+        index["C1"] = 0
```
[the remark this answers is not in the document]
mtanco: This basically creates an empty array of 0's that is the size of the H2O Frame and then for each "searchable" column we update the array with + 1 if the item is there. So that the search result isn't only valid it it's in **all** columns but just at least one. 

So at the end each index of the frame with have 0 or >0 and we return rows >0.

There may be a more effecient way to do this! but this is what I could come up with

PR #1320, opened 2022-04-01
fix: Allow sorting of grouped by rows

### LINKED ISSUE #1287: Grouped by rows are not sortable
#### Steps To Reproduce

1. Run `examples.table`
2. Group by Issue
3. Sort by Issue
4. Try to expand groups - you should see no data rows there.

It would be good to also think about what's the best UX when sorting rows during group by. There are 2 options:

* Keep groups in the order they are displayed and sort rows only within the context of a particular group.
* Sort rows in the context of rows (whole table), thus groups will be rearranged on sort.

marek-mihok:
Fixes missing table rows in specific cases when sorting grouped list. Sorting of grouped list does not sort groups, but only the rows inside the groups.

Closes #1287

### DIFF
--- a/ui/src/table.test.tsx
+++ b/ui/src/table.test.tsx
@@ -644,6 +644,20 @@ describe('Table.tsx', () => {
       expect(getAllByRole('gridcell')[3].textContent).toBe(cell21)
     })
 
+    it('Checks if groups contain rows after sort is applied', () => {
+      const { container, getAllByText, getByTestId, getAllByRole } = render(<XTable model={tableProps} />)
+
+      fireEvent.click(getByTestId('groupby'))
+      fireEvent.click(getAllByText('Col1')[1]!)
+      fireEvent.click(container.querySelector('.ms-GroupHeader-expand')!)
+
+      expect(getAllByRole('gridcell')[3].textContent).toBe(cell11)
+
+      fireEvent.click(container.querySelector('.ms-DetailsHeader-cellTitle i[class*=sortingIcon]')!)
+
+      expect(getAllByRole('gridcell')[3].textContent).toBe(cell11)
+    })
+
     it('Searches grouped list', () => {
       const { container, getAllByText, getByTestId, getAllByRole } = render(<XTable model={tableProps} />)
 
--- a/ui/src/table.test.tsx
+++ b/ui/src/table.test.tsx
@@ -630,32 +630,27 @@ describe('Table.tsx', () => {
       expect(groupHeaders[2]).toHaveTextContent('2012-09-14T18:26:01(1)')
     })
 
-    it('Checks if groups contain rows after sort is applied', () => {
-      const { container, getAllByText, getByTestId, getAllByRole } = render(<XTable model={tableProps} />)
-
-      fireEvent.click(getByTestId('groupby'))
-      fireEvent.click(getAllByText('Col1')[1]!)
-      fireEvent.click(container.querySelector('.ms-DetailsHeader-collapseButton')!)
-
-      expect(getAllByRole('gridcell')[3].textContent).toBe(cell21)
-
-      fireEvent.click(container.querySelector('.ms-DetailsHeader-cellTitle i[class*=sortingIcon]')!)
-
-      expect(getAllByRole('gridcell')[3].textContent).toBe(cell21)
-    })
-
     it('Sorts rows inside the group of the grouped list', () => {
+      // test with non-alphabetical groups order to check whether group alphabetical sorting on groupby is considered
+      tableProps = {
+        ...tableProps,
+        rows: [
+          { name: 'rowname1', cells: [cell11, 'Group2'] },
+          { name: 'rowname2', cells: [cell21, 'Group2'] },
+          { name: 'rowname3', cells: [cell31, 'Group1'] }
+        ]
+      }
       const { container, getAllByText, getByTestId, getAllByRole } = render(<XTable model={tableProps} />)
 
       fireEvent.click(getByTestId('groupby'))
       fireEvent.click(getAllByText('Col2')[1]!)
       fireEvent.click(container.querySelector('.ms-DetailsHeader-collapseButton')!)
 
-      expect(getAllByRole('gridcell')[3].textContent).toBe(cell11)
+      expect(getAllByRole('gridcell')[8].textContent).toBe(cell11)
 
       fireEvent.click(container.querySelector('.ms-DetailsHeader-cellTitle i[class*=sortingIcon]')!)
 
-      expect(getAllByRole('gridcell')[3].textContent).toBe(cell21)
+      expect(getAllByRole('gridcell')[8].textContent).toBe(cell21)
     })
 
     it('Searches grouped list', () => {
--- a/ui/src/table.test.tsx
+++ b/ui/src/table.test.tsx
@@ -630,7 +630,21 @@ describe('Table.tsx', () => {
       expect(groupHeaders[2]).toHaveTextContent('2012-09-14T18:26:01(1)')
     })
 
-    it('Sorts grouped list', () => {
+    it('Checks if groups contain rows after sort is applied', () => {
+      const { container, getAllByText, getByTestId, getAllByRole } = render(<XTable model={tableProps} />)
+
+      fireEvent.click(getByTestId('groupby'))
+      fireEvent.click(getAllByText('Col1')[1]!)
+      fireEvent.click(container.querySelector('.ms-DetailsHeader-collapseButton')!)
+
+      expect(getAllByRole('gridcell')[3].textContent).toBe(cell21)
+
+      fireEvent.click(container.querySelector('.ms-DetailsHeader-cellTitle i[class*=sortingIcon]')!)
+
+      expect(getAllByRole('gridcell')[3].textContent).toBe(cell21)
+    })
+
+    it('Sorts rows inside the group of the grouped list', () => {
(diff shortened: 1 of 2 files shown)


ui/src/table.test.tsx
```diff
@@ -630,7 +630,21 @@ describe('Table.tsx', () => {
...
+
+      fireEvent.click(getByTestId('groupby'))
+      fireEvent.click(getAllByText('Col1')[1]!)
+      fireEvent.click(container.querySelector('.ms-DetailsHeader-collapseButton')!)
+
+      expect(getAllByRole('gridcell')[3].textContent).toBe(cell21)
+
+      fireEvent.click(container.querySelector('.ms-DetailsHeader-cellTitle i[class*=sortingIcon]')!)
+
+      expect(getAllByRole('gridcell')[3].textContent).toBe(cell21)
```
mturoci, line 644: What's the point of this test? It checks if `cell21` is present at a particular place and then it checks it didn't change after sorting? I would assume we want to test whether the sort actually changed row position.

ui/src/table.tsx
```diff
@@ -735,7 +735,7 @@ export const
 
         setGroups(groups => {
           if (groups) {
-            setFilteredItems(filteredItems => groups?.reduce((acc, group) =>
+            setFilteredItems(filteredItems => [...groups].sort((group1, group2) => group1.startIndex - group2.startIndex).reduce((acc, group) =>
```
mturoci, line 738: Why does this work?

mturoci (commented):
Thanks @marek-mihok! The code seems to work, but I need some more clarification as to why :)

ui/src/table.tsx
marek-mihok: The original implementation supposes that value of `groups` is similar to this one:
```ts
const groups = [
  { 
  key: "groupC", 
  name: "groupC", 
  startIndex: 0, 
  count: 3, 
  isCollapsed: false
  },
  { 
  key: "groupA", 
  name: "groupA", 
  startIndex: 3, 
  count: 2, 
  isCollapsed: false
  },
  { 
  key: "groupB", 
  name: "groupB", 
  startIndex: 5, 
  count: 4, 
  isCollapsed: false
  }
]
```
However, this is not true since `groups` are sorted alphabetically [on line 562](https://github.com/h2oai/wave/blob/56a9ce8f2ef3a48bc3a956aa178e16de1aaf3acc/ui/src/table.tsx#L562) in `makeGroups`. Therefore the real value of `groups` is in our case this:
```ts
const groups = [
  { 
  key: "groupA", 
  name: "groupA", 
  startIndex: 3, 
  count: 2, 
  isCollapsed: false
  },
  { 
  key: "groupB", 
  name: "groupB", 
  startIndex: 5, 
  count: 4, 
  isCollapsed: false
  },
  { 
  key: "groupC", 
  name: "groupC", 
  startIndex: 0, 
  count: 3, 
  isCollapsed: false
  }
]
```

There is no problem with viewing groups, Fluent takes it whether the value is as in the first example or the second one, but when it comes to sorting, here lies a problem.
Let's have a look at our original group sorting implementation:
```ts
if (groups) {
 setFilteredItems(filteredItems => groups.reduce((acc, group) =>
 [...acc, ...filteredItems.slice(group.startIndex, acc.length + group.count).sort(sortingF(column, sortAsc))],
 [] as any[]) || [])
}
```
We can start with the first iteration of our reducer function:

1. it takes the first group from our `groups` const
```
  { 
  key: "groupA", 
  name: "groupA", 
  startIndex: 3, 
  count: 2, 
  isCollapsed: false
  }
```
2. It slices our `filteredItems` from index 3 to 5
3. It sorts the items
4. And here is the main problem - it pushes it **to the beginning** of the reducer's accumulator (not back to index 3-5!!)

When all iterations are finished, `filteredItems` where subsets are not just sorted, but also rearranged, are set with `setFilteredItems` and the same `groups` are applied to it. Now our `groups` do not match updated `filteredItems`.

The most simple solution here is to sort the `groups` by its `startIndex` before reducing `filteredItems` without modifying the original `groups` value:
```ts
if (groups) {
 setFilteredItems(filteredItems => [...groups].sort((group1, group2) => group1.startIndex 
 group2.startIndex).reduce((acc, group) =>
 [...acc, ...filteredItems.slice(group.startIndex, acc.length + group.count).sort(sortingF(column, sortAsc))],
 [] as any[]) || [])
}
```

Another solution would be to push the sorted items in reducer to the right accumulator's indexes, but this solution would be more costy and complicated.

The reason why the original implementation was working in most of the cases is that usually we have just 2 groups created after groupby in our .py examples and there was a high chance these groups were in correct order.

ui/src/table.test.tsx
marek-mihok: The point was to check the correct viewing of rows in case we have as much groups created as possible using groupby. So to group it by column containing unique value in each row. This was the most common case where our group sorting did not work. As far as there is just one value in each group in this case, I can expect it will stay the same after applying sort. Before the group sorting fix there was no rows shown in the group and this test was failing.

mturoci: It would be better to test sort on > 1 row and check if the sort actually worked.

ui/src/table.tsx
mturoci: Makes sense, please add a brief comment with explanation for future reference.

ui/src/table.test.tsx
marek-mihok: Test case whether sort works or not is already implemented: https://github.com/h2oai/wave/blob/56a9ce8f2ef3a48bc3a956aa178e16de1aaf3acc/ui/src/table.test.tsx#L647 

However this test was working even before the sorting issue fix.
Is this what you mean?

mturoci: Yes, I just want a test that makes sure grouped by rows are correctly sorted after sorting.

>However this test was working even before the sorting issue fix.

How is that possible?

marek-mihok: Let's make following test case mentioned above:
```ts
it('Sorts rows inside the group of the grouped list', () => {
   const { container, getAllByText, getByTestId, getAllByRole } = render(<XTable model={tableProps} />)

   fireEvent.click(getByTestId('groupby'))
   fireEvent.click(getAllByText('Col2')[1]!)
   fireEvent.click(container.querySelector('.ms-DetailsHeader-collapseButton')!)

   expect(getAllByRole('gridcell')[3].textContent).toBe(cell11)

   fireEvent.click(container.querySelector('.ms-DetailsHeader-cellTitle i[class*=sortingIcon]')!)

   expect(getAllByRole('gridcell')[3].textContent).toBe(cell21)
})
```

on following data:
```ts
rows: [
   { name: 'rowname1', cells: ['Quick brown fox.', 'Group1'] },
   { name: 'rowname2', cells: ['Jumps over a dog.', 'Group1'] },
   { name: 'rowname3', cells: ['Wooo hooo.', 'Group2'] }
]
```

We are grouping by `Col2` so now we have two groups `Group1` and `Group2` and its alphabetical order is the same as default one. In this case sorting works as expected and this test case does not reflect our issue.

I can replace name Group1 with Group2 and vice-versa for this test case to reflect also our issue and remove the test case we are commenting.

marek-mihok: I've edited the existing test case to reflect our issue and now it passes only after the issue is fixed:
```ts
    it('Sorts rows inside the group of the grouped list', () => {
      // test with non-alphabetical groups order to check whether group alphabetical sorting on groupby is considered
      tableProps = {
        ...tableProps,
        rows: [
          { name: 'rowname1', cells: [cell11, 'Group2'] },
          { name: 'rowname2', cells: [cell21, 'Group2'] },
          { name: 'rowname3', cells: [cell31, 'Group1'] }
        ]
      }
      const { container, getAllByText, getByTestId, getAllByRole } = render(<XTable model={tableProps} />)

      fireEvent.click(getByTestId('groupby'))
      fireEvent.click(getAllByText('Col2')[1]!)
      fireEvent.click(container.querySelector('.ms-DetailsHeader-collapseButton')!)

      expect(getAllByRole('gridcell')[8].textContent).toBe(cell11)

      fireEvent.click(container.querySelector('.ms-DetailsHeader-cellTitle i[class*=sortingIcon]')!)

      expect(getAllByRole('gridcell')[8].textContent).toBe(cell21)
    })
```

If you are curious why `getAllByRole('gridcell')[8]`, here is the return value of `getAllByRole('gridcell').map(item => item.innerHTML)` from debug console after groupby and before sort:
```
0:'<button data-is-focusable="false" type="button" class="ms-GroupHeader-expand expand-269" aria-expanded="true" aria-label="expand collapse group"><i data-icon-name="ChevronRightMed" aria-hidden="true" class="expandIsCollapsed-278"></i></button>'
1:'<span>Group1</span><span class="headerCount-267">(1)</span>'
2:''
3:'<button type="button" class="ms-Link root-169" tabindex="-1">Wooo hooo.</button>'
4:'Group1'
5:'<button data-is-focusable="false" type="button" class="ms-GroupHeader-expand expand-269" aria-expanded="true" aria-label="expand collapse group"><i data-icon-name="ChevronRightMed" aria-hidden="true" class="expandIsCollapsed-278"></i></button>'
6:'<span>Group2</span><span class="headerCount-267">(2)</span>'
7:''
8:'<button type="button" class="ms-Link root-169" tabindex="-1">Quick brown fox.</button>'
9:'Group2'
10:''
11:'<button type="button" class="ms-Link root-169" tabindex="-1">Jumps over a dog.</button>'
12:'Group2'
```

mturoci (approved):
Thanks @marek-mihok!

Issue #428, opened 2020-12-21
Allow clickable links in table rows


mturoci:
via @vopani 

#### Is your feature request related to a problem? Please describe

It helps improving a base UX to render urls directly as links thus making them clickable as opposed to plaintext that user has to manually copy and paste to browser url bar.

#### Describe the solution you'd like

We have 2 options here:
* Auto detect if the plaintext is a link and render anchor tag correspondingly.
* Provide a `data_type: 'link'` and render accordingly.
* Add additional `TableCellType` - `LinkTableCellType`

#### Additional context
We already have a `link` attribute on `TableColumnType` that specifies the primary clickable column and might cause confusion so we need to be careful about naming.

cc @lo5 @geomodular for discussion

mturoci:
As suggested by @mneuhaus in our [Discourse](https://h2owave.h2o.ai/t/external-link-in-table-cell/81), one possible solution would be to allow markdown as table cell content (instead of plain string). This would also add additional text formatting flexibility if needed.

sergiomora03:
But if we use that, we couldn't use `ui.progress_table_cell_type()`, right?

mturoci:
Why not? This proposal would only enhance existing string column capabilities with markdown. If you specify `cell_type` as `ui.progress_table_cell_type()` for example, you get the desired progress instead of string like it is now.

lo5:
One aspect we need to consider before doing this is to make our markdown-to-HTML conversion a bit smarter. It currently wraps all text in `<p></p>` which would introduce unwanted whitespace inside table cells, so it would be good to handle some special cases at the markdown level (e.g. unwrap single-lines, etc.) instead of attempting a post-fix in the table's CSS.

mturoci:
On second thought, enabling link support via our existing `ui.link` probably makes more sense as markdown links are quite limited and only support opening links in current tab which would not satisfy #567. Using `ui.link` would also allow downloads which may be a common scenario in a lot of apps.

UK-17:
Any update on this ? I was making a JIRA issue tracker table using ui.table
It would be really helpful if links are allowed and i can redirect each row to its jira page.

mturoci:
We will try to include it within the next release.

mturoci:
Related https://github.com/h2oai/wave/issues/1379

Issue #567, opened 2021-02-13
Allow For Links In Tables That Open New Windows/Tabs


TravisCouture:
#### Is your feature request related to a problem? Please describe

When a row in a table is clicked leading to a trigger the only way to link to an external page is if the name for the row is set as the link.

#### Describe the solution you'd like

It would be pretty useful to be able to embed links within tables so that when the cell is clicked, a new tab or window is opened.

#### Describe alternatives you've considered

Setting the row name as the link address but this is pretty limiting and prevents more than one link per row.

mturoci:
Thanks for the issue @TravisCouture. Would #428 resolve the issue for you? I am not sure I really understood what you are asking in connection with `trigger` though.

PR #1321, opened 2022-04-01
feat: Add clickable links for Table

aalencar:
Proposed API

Set `data_type` to `'link'` and use Markdown notation `[label](url)`

```py
 ui.table(
        name='table',
        columns=[
            ui.table_column(name='link', label='Link', data_type='link')
        ],
        rows=[
            ui.table_row(name='row1', cells=['[Wave](https://wave.h2o.ai/)'])
        ]
    )
```

I tried to use `ui.link` [as suggested](https://github.com/h2oai/wave/issues/428#issuecomment-924927039) by @mturoci but I didn't find a way to pass `label` to it because `cells` only accept strings. I wish I could do something like `cells=[ui.link(label='my label', path='https://website.com')]`. 

```ts
/** Create a table row. */
interface TableRow {
  /** An identifying name for this row. */
  name: Id
  /** The cells in this row (displayed left to right). */
  cells: S[]
}
```


mturoci:
It would be good to also think about supporting links that will be able to also open external pages in new tabs (https://github.com/h2oai/wave/issues/567) and allow download as well. Also use TableCellType instead of data_type as this is a specific type of cell, not a type of data, it's a string after all.

mturoci:
This is a good direction. Let's do the following changes:

* URLTableCellType -> LinkTableCellType
* new_tab -> target - for consistency with `ui.link`

>Should it default to always open in a new tab?

No, the default should be the same as `ui.link`.

>Should I remove the current behavior if the user doesn't provide a URL with http(s) prefix?

Can you provide a link? I was not able to find the string you quoted in `table.md`

>I thought that maybe it could be used for routing to different apps in the same domain, so the user wouldn't have to pass the full domain.

It should work like a common web routing (relative links). I see your point, but that would prevent people who want to just append to the current URL (for some reason). Specifying a full domain for routing to a different app on the same server is fine.

website/widgets/form/table.md
```diff
@@ -355,17 +355,20 @@ Check the [complete example](/docs/examples/table-pagination) with all the suppo
 
 ## With links
 
-You can have links using Markdown notation and setting `data_type` to `'link'`.
+Links can be displayed by setting `cell_type` to `'ui.url_table_cell_type'`. If `'new_tab'` is set to `'True'` the link will be opened on a new tab.
+To provide a label use the Markdown notation `[label](url)`.
+If the URL doesn't have '`http(s)://`' the link will be appended to the current URL, trying to access a page in the same domain.
```
aalencar, line 360: @mturoci 
> Can you provide a link? I was not able to find the string you quoted in table.md

Current behavior
1. example.com -> <current_domain>/example.com
2. http(s)://example.com -> example.com

Maybe we should remove the "trying to access a *page* in the same domain" part because it might confuse people with wave's concept of *page*.

ui/src/table.tsx
```diff
@@ -450,6 +454,10 @@ const
         wave.args[m.name] = [item.key as S]
         wave.push()
       },
+      getLinkComp = (str: S, target?: S) => {
+        const [, label, path = str] = /\[(.+)\].*\((.+)\)/.exec(str) || []
```
mturoci, line 458: We do not expect anything between `](` I guess. Would this work?
```suggestion
        const [, label, path = str] = /\[(.+)\]\((.+)\)/.exec(str) || []
```

ui/src/table.tsx
```diff
@@ -450,6 +454,10 @@ const
         wave.args[m.name] = [item.key as S]
         wave.push()
       },
+      getLinkComp = (str: S, target?: S) => {
```
mturoci, line 457:
```suggestion
      LinkTableCellType = (str: S, target?: S) => {
```

ui/src/table.tsx
```diff
@@ -471,6 +479,7 @@ const
         if (col.cellType?.progress) return <XProgressTableCellType model={col.cellType.progress} progress={item[col.key]} />
         if (col.cellType?.icon) return <XIconTableCellType model={col.cellType.icon} icon={item[col.key]} />
         if (col.cellType?.tag) return <XTagTableCellType model={col.cellType.tag} serializedTags={item[col.key]} />
+        if (col.cellType?.url) return getLinkComp(item[col.key], col.cellType.url.target)
```
mturoci, line 482: nit: Use JSX instead of function call.

ui/src/link_table_cell_type.tsx
```diff
@@ -0,0 +1,22 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { S } from 'h2o-wave'
+
+export interface LinkTableCellType {
```
mturoci, line 17: Make sure the interface and implementation for cell are colocated (for context). So either move this into `table.tsx` or move the cell implementation here. Whatever you prefer.

website/widgets/form/table.md
```diff
@@ -352,3 +352,24 @@ q.page['example'] = ui.form_card(box='1 1 3 6', items=[
 ```
 
 Check the [complete example](/docs/examples/table-pagination) with all the supported features like sort/filter/search/download/reset.
+
+## With links
+
+Links can be displayed by setting `cell_type` to `'ui.url_table_cell_type'`. If `'new_tab'` is set to `'True'` the link will be opened on a new tab.
```
mturoci, line 358: Do not forget to rename the `new_tab` attribute.

website/widgets/form/table.md
mturoci: The behavior is fine (that's how the web works). Changing it would be a breaking change which I am not sure is worth it. However, you are right that the wording can be improved to not interfere with the page concept in Wave - +1.

website/widgets/form/table.md
```diff
@@ -352,3 +352,24 @@ q.page['example'] = ui.form_card(box='1 1 3 6', items=[
...
+To provide a label use the Markdown notation `[label](url)`.
+If the URL doesn't have '`http(s)://`' the link will be appended to the current URL, trying to access a page in the same domain.
+
+```py
+q.page['example'] = ui.form_card(box='1 1 3 3', items=[
+    ui.table(
+        name='table',
+        columns=[
+            ui.table_column(name='link', label='Link', cell_type=ui.url_table_cell_type(name='urls', new_tab=True))
```
mturoci, line 367: rename the attr.

Tip: Good check is to run `tools/showcase - make generate`. It's going to throw an error for these cases since it's an invalid code. However, make sure your UI build is always up to date prior to running it.

ui/src/table.tsx
```diff
@@ -38,6 +40,8 @@ interface TableCellType {
   icon?: IconTableCellType
   /** Renders one or more tags. */
   tag?: TagTableCellType
+  /** Renders a link. Links can have the form of a normal URL or use Markdown notation to provide a label, e.g [label](url)*/
+  url?: LinkTableCellType
```
mturoci, line 44:
```suggestion
  link?: LinkTableCellType
```

mturoci (changes requested):
Thanks @aalencar! Looks good overall, need a few minor changes.

ui/src/table.tsx
aalencar: It was part of an old regex I used, not needed anymore. Removed.

ui/src/link_table_cell_type.tsx
aalencar: Moved

website/widgets/form/table.md
aalencar: Overlooked. Renamed.

ui/src/link_table_cell_type.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { XLink } from './link'
+
+export interface LinkTableCellType {
+  /** An identifying name for this component. */
+  name: S
+  /** Where to display the link. Setting this to an empty string or `'_blank'` opens the link in a new tab or window. */
+  target?: S
+}
+
+export const XLinkTableCellType = ({str, target}: { str: S, target?: S}) => {
```
mturoci, line 26: Nit: Let's reuse the `LinkTableCellType` interface here instead of a standalone `target` prop.

mturoci (commented):
Looks good, just one more nit. Will merge it after `the 0.21` release.

ui/src/link_table_cell_type.tsx
aalencar: I just pushed, not sure if that's what you meant.

mturoci: Yes, looking good.

lo5 (commented):
I'm not clear about why this feature is required. We already support the ability for a row to be linked. If there is a need to provide row-specific actions, there's already the other PR with the contextual menu. If the need is to support links on each row that open in new windows, then the request is too specific, and the implementation/API seem hacky (regex, markdown-subset, etc.)

Is there a link to the specific feature request or issue?

aalencar:
>  If there is a need to provide row-specific actions, there's already the other PR with the contextual menu

That PR implementation has every row with the same actions. We can discuss further improvements there.

> Is there a link to the specific feature request or issue?

I couldn't find any issue or discussion opened by @vopani. @mturoci by any chance, do you remember what was the usecase?

>  If the need is to support links on each row that open in new windows, then the request is too specific, and the implementation/API seem hacky (regex, markdown-subset, etc.)

@lo5  It seems hacky indeed. What do you suggest as an alternative?

lo5:
> What do you suggest as an alternative?

Hard to say, since I'm not sure about the origin of this requirement. 

One option is to have a `MarkdownCellType` which interprets cells as markdown. This opens up the possibility of using arbitrary formatted content (including links, e.g. `[this](/foo) or [that](bar)`). Implementation-wise, this can use the same markdown processor as `ui.text()`, with maybe some niceties like removing margins if `p:only-child`. 

The decision on whether to open links in `_blank` can be based on whether the `<a href=` is an absolute (open in `_blank`) or relative URL (open in same window).

I'm not sure if the table already supports markdown formatting.

mturoci:
The original issue is #428 (found via the branch name) and the requirement is to be able to open the links right from the table. I am fine with MarkdownCell, but make sure the spacing is correct and does not interfere with the rest of the table.

aalencar:
> The original issue is https://github.com/h2oai/wave/issues/428 (found via the branch name)

I meant where this request came from. If I see "via <user>" I assume it comes from somewhere else, GH discussion, another GH issue, or a private conversation. #428 doesn't describe a **use case**. Like @lo5 said, maybe we already support the use case since it's an old feature request.

> I am fine with MarkdownCell, but make sure the spacing is correct and does not interfere with the rest of the table.

Ok

lo5:
> Markdown doesn't support setting the target to open in a new tab. 

Convert markdown to html, insert into DOM, use `querySelectorAll()` on the `ref` to get all `a` elements, inspect the paths, add an `onClick` handler to the ones that need to open in `_blank`.

aalencar:
> Convert markdown to html, insert into DOM, use querySelectorAll() on the ref to get all a elements, inspect the paths, add an onClick handler to the ones that need to open in _blank.

This could also work but I wonder if there is any benefit over using regex and `XLink` component if the provided string is a link.

> I'd prefer not to encourage using HTML.

Ok.

> Not sure what you mean by this. The markdown used in a table cell should be no different from that used elsewhere.

Open in a new tab when `http(s)://` is present is not part of the Markdown specification, so my question was whether we need to document only this behavior and also point the user to some online reference in case they are not familiar with Markdown in general.

lo5:
> This could also work but I wonder if there is any benefit over using regex and XLink component if the provided string is a link.

The main benefit is that the technique I suggested would support arbitrary markdown, not just links.

> my question was whether we need to document only this behavior

Add an attribute to `MarkdownCellType` that specifies whether or not external links should be opened in `_blank`. Default to opening all links in the same window.

lo5:
> Could you please clarify this point? Don't we want to only extend the markdown link feature to support opening in a new tab? Maybe I'm missing something.

The goal is to support arbitrary links in table cells. Links can be opened in the same tab or in a new tab. Agreed?

You're suggesting using regex and `XLink`. I'm saying that's hacky and error-prone.

The way I see it, there's not much of a difference between supporting hyperlinks, and supporting arbitrary Markdown inside cells. 

If we're *only* going to support hyperlinks, it raises more questions:
- What if there are 1+ hyperlinks in a cell? 
- What if the hyperlinks are accompanied by text?
- What if the text requires alternate formatting?

So I don't see a good reason to support *only* hyperlinks. Might as well support Markdown, since Markdown already supports hyperlinks. Hence, `MarkdownCellType`.

That leaves one question open: do we open links in the current tab or a new tab? To which I suggested that we leave that decision to the user by adding an attribute (flag) to `MarkdownCellType` that specifies whether or not external links should be opened in _blank. And if they are to be opened in _blank, I suggested using the `ref`/`querySelectorAll()` technique, since that is the most robust way to implement this.

Hope this makes it clear. If not, call me.

lo5:
> I  dynamically added the target attribute instead of a click handler like you suggested.

Good idea.

> Should we wrap MarkdownCellType by default? 

Yes

> Do you know any other case Markdown would break the cell?

There might be, but should be enough to handle the `p` tag. Blockquotes, tables, headings, etc. don't make sense inside table cells anyway.

py/examples/table_markdown_links.py
```diff
@@ -0,0 +1,27 @@
...
+        ui.text_xl(content='Table with Markdown links'),
+        ui.table(
+            name='table',
+            columns=[
+                ui.table_column(name='description', label='URL type'),
+                ui.table_column(name='markdown', label='Link', cell_type=ui.markdown_table_cell_type(target='_blank'))
+            ],
+            rows=[
+                ui.table_row(name='row1', cells=['Absolute', '[Wave website](http://wave.h2o.ai/)']),
+                ui.table_row(name='row2', cells=['Relative', '[Go to /wave](/wave)']),
```
mturoci, line 22:
```suggestion
                ui.table_row(name='row1', cells=['Absolute path', '[Wave website](http://wave.h2o.ai/)']),
                ui.table_row(name='row2', cells=['Relative path', '[Go to /wave](/wave)']),
```

mturoci (changes requested):

ui/src/markdown_table_cell_type.tsx
```diff
@@ -0,0 +1,42 @@
...
+  /** Where to display the link. Setting this to '_blank'` opens the link in a new tab or window. */
+  target?: S
+}
+
+export const XMarkdownTableCellType = ({ model: m }: { model: MarkdownTableCellType & { content: S}} ) => {
+  const ref = React.useRef<HTMLDivElement>(null)
+
+  React.useEffect(() => {
+    const anchors = document.querySelectorAll<HTMLAnchorElement>('a')
+    anchors?.forEach(a => {if (m.target) a.target = m.target})
```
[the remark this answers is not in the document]
aalencar: I added the ref but didn't use it, thanks!

mturoci (approved):
LGTM, good work @aalencar!

PR #1324, opened 2022-04-04
feat: Add command menu to table

aalencar:
### Example

```py
commands = [
    ui.command(name='details', label='Details'),
    ui.command(name='rename', label='Rename'),
    ui.command(name='delete', label='Delete'),
]

q.page['example'] = ui.form_card(box='1 1 3 3', items=[
    ui.table(
        name='table',
        columns=[ui.table_column(name='actions', label='Actions', cell_type=ui.command_table_cell_type(name='commands', items=commands))],
        rows=[ui.table_row(name='first_row', cells=[])]
    )
])
```

### API

```ts
/** Defines cell content to be rendered instead of a simple text. */
interface TableCellType {
  /** Renders a progress arc with a percentage value in the middle. */
  progress?: ProgressTableCellType
  /** Renders an icon. */
  icon?: IconTableCellType
  /** Renders a command menu. */
  command?: CommandTableCellType
}

/**
 * Create a cell type that renders command menu.
 *
 * Commands are typically displayed as context menu items. Useful when you need to provide
 * multiple actions within a single row.
 */
export interface CommandTableCellType {
  /** Items to render. */
  items: Command[]
  name?: S
}
```

### Submitted value
`q.args.<command_name> = <row_id>`


mturoci:
The current implementation only tells you that a command was clicked for some row, but doesn't tell you which one. Try to make it so that your example will allow deleting a particular row via a command menu, very similar to [our wizard example](https://github.com/h2oai/wave/blob/master/py/examples/wizard.py)

aalencar:
> The current implementation only tells you that a command was clicked for some row, but doesn't tell you which one.

The current implementation returns the ID of the row that was clicked.

> Try to make it so that your example will allow deleting a particular row via a command menu, very similar to [our wizard example](https://github.com/h2oai/wave/blob/master/py/examples/wizard.py)

OK

aalencar:
## New API

```ts
/** Defines cell content to be rendered instead of a simple text. */
interface TableCellType {
  /** Renders a progress arc with a percentage value in the middle. */
  progress?: ProgressTableCellType
  /** Renders an icon. */
  icon?: IconTableCellType
  /** Renders one or more tags. */
  tag?: TagTableCellType
  /** Renders a command menu. */
  command?: CommandTableCellType
}
```

## In Action

```py
commands = [
    ui.command(name='details', label='Details'),
    ui.command(name='delete', label='Delete'),
]

q.page['example'] = ui.form_card(box='1 1 4 -1', items=[
        ui.table(
            name='table',
            columns=[
                ui.table_column(name='name', label='Name'),
                ui.table_column(name='actions', label='Actions',
                                cell_type=ui.command_table_cell_type(name='commands', items=commands))],
            rows=[ui.table_row(name=str(row.id), cells=[str(row.name)]) for row in rows]
        )
 ])
```

py/examples/table_commands.py
```diff
@@ -0,0 +1,69 @@
+# Table / Commands
```
mturoci, line 1: This example has a few problems:

* Uses global variable `rows` - not a problem when used as read-only, but since it's mutated it results displaying the same amount of rows for each browser (try deleting a row and open a new browser tab, the row count is going to be same instead of initial 50)
* Nit: Unnecessary string casts.
* Unnecessary async functions - whole async/await is relatively complex for regular Wave devs so make sure to use it only when really necessary.
* Deleting a specific row recreates every other row as well so it's impossible to tell whether the delete actually worked compared to previous data. It would be better if only a single row was removed and the rest stayed the same.

Take this as a base (may have some problems that need fixing, but should be generally fine):

```py
# Table / Commands
# Allow group of commands with context menu for each row.
# #table #commands
# ---
from faker import Faker

fake = Faker()


class TableRow:
    _id = 0

    def __init__(self):
        TableRow._id += 1
        self.id = f'row_{TableRow._id}'
        self.name = f'{fake.first_name()} {fake.last_name()}'
        self.details = fake.sentence()


def show_table(q) -> None:
    q.page['example'] = ui.form_card(box='1 1 4 -1', items=[
        ui.table(
            name='table',
            columns=[
                ui.table_column(name='name', label='Name'),
                ui.table_column(
                    name='actions', label='Actions',
                    cell_type=ui.command_table_cell_type(name='commands', items=[
                        ui.command(name='details', label='Details'),
                        ui.command(name='delete', label='Delete'),
                    ])
                )
            ],
            rows=[ui.table_row(name=r.id, cells=[r.name]) for r in q.client.rows]
        )
    ])


@app('/demo')
async def serve(q: Q):
    if not q.app.initialized:
        q.app.rows = [TableRow() for _ in range(50)]
        q.app.initialized = True
    if not q.client.initialized:
        q.client.rows = q.app.rows
        show_table(q)
        q.client.initialized = True

    if q.args.delete:
        q.client.rows = [row for row in q.client.rows if row.id != q.args.delete]
        q.page['example'].items[0].table.rows = [ui.table_row(name=r.id, cells=[r.name]) for r in q.client.rows]
    if q.args.details:
        for row in q.client.rows:
            if row.id == q.args.details:
                q.page['example'] = ui.form_card(box='1 1 4 4', items=[
                    ui.text(name='details', content=row.details),
                    ui.button(name='back', label='Back')
                ])
                break
    if q.args.back:
        show_table(q)

    await q.page.save()
```

mturoci (changes requested):
Thanks @aalencar. Looks good overall, but the example is not there yet.

py/examples/table_commands.py
aalencar: > Unnecessary async functions - whole async/await is relatively complex for regular Wave devs so make sure to use it only when really necessary.

I based it on `todo.py` but I will keep that in mind.

aalencar: I couldn't spot any problem with your example, works just fine. Thanks.

website/widgets/form/table.md
```diff
@@ -352,3 +352,22 @@ q.page['example'] = ui.form_card(box='1 1 3 6', items=[
...
+commands = [
+    ui.command(name='details', label='Details'),
+    ui.command(name='delete', label='Delete'),
+]
+
+q.page['example'] = ui.form_card(box='1 1 3 3', items=[
+    ui.table(
+        name='table',
+        columns=[ui.table_column(name='actions', label='Actions', cell_type=ui.command_table_cell_type(name='commands', items=commands))],
+        rows=[ui.table_row(name='first_row', cells=[])]
```
mturoci, line 370: Wondering whether there shouldn't be a data placeholder. I assume this would break if the command cell wouldn't be the last one.
```suggestion
        rows=[ui.table_row(name='first_row', cells=[''])]
```

mturoci (commented):
One last question from me.

cc @lo5 for API review.

website/widgets/form/table.md
aalencar: Indeed it won't work without the placeholder. Docs updated.

aalencar:
I found a bug: the menu doesn't close when clicking outside of the table (`onDismiss`). For some reason `toggleMenu` is being called twice, changing the flag back to its original state. Passing `false` directly to `setIsMenuHidden` instead of a callback worked but I still have to understand why.

https://user-images.githubusercontent.com/15820796/163177942-73338720-7d31-4307-af10-c7cccce0aa00.mov

ui/src/command_table_cell_type.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Create a cell type that renders command menu.
+ *
+ * Commands are typically displayed as context menu items. Useful when you need to provide
+ * multiple actions within a single row.
+ */
+ export interface CommandTableCellType {
+  /** Items to render. */
+  items: Command[]
```
lo5, line 28:
```suggestion
  commands: Command[]
```

ui/src/table.tsx
```diff
@@ -38,6 +39,8 @@ interface TableCellType {
   icon?: IconTableCellType
   /** Renders one or more tags. */
   tag?: TagTableCellType
+  /** Renders a command menu. */
+  command?: CommandTableCellType
```
lo5, line 43:
```suggestion
  menu?: MenuTableCellType
```

ui/src/menu.tsx
```diff
@@ -38,9 +38,9 @@ export interface Menu {
   name?: Id
 }
 
-export const XMenu = ({ model }: { model: Menu }) => {
+export const XMenu = ({ model }: { model: Menu & {hideCaret?: B } }) => {
```
lo5, line 41: I'm not in favor of reusing this component for the table cell menu - the `& {hideCaret?: B }` breaks pattern, and the styling customizations done here may not be relevant to the table cell menu. It should be fairly simply to code up a menu in the cell_type tsx file.

lo5 (changes requested):
Thanks @aalencar - minor comments.

aalencar:
> I found a bug: the menu doesn't close when clicking outside of the table (`onDismiss`). For some reason `toggleMenu` is being called twice, changing the flag back to its original state. Passing `false` directly to `setIsMenuHidden` instead of a callback worked but I still have to understand why. Clicking on the 3 dots icon also doesn't close it.
> 
>  onDismiss.bug.mov

Solved the first issue. Clicking on the 3 dots again doesn't close the dialog again though.

py/examples/tour.conf
```diff
@@ -96,6 +96,7 @@ table_pagination_filter.py
 table_pagination_search.py
 table_pagination_download.py
 table_pagination_groups.py
+table_commands.py
```
mturoci, line 99: Change to `table_menu`

ui/src/menu_table_cell_type.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export const XMenuTableCellType = ({ model }: { model: MenuTableCellType & { rowId: string } }) => {
+  const
+    ref = React.useRef<HTMLDivElement>(null),
+    [isMenuHidden, setIsMenuHidden] = React.useState(true),
+    open = () => setIsMenuHidden(false),
+    close = () => setIsMenuHidden(true)
+
+  return (
+    <div data-test={model.name} className={clas(css.container, 'w-menu')} ref={ref} onClick={open}>
```
mturoci, line 54: No need for `w-menu` selector - it's only used in cards that have background color controllable - header/nav - so that we can select these components and invert their colors accordingly.

ui/src/menu_table_cell_type.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <Fluent.FontIcon className={css.icon} iconName="MoreVertical" />
+      <Fluent.ContextualMenu
+        items={toCommands(model.commands.map(i => ({...i, value: model.rowId })))}
+        target={ref}
+        hidden={isMenuHidden}
+        onDismiss={close}
+        isBeakVisible
+        directionalHint={Fluent.DirectionalHint.bottomRightEdge}
```
mturoci, line 62: bottomCenter seems to fit better

ui/src/menu_table_cell_type.tsx
```diff
@@ -0,0 +1,68 @@
...
+  /** Items to render. */
+  commands: Command[]
+ /** An identifying name for this component. */
+  name?: S
+ }
+
+const css = stylesheet({
+  container: {
+    display: 'inline-flex',
+    alignItems: 'center',
```
mturoci, line 38: Why does this need to be a flex container?
What does alignItems center?

mturoci (changes requested):

ui/src/menu_table_cell_type.tsx
aalencar: Removed `alignItems`. `inline-flex` make the div fit its content, changed to `inline`.

mturoci (approved):
LGTM.

> Clicking on the 3 dots again doesn't close the dialog though.

The problem is both close (since user clicked outside the menu) and open callbacks are fired.
This can be solved by changing `open` function to `toggle` and writing custom logic for when should `onDismiss` fire by implementing `calloutprops.preventDismissOnEvent` and testing click event target element => if menu icon, toggle else close.

Let's leave it as is for now, the UX is fine as is. We can change this once someone files a complaint.

PR #1330, opened 2022-04-07
feat: Display additional information on plot tooltips

### LINKED ISSUE #644: Allow displaying additional information on plot tooltips
- Can we add more information to show on hover?  See Picture below of hover only being the original info ( I don't see a specific mark or anything that looks like this, but I can see it being requested a lot, Plotly and others allow this)

image (2)

-The events of a Line plot return the whole line when you click on it, what the user probably wants is the left and right point of that line segment. A work around is to add 2 marks, a line AND a point. This solution is working today but the user is having a hard time not accidentally clicking on the line. Also it looks a little clunky

marek-mihok:
All plot data are shown in tooltip now, not only data user is plotting.

image

image

Closes #644

### DIFF
--- a/ui/src/plot.tsx
+++ b/ui/src/plot.tsx
@@ -1060,9 +1060,11 @@ export const
           space = spaceTypeOf(raw_data, marks),
           data = refactorData(raw_data, plot.marks),
           { Chart } = await import('@antv/g2'),
+          tooltipCfg = Object.keys(data[0]).join('*'),
           chart = plot.marks ? new Chart(makeChart(el, space, plot.marks, model.interactions || [])) : null
         currentPlot.current = plot
         if (chart) {
+          chart.line().position(tooltipCfg).tooltip(tooltipCfg)
           currentChart.current = chart
           chart.data(data)
           if (model.events) {
--- a/ui/src/plot.tsx
+++ b/ui/src/plot.tsx
@@ -1062,21 +1062,19 @@ export const
             showCrosshairs: true,
             crosshairs: { type: 'xy' },
             // XXX pass container element
-            customContent: (name, items) => {
+            customContent: (title, items) => {
               const container = document.createElement('div')
               container.className = 'g2-tooltip'
-              let listItem = ''
-              items.forEach(({ data, name, mappingData, color }) => {
-                Object.keys(data).forEach(item => {
-                  listItem += `<li class="g2-tooltip-list-item" data-index={index} style="margin-bottom:4px;display:flex;align-items: center;">
-                      <span style="background-color:${(item === name || data[item] === name) && (mappingData?.color || color)};" class="g2-tooltip-marker"></span>
-                      <span style="display:inline-flex;flex:1;justify-content:space-between">
-                      <span style="margin-right: 16px;">${item}:</span><span>${data[item] instanceof Date ? data[item].toISOString().split('T')[0] : data[item]}</span>
-                      </span>
-                    </li>`
-                })
-              })
-              container.innerHTML = listItem
+              container.innerHTML = items.map(({ data, name, mappingData, color }) =>
+                Object.keys(data).map(item =>
+                  `<li class="g2-tooltip-list-item" data-index={index} style="margin-bottom:4px;display:flex;align-items: center;">
+                    <span style="background-color:${(item === name || data[item] === name) && (mappingData?.color || color)};" class="g2-tooltip-marker"></span>
+                    <span style="display:inline-flex;flex:1;justify-content:space-between">
+                    <span style="margin-right: 16px;">${item}:</span><span>${data[item] instanceof Date ? data[item].toISOString().split('T')[0] : data[item]}</span>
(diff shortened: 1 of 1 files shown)


ui/src/plot.tsx
```diff
@@ -1060,9 +1064,41 @@ export const
...
           { Chart } = await import('@antv/g2'),
+          tooltipCfg = Object.keys(data[0]).join('*'),
           chart = plot.marks ? new Chart(makeChart(el, space, plot.marks, model.interactions || [])) : null
         currentPlot.current = plot
         if (chart) {
+          marks.forEach(({ type }) => {
+            switch (type) {
+              case 'interval':
+                setChartTooltip(chart.interval(), tooltipCfg)
+                break
```
mturoci, line 1075: Wondering if something like this would work:

```suggestion
            setChartTooltip(chart[type](), tooltipCfg)
```

Might need edgecase check for invalid type so that undefined() is not called.

mturoci (commented):
Thanks @marek-mihok. Looks good overall, just a single comment.

mturoci:
Oops, mismatched branches, please ignore my last commit.

ui/src/plot.tsx
marek-mihok: I was wondering if something like this exists in javascript. Now it can be written in one line like this: 
`if (type) chart[type]().position(tooltipCfg).tooltip(tooltipCfg)` 
And it works! Thanks.

mtanco:
Before merging / closing, what does this look like with multi - groups? do we still get the color in the tool tip?

marek-mihok:
Yes @mtanco, colors are still shown in tooltips:

image

image

image

While browsing through G2 docs, I've found out that shared tooltip is a common scenario. @mturoci @lo5 what we can also think about is whether to provide this functionality as well. We can add an optional `tooltip_shared` prop on mark to provide shared tooltips like this one:
image

What do you think?

mturoci:
>I've found out that shared tooltip is a common scenario

Sounds good to me, but leaving the decision to @lo5 as a DataViz expert.

mturoci:
> I've found out that shared tooltip is a common scenario

@marek-mihok let's move this into a separate issue and merge this one

marek-mihok:
> let's move this into a separate issue and merge this one

@mturoci Sure it's for other issue than this one, I've just used this related channel for your opinion. I will create a separate issue for that.

ui/src/plot.tsx
```diff
@@ -1063,6 +1058,26 @@ export const
...
+                Object.keys(data).map(item =>
+                  `<li class="g2-tooltip-list-item" data-index={index} style="margin-bottom:4px;display:flex;align-items: center;">
+                    <span style="background-color:${mappingData?.color || color};" class="g2-tooltip-marker"></span>
+                    <span style="display:inline-flex;flex:1;justify-content:space-between">
+                    <span style="margin-right: 16px;">${item}:</span><span>${data[item] instanceof Date ? data[item].toISOString().split('T')[0] : data[item]}</span>
+                    </span>
+                  </li>`
+                ).join('')
+              ).join('')
+              return container
```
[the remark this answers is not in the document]
mturoci: >but you cannot depend on its return value.

I see, in that case, return `container` with rendered React within.

mturoci (changes requested):
I added some perf improvements so that tooltip animations are not laggy (15fps => 60fps). @marek-mihok please add the proper types instead of `any` and we are good to go.

mturoci (approved):

Nayananga:
Hi @mturoci, could you please point me to a location where I can find documentation for this feature?

marek-mihok:
@Nayananga, thanks for asking. As I am the one who implemented this feature, I will try to give you the answer. 

This PR introduced just a change of what is viewed in a tooltip content. Before, tooltips were showing just values user is plotting (values visible on plot), but now it shows all of the values. 

Let's say we have the following example:
```py
page.add('example', ui.plot_card(
    box='1 1 4 5',
    title='Line',
    data=data('year value increase', 8, rows=[
        ('1991', 3, 0),
        ('1992', 4, 1),
        ('1993', 3.5, -0.5),
        ('1994', 5, 1.5),
        ('1995', 4.9, -0.1),
        ('1996', 6, 1.1),
        ('1997', 7, 1),
        ('1998', 9, 2),
        ('1999', 13, 4),
    ]),
    plot=ui.plot([ui.mark(type='line', x_scale='time', x='=year', y='=value', y_min=0)])
))
```

**How tooltip was shown before:**
image


**How tooltip is shown after:**
image

This is not the API change, therefore it is not included in our docs.

@mturoci do you think we should describe this behaviour also in our documentation?

mturoci:
> do you think we should describe this behaviour also in our documentation?

I think it would be good to mention it. You will need to include a custom-made screenshot (same as in `dialog.md`) with `py ignore` annotation. However, this will probably need some extra adjustments in `showcase.py` as well.

PR #1332, opened 2022-04-07
feat: Allow plot marks to be non-interactive.

### LINKED ISSUE #1331: Modify return value for line plot event
This feature request is created based on description of [Issue 644](https://github.com/h2oai/wave/issues/644) reported by @fousasg.  

> The events of a Line plot return the whole line when you click on it, what the user probably wants is the left and right point of that line segment. A work around is to add 2 marks, a line AND a point. This solution is working today but the user is having a hard time not accidentally clicking on the line. Also it looks a little clunky

So what we want is to have return value for line plot event to be just left and right point of the clicked line segment instead of the whole line.

marek-mihok:
This PR modifies the return value for line plot event to be left and right point of the clicked line segment instead of the whole line.

image

Closes #1331


ui/src/plot.tsx
```diff
@@ -1070,11 +1070,16 @@ export const
...
-                    }
+                  chart.on('element:click', (ev: any) => {
+                    const
+                      { x, data: eventData } = ev,
+                      { points, data, shape } = eventData,
+                      selectedElements = chart.getElementsBy(el => el.getStates().includes('selected')),
+                      dataInterval = shape === 'line' ? data.filter((item: any, idx: number) => {
+                        if (idx !== points.length - 1 && x >= points[idx].x && x <= points[idx + 1].x) return true
+                        if (idx !== 0 && x <= points[idx].x && x >= points[idx - 1].x) return true
+                      }) : data
```
mturoci, line 1081:
```suggestion
                      dataInterval = shape === 'line' ? data.filter((item: any, idx: U) =>
                        (idx !== points.length - 1 && x >= points[idx].x && x <= points[idx + 1].x)
                        || (idx !== 0 && x <= points[idx].x && x >= points[idx - 1].x)
                      ) : data
```

mturoci (changes requested):
Thanks @marek-mihok.

Did you make sure this does not break the rest of existing plot types?

cc @lo5 since this is a breaking change

lo5:
The new behavior is odd. Line segments might make sense from a graphics perspective, but not from a data visualization perspective. I can't think of any common scenarios where the app author might need to deal with line segments. 

Clicking on a line has a very specific intent - selecting the *series*. e.g. On a multi-series line plot of, say, revenue over time by product, where each line represents a product, clicking on a line would mean "the user is interested in the product", which could be followed up with a, say, drill-down operation or such.

This behavior obviously would make no sense if there is a single line plot, which is why the plots support adding additional layers of marks for greater control over interactivity.

The problem here, as pointed out by @fousasg, is that if the plot only has a single line, clicking on a line doesn't make the user's intent clear. 

So the correct fix to this problem could be: introduce an `interactive: boolean` attribute on each mark, defaulting to `true`. If `interactive === false`, don't raise events for that mark layer. This way, the multi-series line plots are not affected, and if the author desires a single-line plot with interactivity, using a point layer on top of a line layer with the line layer set to `interactive=false` will solve the inadvertent-clicking problem.

What do you think?

website/widgets/plots/overview.md
```diff
@@ -105,6 +105,28 @@ q.page['example'] = ui.plot_card(
 )
 ```
 
+### Disable events
+
+If you do not wish to raise wave events on interaction with specific mark, simply specify it `interactive=False` attribute.
```
mturoci, line 110:
```suggestion
If you do not wish to raise wave events on interaction with a specific mark, simply specify the `interactive=False` attribute.
```

website/widgets/plots/overview.md
```diff
@@ -105,6 +105,28 @@ q.page['example'] = ui.plot_card(
...
+    title='Interval, range',
+    data=data('profession salary', 5, rows=[
+        ('medicine', 33000),
+        ('fire fighting', 18000),
+        ('pedagogy', 24000),
+        ('psychology', 22500),
+        ('computer science', 36000),
+    ]),
+    events=['select_marks'],
+    plot=ui.plot([ui.mark(type='interval', x='=salary', y='=profession', y_min=0, interactive=False)])
```
mturoci, line 126: Would be better to include the solution to a problem described in #1331.

mturoci (changes requested):
Thanks @marek-mihok, looks good overall, just a few docs nitpicks.

mturoci (approved):
Good job @marek-mihok!

PR #1340, opened 2022-04-11
feat: Build apps for managed cloud #1335

### LINKED ISSUE #1335: GitHub Actions for Tagged Releases of Example Apps
We turn the example apps that come with Wave (Dashboards, Tour, Theme Generator) into App Bundles that can be deployed in an H2O AI Cloud environment. This is a manual process today that happens for each Wave release and currently we're making the new bundle but it's not a specific tagged release or in S3. This can make getting these apps into the platform slow and cumbersome. 

When there is a new version of H2O Wave, we would like the tagged release to also contain `.wave` bundles for
* A Tour of Wave
* Dashboards
* Theme Generator

Continue the manual process

#### Additional context
Today, we have repos with Makefiles to make the Tour and Dashboard apps - maybe we continue to use those or maybe they can be retired. 
https://github.com/h2oai/wave-dashboards
https://github.com/h2oai/wave-tour

mturoci:
Needs metadata / screenshots / icon for theme generator app (see [this comment](https://github.com/h2oai/wave/issues/1335#issuecomment-1094694477)).

~~Waiting for @ozahavi to resolve S3 creds for the HAC bucket.~~ Done

Closes #1335


mturoci:
Everything is ready, the only thing I am thinking about is that both Tour and Theme generator apps do not listen on `/`. AFAIK HAC does not support opening instance URLs right away on different paths than `/` which may worsen UX. I should probably change it to /as part of the build process. Wdyt @mtanco?

mtanco:
Yes, changing the list path to root would be great!

PR #1345, opened 2022-04-12
feat: Display sorting icons only if sort is active

### LINKED ISSUE #1314: Display sorting icons only if sort is active
via @mtanco 

Our current Wave table tries to explicitly show users which columns are sortable by displaying an arrow icon. However, this may conflict with assumption that it represents the sort direction the data is sorted by default (which is not true).

According to both [aggrid](https://www.ag-grid.com/example.php) and [material design](https://material.angular.io/components/table/overview#sorting) it seems like the general accepted UX is to simply click the col, assume it's sortable and display sorting arrow only after the sort is performed in order to depict currently applied sort.

marek-mihok:
https://user-images.githubusercontent.com/23740173/162934246-aa0da8d0-6f7b-4231-9a24-cff6485eca45.mov

Closes #1314

### DIFF
--- a/ui/src/table.test.tsx
+++ b/ui/src/table.test.tsx
@@ -211,8 +211,8 @@ describe('Table.tsx', () => {
     it('Does not render sort arrow on sortable columns by default', () => {
       const { container } = render(<XTable model={tableProps} />)
 
-      // the sortingIcon class is present even if sort icon is not rendered, therefore we check for its child presence
-      expect(container.querySelector('i[class*=sortingIcon]')!.firstChild).not.toBeTruthy()
+      expect(container.querySelector("[data-icon-name='SortUp']")!).not.toBeInTheDocument()
+      expect(container.querySelector("[data-icon-name='SortDown']")!).not.toBeInTheDocument()
     })
 
     it('Renders sort arrow after clicking on sortable column header', () => {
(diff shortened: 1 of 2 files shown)


ui/src/table.test.tsx
```diff
@@ -208,13 +208,37 @@ describe('Table.tsx', () => {
...
+      fireEvent.click(container.querySelectorAll('.ms-DetailsHeader-cellTitle')[1])
+      expect(container.querySelector('i[class*=sortingIcon]')!.firstChild).not.toBeTruthy()
+    })
+
     it('Sorts by column - string', () => {
       const { container, getAllByRole } = render(<XTable model={tableProps} />)
 
       expect(getAllByRole('gridcell')[0].textContent).toBe(cell11)
-      fireEvent.click(container.querySelector('i[class*=sortingIcon]')!)
+      fireEvent.click(container.querySelectorAll('.ms-DetailsHeader-cellTitle')[0])
```
mturoci, line 239: This should work as well.

```suggestion
      fireEvent.click(container.querySelector('.ms-DetailsHeader-cellTitle'))
```

ui/src/table.test.tsx
```diff
@@ -208,13 +208,37 @@ describe('Table.tsx', () => {
   })
 
   describe('sort', () => {
+    it('Does not render sort arrow on sortable columns by default', () => {
+      const { container } = render(<XTable model={tableProps} />)
+
+      // the sortingIcon class is present even if sort icon is not rendered, therefore we check for its child presence
+      expect(container.querySelector('i[class*=sortingIcon]')!.firstChild).not.toBeTruthy()
```
mturoci, line 215: Can we target the firstChild directly? Role/class would be the best, but if not we can also go with `'i[class*=sortingIcon]:first-child'`

```suggestion
      expect(container.querySelector('i[class*=sortingIcon]')!.firstChild).not.toBeInTheDocument()
```

ui/src/table.test.tsx
```diff
@@ -208,13 +208,37 @@ describe('Table.tsx', () => {
   })
 
   describe('sort', () => {
```
mturoci, line 210: Add a test that checks if reset works correctly - hides sort icon.

mturoci (changes requested):
Thanks @marek-mihok. A couple of comments.

ui/src/table.test.tsx
mturoci: Not resolved yet, I still see a bunch of `!.firstChild.toBeTruthy`

mturoci (approved):

PR #1347, opened 2022-04-12
feat: Allow specifying path for button

### LINKED ISSUE #1329: Allow specifying path for button
### Discussed in https://github.com/h2oai/wave/discussions/1318

Add a `path` prop to button which opens an external URL in a new tab after clicking.

marek-mihok:
Adds optional `path` prop to ui.button component:
```ts
export interface Button {
  /** An identifying name for this component. If the name is prefixed with a '#', the button sets the location hash to the name when clicked. */
  name: Id
  /** The text displayed on the button. */
  label?: S
  /** The caption displayed below the label. */
  caption?: S
  /** A value for this button. If a value is set, it is used for the button's submitted instead of a boolean True. */
  value?: S
  /** True if the button should be rendered as the primary button in the set. */
  primary?: B
  /** True if the button should be disabled. */
  disabled?: B
  /** True if the button should be rendered as link text and not a standard button. */
  link?: B
  /** An optional icon to display next to the button label (not applicable for links). */
  icon?: S
  /** The width of the button, e.g. '100px'. */
  width?: S
  visible?: B
  tooltip?: S
  /** The path or URL to link to. */
  path?: S
}
```

Closes #1329

### DIFF
--- a/ui/src/button.test.tsx
+++ b/ui/src/button.test.tsx
@@ -85,8 +85,7 @@ describe('Button.tsx', () => {
     it('Does redirect if path is specified', () => {
       const
         windowOpenMock = jest.fn(),
-        btnPathProps: Buttons = { items: [{ button: { name, label: name, path: 'https://h2o.ai/' } }] },
-        { getByTestId } = render(<XButtons model={btnPathProps} />)
+        { getByTestId } = render(<XButtons model={{ items: [{ button: { name, label: name, path: 'https://h2o.ai/' } }] }} />)
 
       window.open = windowOpenMock
       fireEvent.click(getByTestId(name))
--- a/ui/src/button.test.tsx
+++ b/ui/src/button.test.tsx
@@ -15,7 +15,6 @@
 import { fireEvent, render } from '@testing-library/react'
 import React from 'react'
 import { Buttons, MiniButtons, XButtons, XMiniButtons } from './button'
-import { Component } from './form'
 import { wave } from './ui'
 
 const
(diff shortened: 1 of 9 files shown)


ui/src/button.test.tsx
```diff
@@ -82,6 +81,17 @@ describe('Button.tsx', () => {
 
       expect(queryByTestId(name)).not.toHaveClass('ms-Link')
     })
+
+    it('Does redirect if path is specified', () => {
+      const
+        windowOpenMock = jest.fn(),
+        btnPathProps: Buttons = { items: [{ button: { name, label: name, path: 'https://h2o.ai/' } }] },
+        { getByTestId } = render(<XButtons model={btnPathProps} />)
```
mturoci, line 89:
```suggestion
        { getByTestId } = render(<XButtons model={{ items: [{ button: { name, label: name, path: 'https://h2o.ai/' } }] }} />)
```

mturoci (dismissed):
Thanks @marek-mihok. LGTM except 1 small comment.

cc @lo5 for API review - another option is to not open all paths in a new tab by default, but specify a `target`  attr instead.

ui/src/button.tsx
```diff
@@ -62,6 +62,8 @@ export interface Button {
   visible?: B
   /** An optional tooltip message displayed when a user clicks the help icon to the right of the component. */
   tooltip?: S
+  /** The path or URL to link to. */
```
lo5, line 65:
```suggestion
  /** The path or URL to link to. If specified, the `name` is ignored. The URL is opened in a new browser window or tab. */
```

lo5 (approved):
Looks good! Added a suggestion to improve API docstring.

PR #1350, opened 2022-04-13
fix: Add placeholder to dropdown with dialog

### LINKED ISSUE #1349: Missing placeholder for dropdown with dialog
via [azim-b](https://github.com/azim-b)

#### Actual behavior

Provided placeholder doesn't show up in the dropdown text input field when the dialog is enabled.

#### Expected behavior

Display placeholder.

#### Steps To Reproduce

```py

ui.dropdown(name='dropdown', label='Pick multiple', placeholder='test', choices=choices_dialog)
```

aalencar:
Also removed the placeholder from the search box.

https://user-images.githubusercontent.com/15820796/163210499-418df6b4-9301-43fc-8715-cd0719fa0805.mov

closes #1349

### DIFF
--- a/py/examples/dropdown.py
+++ b/py/examples/dropdown.py
@@ -34,7 +34,7 @@ async def serve(q: Q):
             ui.dropdown(name='dropdown_disabled', label='Pick one (Disabled)', value='B', choices=choices,
                         disabled=True),
             ui.dropdown(name='dropdown_dialog', label='Pick multiple in dialog (>100 choices)', values=['1'],
-                        required=True, choices=choices_dialog),
+                        required=True, choices=choices_dialog, placeholder='Select values...'),
             ui.dropdown(name='dropdown_popup_always', label='Always show popup even when choices < 100', value='A',
                         required=True, choices=choices, popup='always'),
             ui.dropdown(name='dropdown_popup_never', label='Never show popup even when choices > 100', value='1',
--- a/py/examples/dropdown.py
+++ b/py/examples/dropdown.py
@@ -34,7 +34,7 @@ async def serve(q: Q):
             ui.dropdown(name='dropdown_disabled', label='Pick one (Disabled)', value='B', choices=choices,
                         disabled=True),
             ui.dropdown(name='dropdown_dialog', label='Pick multiple in dialog (>100 choices)', values=['1'],
-                        required=True, choices=choices_dialog, placeholder='Select values...'),
+                        required=True, choices=choices_dialog),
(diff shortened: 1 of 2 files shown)


ui/src/dropdown.tsx
```diff
@@ -219,11 +219,11 @@ const
           disabled={disabled}
           required={required}
           styles={{ field: { cursor: 'pointer' }, icon: { fontSize: 12, color: cssVar('$neutralSecondary') } }}
-          value={textValue || ''} />
+          value={textValue || placeholder || ';'} />
```
mturoci, line 222: The correct way should be using `placeholder` instead of `value`.

py/examples/dropdown.py
```diff
@@ -34,7 +34,7 @@ async def serve(q: Q):
             ui.dropdown(name='dropdown_disabled', label='Pick one (Disabled)', value='B', choices=choices,
                         disabled=True),
             ui.dropdown(name='dropdown_dialog', label='Pick multiple in dialog (>100 choices)', values=['1'],
-                        required=True, choices=choices_dialog),
+                        required=True, choices=choices_dialog, placeholder='Select values...'),
```
mturoci, line 37: Placeholder should not replace a label. See https://www.nngroup.com/articles/form-design-placeholders/ for more UX insights.

mturoci (changes requested):
I am not convinced `placeholder` attr makes sense for dropdown UX-wise, but since we already support them, this should probably get fixed.

py/examples/dropdown.py
aalencar: Sounds like it can cause confusion indeed. 

From your link:
> [placeholder] is reserved for supplementary information.

In this case, should we provide a placeholder for these examples?

mturoci: I cannot think of a valid scenario (but maybe there is) for a dropdown with a placeholder. Placeholder serves to show example values - primary case or some extra info needed, but with dropdown example / extra info are not needed since options are constrained. It's a whole different story for textboxes/comboboxes where a user is allowed to specify arbitrary strings.

Let's not use placeholder in this dropdown example, but make sure it is displayed correctly if specified.

aalencar: Agreed. 
The latest changes don't specify `placeholder` for examples and display it correctly if specified.
Besides that, I removed a hardcoded placeholder "Select ...", it was being displayed in the textbox after the user deselects all options.

mturoci (approved):
Looking good @aalencar, thanks!

PR #1352, opened 2022-04-14
docs: Add examples in layout docs about the size value

### LINKED ISSUE #1252: Add examples in Layout Docs about the size value
Many H2O Wave power users have ran into and solved the following "problem":

**What you want**
image
 **What you get**
image

This is because the default size for all cards is `1` so the zone is split equally. The solution is to set the size of the "small thing" to 0 like:
```python
    q.page["tab_card"] = ui.tab_card(
        box=ui.box("focus_table", size="0"),
        items=[ui.tab("1", "1")]
    )
    q.page["form2"] = ui.form_card(box='focus_table', items=[
        ui.text("focus table")
    ])
```

**The default behavior makes sense and once people see the solution they get it - no request in functionality is being requested**. The problem is, 4+ power users have asked this same question, so something is probably missing from the docs / examples

Add a subsection to the Layouts documentation page showing this exactly use case. This should explain (again) the default behavior, why it exists, and what to do in these cases.

aalencar:
image

image

image

closes #1252


mtanco (approved):
This is great, thank you!

mturoci (approved):
LGTM, thanks @aalencar!

Issue #795, opened 2021-05-07
Form items width control


sandruparo:
The input fields width should be adjustable , currently it takes the whole body like a block element. we should make it flexible  to define by the dev.

![image](https://user-images.githubusercontent.com/9942408/117422248-05329900-af3d-11eb-8249-b7e873291c02.png)

@lo5 @shanaka-rajitha @MartinBarus @mtanco

dulajra:
To add to this, the width should be customizable when the components are placed inline also. Currently the `ui.textbox` has a fixed width so that when the text is long the content is not fully readable. 
![Screenshot 2021-06-21 at 17 20 24](https://user-images.githubusercontent.com/12434380/122769235-06880d00-d2c2-11eb-8d52-77438bbdf345.png)

mturoci:
@lo5 I feel like this could make sense for every form item (similar to how `visible` is implemented), that way simple designs like this
![image](https://user-images.githubusercontent.com/64769322/129567266-0208ce08-2a6e-44ea-83f0-43c513b866c4.png)

could be simply implemented within a `form_card`.

lo5:
@mturoci +1, would be good to have on all form components.

PR #1397, opened 2022-05-04
fix: remove scrollbar from stats

### LINKED ISSUE #1376: Unnecessary scrollbar in ui.stats
### Discussed in https://github.com/h2oai/wave/discussions/1368

<sup>Originally posted by **azim-b** April 20, 2022</sup>
broswer: Safari Version 15.2
Wave version 0.21.0

Issue: `ui.stats()` rendering a scroll bar. Example:

image

Example 2:

image

**Cannot** reproduce this in wave 0.20.0

**Can** partially reproduce this in wave v0.21.0 in Firefox.
By partially, I mean that the scroll bars are rendered and then immediately disappear.</div>

### Repro

```py
q.page['body'] = ui.form_card(box='body',items=[
        # without inset
        ui.separator('Without inset'),
        ui.stats(justify='center',items=[
            ui.stat(label='H2O AI Health', value='AI4Good', icon='HealthSolid', icon_color='Red'),
        ]),
        # with inset set
        ui.separator('With inset'),
        ui.stats(justify='center', inset=True, items=[
            ui.stat(label='H2O AI Health', value='AI4Good', icon='HealthSolid', icon_color='Red'),
        ])
    ])
```

aalencar:
Taking into account #795, stats will not overflow horizontally since the label is going to wrap when it's too long.

After removing `overflow-x: auto`:
image

@mturoci let me know if I'm missing something on this one since you added `overflow-x: auto`.

closes #1376

### DIFF
--- a/ui/src/stats.tsx
+++ b/ui/src/stats.tsx
@@ -66,7 +66,8 @@ const
       marginRight: '0.5em',
     },
     icon: {
-      fontSize: 45,
+      height: 45,
+      width: 45,
     },
     statCaption: {
       color: cssVar('$text5')
@@ -97,7 +98,7 @@ export const
         <div key={`${i}:${label}`} className={css.stat} style={statStyle}>
           {icon && (
             <div className={css.lhs} style={icon_color ? { color: cssVar(icon_color) } : undefined}>
-              <Icon className={css.icon} iconName={icon} styles={{ root: { '> span': { height: '100%'}}}}/>
+              <Icon className={css.icon} iconName={icon} styles={{ root: { '> span': { height: '100%', width: '100%'}}}}/>
             </div>
           )}
           <div>
--- a/ui/src/stats.tsx
+++ b/ui/src/stats.tsx
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { FontIcon } from '@fluentui/react'
+import { Icon } from '@fluentui/react'
 import { B, Dict, S } from 'h2o-wave'
 import React from 'react'
 import { stylesheet } from 'typestyle'
@@ -97,7 +97,7 @@ export const
         <div key={`${i}:${label}`} className={css.stat} style={statStyle}>
           {icon && (
             <div className={css.lhs} style={icon_color ? { color: cssVar(icon_color) } : undefined}>
-              <FontIcon className={css.icon} iconName={icon} />
+              <Icon className={css.icon} iconName={icon} styles={{ root: { '> span': { height: '100%'}}}}/>
             </div>
           )}
           <div>


mturoci:
I am not sure why/how does removing `overflowX` solve the issue? According to screenshots, the problematic scrollbar is vertical and overflowX deals with horizontal one.

aalencar:
> I am not sure why/how does removing `overflowX` solve the issue? According to screenshots, the problematic scrollbar is vertical and overflowX deals with horizontal one.

[Check this answers on Stackoverflow](https://stackoverflow.com/questions/6421966/css-overflow-x-visible-and-overflow-y-hidden-causing-scrollbar-issue/6433475#6433475).

Another solution would be to leave `overflow-x: auto` and set `overflow-y: hidden` because the card grows vertically to fit the content and won't overflow.

mturoci:
The overflow-x is needed for cases when there are more stat items:
image

> [Check this answers on Stackoverflow](https://stackoverflow.com/questions/6421966/css-overflow-x-visible-and-overflow-y-hidden-causing-scrollbar-issue/6433475#6433475).

I see, didn't know about that one (+1). Unfortunately, this fix is more of "ignoring" the overflow rather than fixing it. See the following section for a detailed explanation.

> Another solution would be to leave overflow-x: auto and set overflow-y: hidden because the card grows vertically to fit the content and won't overflow.

overflow-y: hidden is just a way of ignoring the real problem. The real problem is that some element is overflowing the container when it shouldn't. So let's identify the problematic element, see why it overflows, and fix it accordingly. That's how unnecessary scrollbar problems should be handled in general.

aalencar:
Changing the height of the svg container (span) from 1em to 100% fixed the issue.

ui/src/stats.tsx
```diff
@@ -97,7 +97,7 @@ export const
         <div key={`${i}:${label}`} className={css.stat} style={statStyle}>
           {icon && (
             <div className={css.lhs} style={icon_color ? { color: cssVar(icon_color) } : undefined}>
-              <FontIcon className={css.icon} iconName={icon} />
+              <Icon className={css.icon} iconName={icon} styles={{ root: { '> span': { height: '100%'}}}}/>
```
mturoci, line 100: The font icon styling is not necessary anymore. Make sure the rendered icon is 45x45.
```suggestion
              <Icon iconName={icon} styles={{ root: { '> span': { height: '100%'}}}}/>
```

mturoci (changes requested):
Thanks @aalencar, needs minor code clean up.

ui/src/stats.tsx
aalencar: Changed `font-size` -> `height` + `width`
The icon is not rendered if the `height` or `width` is not set for `<i />`

mturoci (approved):
Thanks @aalencar!

PR #1400, opened 2022-05-04
feat: Image annotator

### LINKED ISSUE #1323: Image annotator component
via @dott1718

Similar to the text annotation component https://github.com/h2oai/wave/issues/227, the box annotation component should have three major properties: 

* image
* tags: list of tags
* items: list of boxes

The image is displayed as a background. The tags are rendered as clickable elements that tag the highlighted boxes. The items are rendered as semi-transparent colored rectangles and can be deleted.
Each box is described by the corresponding tag and 4 numbers.

```
xmin, ymin, xmax, ymax
```

It should be easy to draw new boxes. Bounding boxes could have overlaps or even contain other boxes. Selecting a single box it should be easy to:
 
* Drag and move the box
* Resize the box by the corners
* Delete the box
* Change the tag

mturoci:
## Demo

https://user-images.githubusercontent.com/64769322/166687217-85c0c67c-e6d9-4e30-b1a8-51172e5e70ed.mov

## Proposed API

```ts
/** Rectangle (box) annotation. */
interface ImageAnnotatorRect {
  /** Start X dimension. */
  x1: U
  /** End X dimension. */
  x2: U
  /** Start Y dimension. */
  y1: U
  /** End Y dimension. */
  y2: U
}

/** Defines a particular shape to be used for the annotation. */
interface ImageAnnotatorShape {
  rect: ImageAnnotatorRect
}

/** Image annotator tag. */
interface ImageAnnotatorTag {
  /** An identifying name for this tag. */
  name: Id
  /** Text to be displayed. */
  label: S
  /** HEX or RGB color string used as background for highlighted phrases. */
  color: S
}

/** Create an annotator item with initial selected tags or no tag for plaintext. */
interface ImageAnnotatorItem {
  /** The annotation shape. */
  shape: ImageAnnotatorShape
  /** Tag connected to the highlighted section. */
  tag: S
}

/**
 * Create an image annotator component.
 * 
 * The image annotator component enables user to manually annotate parts of an image.
 */
export interface ImageAnnotator {
  name: Id
  /** The image to annotate. */
  image: S
  /** The text annotator's title. */
  title: S
  /** List of tags the user can annotate with. */
  tags: ImageAnnotatorTag[]
  /** Existing annotations if any. */
  items?: ImageAnnotatorItem[]
  /** True if the form should be submitted when the annotator value changes. */
  trigger?: B
  /** The card’s image height. Intrinsic image size is used by default. */
  image_height?: S
}
```

Returning `ImageAnnotatorItem `.

@lo5 please have a look at the API - I tried to design it in a way so that we can support multiple shapes (circles/paths) in future as well. Let me know if you see room for improvement though.

## Notes

* Unit tests / docs to be delivered once the API is settled.

cc @dott1718 for shallow feedback based on just video. I expect more detailed feedback after you play around with it in nightly.
Closes #1323

### DIFF
--- a/py/examples/image_annotator.py
+++ b/py/examples/image_annotator.py
@@ -24,10 +24,7 @@ async def serve(q: Q):
                     ui.image_annotator_tag(name='f', label='Food', color='$blue'),
                 ],
                 items=[
-                    ui.image_annotator_item(
-                        shape=ui.image_annotator_shape(rect=ui.image_annotator_rect(x1=649, y1=393, x2=383, y2=25)),
-                        tag='p'
-                    ),
+                    ui.image_annotator_item(shape=ui.image_annotator_rect(x1=649, y1=393, x2=383, y2=25), tag='p'),
                 ],
             ),
             ui.button(name='submit', label='Submit', primary=True)
--- a/py/examples/image_annotator.py
+++ b/py/examples/image_annotator.py
@@ -25,7 +25,7 @@ async def serve(q: Q):
                 ],
                 items=[
                     ui.image_annotator_item(
-                        shape=ui.image_annotator_shape(rect=ui.image_annotator_rect(x1=569, y1=361, x2=364, y2=22)),
+                        shape=ui.image_annotator_shape(rect=ui.image_annotator_rect(x1=649, y1=393, x2=383, y2=25)),
                         tag='p'
                     ),
                 ],
--- a/py/examples/image_annotator.py
+++ b/py/examples/image_annotator.py
@@ -13,12 +13,12 @@ async def serve(q: Q):
             ui.button(name='show_form', label='Back', primary=True),
         ]
     else:
-        q.page['example'] = ui.form_card(box='1 1 4 7', items=[
+        q.page['example'] = ui.form_card(box='1 1 9 -1', items=[
             ui.image_annotator(
                 name='annotator',
                 title='Drag to annotate',
                 image='https://images.pexels.com/photos/2696064/pexels-photo-2696064.jpeg?auto=compress&cs=tinysrgb&w=1260&h=750&dpr=1',
-                image_height='300px',
+                image_height='700px',
                 tags=[
                     ui.image_annotator_tag(name='p', label='Person', color='#F1CBCB'),
                     ui.image_annotator_tag(name='f', label='Food', color='#CAEACA'),
--- a/py/examples/image_annotator.py
+++ b/py/examples/image_annotator.py
@@ -20,8 +20,8 @@ async def serve(q: Q):
                 image='https://images.pexels.com/photos/2696064/pexels-photo-2696064.jpeg?auto=compress&cs=tinysrgb&w=1260&h=750&dpr=1',
                 image_height='700px',
                 tags=[
-                    ui.image_annotator_tag(name='p', label='Person', color='#F1CBCB'),
-                    ui.image_annotator_tag(name='f', label='Food', color='#CAEACA'),
+                    ui.image_annotator_tag(name='p', label='Person', color='$cyan'),
+                    ui.image_annotator_tag(name='f', label='Food', color='$blue'),
                 ],
                 items=[],
             ),
--- a/py/examples/image_annotator.py
+++ b/py/examples/image_annotator.py
@@ -10,7 +10,7 @@ async def serve(q: Q):
     if q.args.annotator is not None:
         q.page['example'].items = [
             ui.text(f'annotator={q.args.annotator}'),
-            ui.button(name='show_form', label='Back', primary=True),
+            ui.button(name='back', label='Back', primary=True),
         ]
     else:
         q.page['example'] = ui.form_card(box='1 1 9 -1', items=[
--- a/py/examples/image_annotator.py
+++ b/py/examples/image_annotator.py
@@ -7,7 +7,7 @@ from h2o_wave import main, app, Q, ui
 
 @app('/demo')
 async def serve(q: Q):
-    if q.args.annotator:
+    if q.args.annotator is not None:
         q.page['example'].items = [
             ui.text(f'annotator={q.args.annotator}'),
             ui.button(name='show_form', label='Back', primary=True),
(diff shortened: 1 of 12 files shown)


dott1718:
Thank you @mturoci, the component looks great! We will provide more feedback as soon as we can try it out.

lo5:
Demo looks great :guitar:!

It's not clear to me how `items` and `ImageAnnotatorItem`, `ImageAnnotatorShape`, `ImageAnnotatorRect` are intended to be used. The one example in this PR doesn't use any of them. Maybe that'll help clarify why `tag` is a string, and why `rect` is an attribute of `~Shape`. It would be good to add another example that show those components in use.

mturoci:
> It's not clear to me how items and ImageAnnotatorItem, ImageAnnotatorShape, ImageAnnotatorRect are intended to be used.

I updated the existing example to show predefined annotation. The basic reasoning is:

* Tag is a string so that we are consistent with tag in `ui.text_annotator` + there are not many other options.
* The `ImageAnnotatorShape` is a bit more tricky, but the overarching goal is to do us a favor and allow for extending the image annotator with other shapes in the future, e.g. circle/paths. I took this approach to be consistent with how we define our table cell types.

mturoci:
@lo5 one more thing that is worth discussing is the image aspect ratio. Since images to be annotated can be various sizes, we have basically 2 options:

* Display them in their original (intrinsic) size even at the price of scrollbars (in case of overflow).
* Provide `height` attr and scale the image accordingly - I went for this option.

Since I chose the second option as it is more UX-friendly, I had to make adjustments to the coordinates based on the aspect ratio so that developers can specify and receive coordinates relative to the original size of their img. Example:

Original img: 100x100 with coordinates 10,10.
Scaled image with height 50px: 50x50 with coordinates translated to 5,5.

The proposed approach makes sense to me, but feel free to point out any alternatives as you see fit.

lo5:
OK, so if I'm understanding this correctly, `ImageAnnotatorItem.tag` looks up an *existing* `ImageAnnotatorTag.name` to determine the label and color?

mturoci:
>OK, so if I'm understanding this correctly, ImageAnnotatorItem.tag looks up an existing ImageAnnotatorTag.name to determine the label and color?

That's right.

ui/src/image_annotator.tsx
```diff
@@ -0,0 +1,356 @@
...
+  label: S
+  /** HEX or RGB color string used as background for highlighted phrases. */
+  color: S
+}
+
+/** Create an annotator item with initial selected tags or no tag for plaintext. */
+interface ImageAnnotatorItem {
+  /** The annotation shape. */
+  shape: ImageAnnotatorShape
+  /** Tag connected to the highlighted section. */
```
lo5, line 40:
```suggestion
  /** The `name` of the image annotator tag to refer to for the `label` and `color` of this item. */
```

(Also change the docstring for the text annotator item if possible).

ui/src/image_annotator.tsx
```diff
@@ -0,0 +1,356 @@
...
+  rect: ImageAnnotatorRect
+}
+
+/** Image annotator tag. */
+interface ImageAnnotatorTag {
+  /** An identifying name for this tag. */
+  name: Id
+  /** Text to be displayed. */
+  label: S
+  /** HEX or RGB color string used as background for highlighted phrases. */
```
lo5, line 32:
```suggestion
  /** Hex or RGB color string to be used as the background color. */
```

ui/src/image_annotator.tsx
```diff
@@ -0,0 +1,356 @@
...
+/** Defines a particular shape to be used for the annotation. */
+interface ImageAnnotatorShape {
+  rect: ImageAnnotatorRect
+}
+
+/** Image annotator tag. */
+interface ImageAnnotatorTag {
+  /** An identifying name for this tag. */
+  name: Id
+  /** Text to be displayed. */
```
lo5, line 30:
```suggestion
  /** Text to be displayed for the annotation. */
```

ui/src/image_annotator.tsx
```diff
@@ -0,0 +1,356 @@
...
+  /** The annotation shape. */
+  shape: ImageAnnotatorShape
+  /** Tag connected to the highlighted section. */
+  tag: S
+}
+
+/**
+ * Create an image annotator component.
+ * 
+ * The image annotator component enables user to manually annotate parts of an image.
```
lo5, line 47:
```suggestion
 * This component allows annotating and labeling parts of an image by drawing shapes with a pointing device.
```

ui/src/image_annotator.tsx
```diff
@@ -0,0 +1,356 @@
...
+ */
+export interface ImageAnnotator {
+  /** An identifying name for this component. */
+  name: Id
+  /** The image to annotate. */
+  image: S
+  /** The text annotator's title. */
```
lo5, line 54:
```suggestion
  /** The image annotator's title. */
```

lo5 (changes requested):

ui/src/image_annotator.tsx
```diff
@@ -0,0 +1,356 @@
...
+  title: S
+  /** List of tags the user can annotate with. */
+  tags: ImageAnnotatorTag[]
+  /** Existing annotations if any. */
+  items?: ImageAnnotatorItem[]
+  /** True if the form should be submitted when the annotator value changes. */
+  trigger?: B
+  /** The card’s image height. Intrinsic image size is used by default. */
```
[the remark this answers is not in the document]
mturoci: Intrinsic means actual (real) image size.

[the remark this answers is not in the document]
mturoci: Changed to `The card’s image height. The actual image size is used by default`.

py/examples/image_annotator.py
```diff
@@ -0,0 +1,35 @@
...
+                title='Drag to annotate',
+                image='https://images.pexels.com/photos/2696064/pexels-photo-2696064.jpeg?auto=compress&cs=tinysrgb&w=1260&h=750&dpr=1',
+                image_height='700px',
+                tags=[
+                    ui.image_annotator_tag(name='p', label='Person', color='$cyan'),
+                    ui.image_annotator_tag(name='f', label='Food', color='$blue'),
+                ],
+                items=[
+                    ui.image_annotator_item(
+                        shape=ui.image_annotator_shape(rect=ui.image_annotator_rect(x1=649, y1=393, x2=383, y2=25)),
```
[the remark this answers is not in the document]
mturoci: Much better now.

Issue #1004, opened 2021-09-17
Card layout with multi text sections and avatars


shanaka-rajitha:
# Features

- Display multiple rows of information.
- Bottom section of card will show clickable avatars.

## Card layout with Avatars
<img width="837" alt="Card layout" src="https://user-images.githubusercontent.com/12246331/133738747-48f724b9-f439-4c33-8f44-f01edacb74a4.png">

mturoci:
@shanaka-rajitha as discussed on Slack, this card is implementable as a `form_card`. The only feature missing is bottom facepile, please edit the issue accordingly.

shanaka-rajitha:
Hi @mturoci the following feature card is part of the MLOps screens, so its vital we get the entire card with all the sections. We will plan to use the same card on other future designs too.

mturoci:
IMO the proposed component is too specific for Wave, wdyt @lo5? Is this the path we want to take?

shanaka-rajitha:
Including @mtanco @iamabhishekmathur

iamabhishekmathur:
@mturoci this card does indeed contain a lot of information ... how else do you recommend it be implemented, with this many data elements?

mturoci:
Wave is currently capable of this:

![image](https://user-images.githubusercontent.com/64769322/134674760-68ecd5be-c131-4646-b547-78475c29f1d6.png)

The only 2 missing pieces are bottom facepile component and spacing.

The facepile component will be implemented, that's the easy part.

Note that one of the Wave goals is to provide good-looking usable UIs with zero knowledge of UI/UX design. That means that we should enhance our existing `ui.text` component that is currently only capable of specifying `size` in a way that it is impossible to build bad looking cards. For example:

We can think of mapping our existing text components in following way:
* `ui.text_xl` -> title
* `ui.text_l` -> subtitle
* `ui.text_m` -> regular text content
and so on...

We can then assume that if title and subtitle are next to each other, they should have just a minimal spacing. Regular text content on the other hand should probably have more spacing in relation to header part etc. Hope you all got the point. This is something for design team to lay out as they are the typography experts.

The current text sizes need refinements based on - https://xd.adobe.com/view/0790f950-abbd-41fa-b372-332295fd876f-52c3/screen/0eaf1110-4f69-40da-9417-3f764032248f/variables/?x_product=cc-slack%2F1.5.1.

This way, we can easily keep typography consistency across all the Wave apps, help people make less decisions and reduce Wave development time.

[...]

PR #1408, opened 2022-05-06
feat: add support to align and justify form card items

### LINKED ISSUE #556: Allow justifying elements vertically
I would have the following two feature requests with respect to justification:

- Currently, it does not seem possible to justify elements vertically, only horizontally. This looks specifically odd for buttons or similar objects being placed in form cards if they cannot be vertically aligned (e.g., center).
- Extend the justification options of `ui.inline` to also allow `center` alignment.

aalencar:
This PR adds a new container `Column` that allows wave app developers lay out components vertically within a form.

Same interface as `Inline`. I suppose we have to keep it in order to generate the types in python.
```ts
/** Creates a container to lay out components along the vertical axis. */
interface Column {
  items: Component[]
  inset?: B
}
```

Also, it adds support to position / align all components (in the flexbox model `justify-content` and `align-items`) using the new parameters `justify` (horizontally) and `align` (vertically).

```ts
/** Create a form. */
interface State {
  ...
  /** Justify all components along the horizontal axis. */
  justify?: 'start' | 'center' | 'end' | 'around'
  /** Defines the alignment along the vertical axis. */
  align?: 'start' | 'center' | 'end'
}
```

## Example
The first 9 cards show the usage of `justify` and `align` for the whole form card and the last one using the `Column` component.
image

Question: Can we deprecate `ui.inline` in favor of `ui.row`? It would make more sense if we introduce `ui.column`?

The code needs some refinements but I would like to validate this approach before proceeding with docs and examples.

closes #556

### DIFF
--- a/py/examples/form_alignment.py
+++ b/py/examples/form_alignment.py
@@ -1,0 +1,65 @@
+from h2o_wave import main, Q, ui, app
+
+
+@app('/demo')
+async def serve(q: Q):
+
+    q.page['form1'] = ui.form_card(box='1 1 3 3', title='justify start (default)', items=[
+        ui.text_xl(content='Text'),
+        ui.button(name='btn', label='click me')
+    ])
+
+    q.page['form2'] = ui.form_card(box='4 1 3 3', justify='center', items=[
+        ui.text_xl(content='Text'),
+        ui.button(name='btn', label='click me')
+    ])
+
+    q.page['form3'] = ui.form_card(box='7 1 3 3', title='justify end', justify='end', items=[
+        ui.text_xl(content='Text'),
+        ui.button(name='btn', label='click me')
+    ])
+
+    q.page['form4'] = ui.form_card(box='1 4 3 3', title='justify start + align center', align='center', items=[
+        ui.text_xl(content='Text'),
+        ui.button(name='btn', label='click me')
+    ])
+
+    q.page['form5'] = ui.form_card(box='4 4 3 3', title='justify center + align center', justify='center', align='center', items=[
+        ui.text_xl(content='Text'),
+        ui.button(name='btn', label='click me')
+    ])
+
+    q.page['form6'] = ui.form_card(box='7 4 3 3', title='justify end + align center', justify='end', align='center', items=[
+        ui.text_xl(content='Text'),
+        ui.button(name='btn', label='click me')
+    ])
+
+    q.page['form7'] = ui.form_card(box='1 7 3 3', title='justify start + align end', align='end', items=[
+        ui.text_xl(content='Text'),
+        ui.button(name='btn', label='click me')
+    ])
+
+    q.page['form8'] = ui.form_card(box='4 7 3 3', title='justify center + align end', justify='center', align='end', items=[
+        ui.text_xl(content='Text'),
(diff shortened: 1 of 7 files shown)


mturoci:
> Also, it adds support to position / align all components (in the flexbox model justify-content and align-items) using the new parameters justify (horizontally) and align (vertically).

According to your examples, it seems like there are 2 kinds of alignment possible now:

* Card-wide alignment - justify/align on form_card.
* Group alignment - justify/align on inline/row/col.
 
It would be good though to have only one way of doing things. IMO the group alignment is more flexible so I would go with that one. 

> Can we deprecate ui.inline in favor of ui.row? It would make more sense if we introduce ui.column?

Deprecating `ui.inline` would be a massive breaking change. Wdyt about adding a `direction` (row/col) prop to the `ui.inline` instead of introducing `ui.row`/`ui.col`?

> The code needs some refinements but I would like to validate this approach before proceeding with docs and examples.

+1

What are the scenarios this feature tries to solve? I can think of maybe [this one](https://github.com/h2oai/wave/issues/1004), which is achievable today but is quite hacky.

mturoci:
> Also, it adds support to position / align all components (in the flexbox model justify-content and align-items) using the new parameters justify (horizontally) and align (vertically).

According to your examples, it seems like there are 2 kinds of alignment possible now:

* Card-wide alignment - justify/align on form_card.
* Group alignment - justify/align on inline/row/col.
 
It would be good though to have only one way of doing things. IMO the group alignment is more flexible so I would go with that one. 

> Can we deprecate ui.inline in favor of ui.row? It would make more sense if we introduce ui.column?

Deprecating `ui.inline` would be a massive breaking change. Wdyt about adding a `direction` (row/col) prop to the `ui.inline` instead of introducing `ui.row`/`ui.col`?

> The code needs some refinements but I would like to validate this approach before proceeding with docs and examples.

+1

What are the scenarios this feature tries to solve? I can think of maybe [this one](https://github.com/h2oai/wave/issues/1004), which is achievable today but is quite hacky.

aalencar:
Thanks for the feedback @mturoci.

> It would be good though to have only one way of doing things. IMO the group alignment is more flexible so I would go with that one.

The card-wide alignment can be handy in some situations like centering all elements without using `ui.column` or `ui.inline`. If we can replicate most scenarios with `ui.column` and `ui.inline` we can remove the card-wide alignment because it can be confusing for someone not used to these mental models.

> Deprecating ui.inline would be a massive breaking change. 

We wouldn't have to remove `ui.inline`, keep it for backward compatibility and introduce the same interface with a different name. That would have duplicate code though.

> Wdyt about adding a direction (row/col) prop to the ui.inline instead of introducing ui.row/ui.col?

Usually inline means a horizontal line. Introducing direction probably is more concise but not as semantic as `ui.column`.

> What are the scenarios this feature tries to solve? I can think of maybe https://github.com/h2oai/wave/issues/1004, which is achievable today but is quite hacky.

Align elements along the vertical axis.

mturoci:
>Align elements along the vertical axis.

Can you provide some real-world scenarios?

aalencar:
Lacking concrete use cases.

PR #1411, opened 2022-05-11
fix: remove unnecessary scrollbar from menu

### LINKED ISSUE #1395: Unnecessary scroll bar in `ui.menu()` commands
macOS, Wave 0.21.0, Safari Version 15.4

Unnecessary scroll bar in `ui.menu()` commands

image

Code:

```py
ui.menu(name='user', items=[
    ui.command(name='info', label='About', icon='Info'),
    ui.command(name='admin', label='Admin', icon='Admin'),
    ui.command(name='logout', label='Logout', icon='Signout')
])
```

aalencar:
## Before

image

## After

image

Tested on Safari 15.1, Chrome 101.0, and Firefox 100.0.

## Fix

As in #1376 , we know that the icon is the root cause but strangely no container has more height than its parent.

This time it couldn't be solved by simply changing the height and width values, I had to change `vertical-align` which was causing trouble.

Fluent Icon component has the following structure:
```html
<i> vertical-align middle
  <span> vertical-align middle
     <svg> vertical-align top
```

Setting `vertical-align` of span to `baseline` (default value) removes the scrollbar but the SVG's `vertical-align: top` kicks in:

image

Removing the SVG `vertical-align: top` helps but still misaligned:

image

With the offset `vertical-align:-3px`:

image

closes #1395


aalencar:
> Seems like the real problem is `line-height` which adds extra space to both top and bottom parts of the text line, causing overflow and not being taken into account when computing height - it's still reported as `36px`.
> 
>  Screen.Recording.2022-05-11.at.11.06.22.AM.mov 
> Let's try to go that route and see if it reliably fixes the issue across all the browsers.

better.

ui/src/menu.tsx
```diff
@@ -58,7 +58,7 @@ export const XMenu = ({ model }: { model: Menu }) => {
         isBeakVisible
         directionalHint={Fluent.DirectionalHint.bottomRightEdge}
         calloutProps={{ styles: { beak: { border: border(1, cssVar('$neutralQuaternaryAlt')) } } }}
-        styles={{ list: { border: border(1, cssVar('$neutralQuaternaryAlt')) } }}
+        styles={{ list: { border: border(1, cssVar('$neutralQuaternaryAlt')), 'button': { lineHeight: '1em'} } }}
```
mturoci, line 61: Let's be more specific here and use the class (`.ms-ContextualMenu-link` seems like a good fit) instead of targeting every button component. Why is lineHeight set to `1em`? `unset` seems like a better fit.

```suggestion
        styles={{ list: { border: border(1, cssVar('$neutralQuaternaryAlt')), 'button': { lineHeight: 'unset'} } }}
```

mturoci (changes requested):
Thanks @aalencar. We need to make sure that nested menus are not broken as well.

```py
commands = [
    ui.command(name='info', label='About', icon='Info'),
    ui.command(name='admin', label='Admin', icon='Admin'),
    ui.command(name='logout', label='Logout', icon='Signout', items=[
        ui.command(name='info', label='About', icon='Info'),
        ui.command(name='admin', label='Admin', icon='Admin'),
    ])
]
```
image

aalencar:
Fixed

image

mturoci (changes requested):

mturoci:
Also please check the rest of ContextualMenus throughout Wave if they do not suffer from the same bug.

aalencar:
Ready for review.

mturoci (dismissed):
Thanks @aalencar! LGTM. Does the context menu in the table work fine?

aalencar:
> Does the context menu in the table work fine?

Yes, we only have problems when the contextual menu has icons, which is not the case for the table menu.

mturoci:
> which is not the case for the table menu.

Doesn't the table menu support icons?

aalencar:
> Fixed
> 
> image

I found out that this is not a good example for submenus because the `Admin` icon is not working and we only get the scrollbar when the last item has an icon. If we change to `Delete` icon we get the scrollbar again. 

image

Looks like the styles are not being applied for submenus container. Working on a fix.

aalencar:
### Fix for submenus

Repro code:
```py
commands = [
        ui.command(name='info1', label='About', icon='Info'),
        ui.command(name='delete1', label='Delete', icon='Delete'),
        ui.command(name='logout1', label='Logout', icon='Signout', items=[
            ui.command(name='info2', label='About', icon='Info'),
            ui.command(name='delete2', label='Delete', icon='Delete'),
            ui.command(name='logout2', label='Logout', icon='Signout', items=[
                ui.command(name='info3', label='About', icon='Info'),
                ui.command(name='delete3', label='Delete', icon='Delete'),
                ui.command(name='logout3', label='Logout', icon='Signout'),
            ])
        ])
    ]
```

Menu
image

Table menu
image

Card menu
image

### Implementation Notes
I had a problem trying to reuse `fixIconScroll` styles for CardMenu. It looks like it changes the execution order of files when I import `fixIconScroll`. Cards registry is not initialized by the time toolbar tries to register itself.
image

I spent some time trying to spot the issue but I don't know if it's worth investing too much time only for this card.

aalencar:
Fixed the issue above, it was caused by circular dependency.

mturoci (approved):
Thanks @aalencar! I added type and moved the styles into existing `utils` file, no need to introduce a new one.

PR #1412, opened 2022-05-11
Fix: Textbox's scrollbar and text overflow when height is set

### LINKED ISSUE #1385: textbox's scrollbar and text overflow when height is set
#### Actual behavior

```python
from h2o_wave import site, ui

page = site['/demo']

page['example'] = ui.form_card(
    box='1 1 4 2',
    items=[ui.textbox(name='textbox_multiline', label='Multiline textarea',
                      value='Test\nTest\nTest\nTest\nTest\n',
                      multiline=True,
                      height='50px'
                      )]
)

page.save()
```
when height= is set, textbox resize makes scrollbar and text overflow.
image

#### Expected behavior

no overflow

marek-mihok:
Fixes [textbox overflow issue](https://github.com/h2oai/wave/issues/1385) when height is set for multiline **ui.textbox**.

Closes #1385

### DIFF
--- a/ui/src/textbox.tsx
+++ b/ui/src/textbox.tsx
@@ -100,7 +100,7 @@ export const
           styles={m.multiline && m.height
             ? {
               field: { height: m.height },
-              fieldGroup: { border: 'none', borderRadius: 0, outline: '1px solid' }
+              fieldGroup: { minHeight: m.height }
             }
             : undefined}
           label={m.label}
--- a/ui/src/textbox.tsx
+++ b/ui/src/textbox.tsx
@@ -97,7 +97,12 @@ export const
       : (
         <Fluent.TextField
           data-test={m.name}
-          styles={m.multiline && m.height ? { fieldGroup: { height: m.height } } : undefined}
+          styles={m.multiline && m.height
+            ? {
+              field: { height: m.height },
+              fieldGroup: { border: 'none', borderRadius: 0, outline: '1px solid' }
+            }
+            : undefined}
           label={m.label}
           placeholder={m.placeholder}
           iconProps={{ iconName: m.icon }}
--- a/ui/src/textbox.tsx
+++ b/ui/src/textbox.tsx
@@ -97,12 +97,7 @@ export const
       : (
         <Fluent.TextField
           data-test={m.name}
-          styles={m.multiline && m.height
-            ? {
-              field: { height: m.height },
-              fieldGroup: { minHeight: m.height }
-            }
-            : undefined}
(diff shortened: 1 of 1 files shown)


mturoci:
@marek-mihok can you explain what the problem was and how does your code fix it?

mturoci:
@marek-mihok can you explain what the problem was and how does your code fix it?

marek-mihok:
@mturoci sure. There are two important css classes related to the problem: `fieldGroup` and its child - the `field` class. 

While the `fieldGroup` is for input's container, the `field` class is for the input itself. Before my fix, when the condition `m.multiline && m.height` was met, the following style was applied to the `fieldGroup` class - `{ height: m.height }`. 

By default, dragging the bottom right corner of the textbox input is tied to the height of the element with the `field` class. Its parent container has height 'auto' and resizes to fit its child. 

The problem with our implementation was that when you dragged down the bottom right corner, only the element with the `field` class was expanding its height, but its parent container with `fieldGroup` class was not resizing since we set it the fixed height. 

To solve this I've decided to set height to the `field` class element instead of its parent container and therefore the overflow problem is fixed (the dragging changes the height value set by us). Also it was necessary to remove border from parent container since it added 2px to the textbox height and replace it with the outline which does not take space. After that both `field` and `fieldGroup` elements have the same height.

ui/src/textbox.tsx
```diff
@@ -97,7 +97,12 @@ export const
       : (
         <Fluent.TextField
           data-test={m.name}
-          styles={m.multiline && m.height ? { fieldGroup: { height: m.height } } : undefined}
+          styles={m.multiline && m.height
+            ? {
+              field: { height: m.height },
+              fieldGroup: { border: 'none', borderRadius: 0, outline: '1px solid' }
```
mturoci, line 103: The default `min-height` is set to `60px` so let's override that as well to give user the behavior according to spec - if height is specified, use that height.
```suggestion
              fieldGroup: { border: 'none', borderRadius: 0, outline: '1px solid', minHeght: m.height }
```

mturoci (changes requested):
Looking good @marek-mihok, just a minor comment.

> Also it was necessary to remove border from parent container since it added 2px to the textbox height and replace it with the outline which does not take space. After that both field and fieldGroup elements have the same height.

The problem with this is that you cannot really recreate the same rounded corners with outline. The extra `2px` is fine on container, we only care about the `field` height, which is correct. Please revert these stylings.

ui/src/textbox.tsx
marek-mihok: Good point @mturoci , thanks.

mturoci (approved):

PR #1413, opened 2022-05-11
Ashrith/wave r examples

ashrith:
New pull.
1. Code clean up
2. Addition of a corner case control handing to core.R
3. Fixes with examples


lo5 (approved):
Thanks @ashrith!

PR #1414, opened 2022-05-12
🐞 fix: use return value of event handler to decide if dispatching sho…

zillionare:
…uld stop

if there's multiple controls in the view, when it's triggered, some textbox may contain value will be treated as True, thus trigger event handler accordingly. These value may not be changed at the moment, thus event handler should just ignore it by return False, however, routing.py currently lack the mech to leverage the return value, and search next event handler.

please refer to https://github.com/h2oai/wave/discussions/1404 for more details

### DIFF
--- a/py/h2o_wave/routing.py
+++ b/py/h2o_wave/routing.py
@@ -110,24 +110,22 @@ def on(arg: str = None, predicate: Optional[Callable] = None):
     return wrap
 
 
-async def _invoke_handler(func: Callable, arity: int, q: Q, arg: any):
+async def _invoke_handler(func: Callable, arity: int, q: Q, arg: any)->bool:
     if arity == 0:
-        await func()
+        return await func()
     elif arity == 1:
-        await func(q)
+        return await func(q)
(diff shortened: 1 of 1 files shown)


Far0n:
I also find the current behavior a bit inconvenient, because of the `Trigger=True` side effects (see also: https://github.com/h2oai/wave/discussions/952).

This PR here by @zillionare looks similar to one of my workarounds, but I have propagation as opt-in rather than opt-out:

```python
async def _match_predicate(predicate: Callable, func: Callable, arity: int, q: Q, arg: any) -> bool:
    if predicate:
        if predicate(arg):
            res = await _invoke_handler(func, arity, q, arg, **params)
            return True and res is not False
    else:
        if arg:
            res = await _invoke_handler(func, arity, q, arg, **params)
            return True and res is not False
    return False
```

Wouldn't that be okay @lo5? That doesn't change behavior of any current code, but would open a litte backdoor to allow invokation of more than one @on-matched handler.

PR #1438, opened 2022-05-23
feat: Image annotator improvements

mturoci:
Feedback from HT team. Changed:

* `ImageAnnotatorRect` takes floats instead of ints for better precision.
* Added a limit of `5px` for minimal boxes as anything less than that is undetectable. This should be a reasonable default and should not be limiting for anyone.
* Handle edge case when a user draws outside of the canvas.
* Stop shapes on the image edge when moving.
* Ignore right clicks.

## API review

Changed ints to floats.

```ts
/** Create a rectangular annotation shape. */
interface ImageAnnotatorRect {
  /** `x` coordinate of the rectangle's corner. */
  x1: F
  /** `y` coordinate of the rectangle's corner. */
  y1: F
  /** `x` coordinate of the diagonally opposite corner. */
  x2: F
  /** `y` coordinate of the diagonally opposite corner. */
  y2: F
}
```

Thanks @gaborfodor @pascal-pfeiffer for awesome feedback!


lo5 (approved):
Looks good. Thanks @mturoci!

PR #1445, opened 2022-05-26
feat: Allow full Vega grammar in vega_card. #1442

### LINKED ISSUE #1442: Support full vega spec, not just lite
via @bender404

Some existing JS apps use full vega spec and would like to migrate to Wave. However, they would like to reuse as much work as possible (existing full vega spec plots).

`vega-embed` supports both modes (`lite` and `vega`) so seems like a single extra attribute should do the job.

cc @lo5

mturoci:
## API addition

```ts
  /** Vega grammar to use. Defaults to 'vega-lite'. */
  grammar?: 'vega-lite' | 'vega'
```

Closes #1442


lo5 (approved):
Thanks @mturoci!

PR #1447, opened 2022-05-27
updated wavegen to generate view=card

ashrith:
Add the ability to generate the variable and parameter `view=card` function to `wavegen.ts`. Now `ui.R` contains `view` and does not need to be generated ad-hoc from core.


lo5 (approved):
Thanks @ashrith!

ashrith:
@lo5 🙏

PR #1454, opened 2022-05-30
Allow parameter matching in arg of decorator on

### LINKED ISSUE #1008: @on pattern matching for query arguments
via @Far0n @vopani @AakashKumarNain @mtanco

It is a common pattern to have a single function to be used for multiple handlers, e.g.

```py
@on('button_1')
@on('button_2')
@on('button_3')
async def render_home(q: Q):
   ...
   < do stuff to render homepage >
   ...

   await q.page.save()
```

Syntax sugar pattern matching

```py
@on('button_{id}')
async def render_home(q: Q, id):
```

cc @lo5

Far0n:
Allows parameter matching of non **#**-route args for the **on**-decorator:

```python
@on('button_{id}')
async def button_clicked(q: Q, id):
```

Predicates and value arguments also work, e.g.
```python
@on('button_{id}', lambda x: True)
async def button_clicked(q: Q, arg_value, id):
```

Type conversion:
```python
@on('button_{id:int}')
async def button_clicked(q: Q, id):
```

Multiple params:
```python
@on('button_{id1}:{id2}')
async def button_clicked(q: Q, id1, id2):
```

- Fixes #1008

### DIFF
--- a/py/h2o_wave/routing.py
+++ b/py/h2o_wave/routing.py
@@ -12,11 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Optional, Callable
+from typing import Optional, Callable, Dict, Pattern
 from inspect import signature
 import asyncio
 import logging
 from starlette.routing import compile_path
+from starlette.convertors import Convertor
 from .core import expando_to_dict
 from .server import Q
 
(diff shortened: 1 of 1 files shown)


lo5:
Thanks for the PR @Far0n! Changes look good. Maybe we can push this out in 0.23? The routing docs need an update as well.

Far0n:
Happy to help with the documentation. Guess starter would be to update the doc-string of `def on`, but what are other places?

Issue #1448, opened 2022-05-27
Make post_logout_redirect_uri configurable


henrycs:
#### Wave SDK Version, OS
h2o-wave, 0.21.1, Ubuntu 20.04, Keycloak 10.0.2 (docker), Chrome 100.0.4896.127

#### Actual behavior
We test OAuth in wave app with Keycloak 10.0.2, it's OK for login testing, but we cannot logout user session using "_auth/logout", Keycloak reports "invalid redirect uri" after we call "_auth/logout", we can see the session still exists in Keycloak Web console.
URL to logout session generated by wave:
http://192.168.100.201:8080/auth/realms/master/protocol/openid-connect/logout?id_token_hint=xxxxxxxxxx&post_logout_redirect_uri=192.168.100.201%3A10101
I try to replace post_logout_redirect_uri with "http://192.168.100.201%3A10101/demo", and execute this call manually in browser, everything is OK, session removed from Keycloak, the browser redirect to login page again.

#### Expected behavior
Session logout and we can login again (including session in Keycloak)

#### Steps To Reproduce
1. start Keycloak instance follow the guide: https://wave.h2o.ai/docs/development#run-keycloak
2. configure wave app:
"H2O_WAVE_OIDC_CLIENT_ID": "wave",
"H2O_WAVE_OIDC_CLIENT_SECRET": "1d43a43c-165a-4e83-bfd1-722dc5b1bf8a",
"H2O_WAVE_OIDC_END_SESSION_URL": "http://192.168.100.201:8080/auth/realms/master/protocol/openid-connect/logout",
"H2O_WAVE_OIDC_PROVIDER_URL": "http://192.168.100.201:8080/auth/realms/master",
"H2O_WAVE_OIDC_REDIRECT_URL": "http://192.168.100.201:10101/_auth/callback"
3. start wave app, open browser, access the URL of our app, login
4. click logout button:  ui.link(label='Logout', path='_auth/logout', target=''),

Log and error messages:
Keycloak console:
09:37:31,990 WARN  [org.keycloak.events] (default task-4) type=LOGOUT_ERROR, realmId=master, clientId=null, userId=null, ipAddress=192.168.100.19, error=invalid_redirect_uri, redirect_uri=192.168.100.201:10101
Message in browser after logout URL called: invalid redirect uri
URL in browser address bar:
[...]

PR #1457, opened 2022-05-31
✨ feat: add post-logout-redirect-url option

henrycs:
fix/feature: https://github.com/h2oai/wave/issues/1448
add oidc-post-logout-redirect-url (server param) and H2O_WAVE_OIDC_POST_LOGOUT_REDIRECT_URL (env var) server config options for the highest flexibility
I have tested this change with Keycloak 10.0.2


henrycs:
name of option is wrong

PR #1486, opened 2022-06-06
Fix: Ranged values in plot tooltip are not shown properly

### LINKED ISSUE #1473: Histogram tooltip issue
#### Wave SDK Version, OS
h2o_wave 0.22.0

#### Actual behavior
The `low - high` value is the two number concatenated together which seems strange
image

#### Steps To Reproduce

```python
from h2o_wave import main, app, Q, ui, data


@app('/')
async def serve(q: Q):
    q.page['example'] = ui.plot_card(
        box='1 1 7 5',
        title='Interval',
        data=data('price low high', 8, rows=[
            (4, 50, 100),
            (6, 100, 150),
            (8, 150, 200),
            (16, 350, 400),
            (18, 400, 450),
            (10, 200, 250),
            (12, 250, 300),
            (14, 300, 350),
        ]),
        plot=ui.plot([ui.mark(
            type='interval',
            y='=price', y_min=0,
            x1='=low', x2='=high',
        )])
    )
    
    await q.page.save()
```

marek-mihok:
When tooltip item value is an array of values these are separated by hyphens now.

image

Closes #1473

### DIFF
--- a/ui/src/plot.tsx
+++ b/ui/src/plot.tsx
@@ -1038,7 +1038,12 @@ const PlotTooltip = ({ items }: { items: TooltipItem[] }) =>
           <span style={{ backgroundColor: mappingData?.color || color }} className="g2-tooltip-marker" />
           <span style={{ display: 'inline-flex', flex: 1, justifyContent: 'space-between' }}>
             <span style={{ marginRight: 16 }}>{item}:</span>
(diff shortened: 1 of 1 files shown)


mturoci (approved):
Thanks @marek-mihok!

PR #1504, opened 2022-06-16
🐞 fix: using r.Host if post_logout_redirect_url not set

henrycs:
I found the new feature not worked in my project, after checking code, I fixed this issue, would you please help me review the change? thanks!


mturoci (approved):
Thanks for catching this @henrycs!

PR #1516, opened 2022-06-24
Fix: Prevent opening all folders on folder/file create

marek-mihok:


studio/studio.js
```diff
@@ -125,10 +127,6 @@ const Tree = {
       this.folder.action = null
       window.parent.wave.emit('file_viewer', this.folder.isFolder ? 'new_folder' : 'new_file', { path: this.folder.path, name: this.folder.label })
     },
-    rename() {
-      this.folder.action = 'rename'
-      this.folder._labelCache = this.folder.label
-    },
```
mturoci, line 131: Why was this removed?

mturoci (changes requested):
The fix looks good.

> Using v-show instead would preserve the state of all folders and it's sub-folders (like in VS Code).

Yes, definitely go for it. Also, let's open just the root folder on page load instead of opening them all.

studio/studio.js
marek-mihok: It's just a leftover I forgot to remove while removing double click to rename feature implementation.

mturoci (approved):

PR #1517, opened 2022-06-24
Feat: Do not apply card margin when card height/width is 0px

marek-mihok:
Allows to completely hide card. Useful for apps where it is necessary to hide/unhide cards.

### DIFF
--- a/ui/src/page.tsx
+++ b/ui/src/page.tsx
@@ -164,6 +164,7 @@ const
       if (grow) style.flexGrow = grow // grow only if non-zero (else default to shrink)
       if (width) style.width = width
       if (height) style.height = height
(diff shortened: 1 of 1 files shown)


marek-mihok:
@mturoci do you think we need this behavior for other units like `%, rem,...`?

mturoci:
There are 2 options we can approach this:

1. Already implemented in this PR - remove margins in case of `0px`. I guess there is no need to support other units, zero is still zero.
2. Introduce a new card attr `hidden` defaulting to `False`.

cc @lo5 for discussion. For context, this feature is needed when one wants to hide the card but keep its state (`del q.page['card']` does not preserve the state).

lo5 (approved):

PR #1518, opened 2022-06-24
Fix: Unable to control resizing of textbox when height is set in %

### LINKED ISSUE #1490: Textbox resizing issue(s)
#### Wave SDK Version, OS
v0.22.0 on MacOS

#### Actual behaviour
Textbox resizing is buggy.

https://user-images.githubusercontent.com/8954217/172659025-6f4e690c-2b2f-4593-9295-75b3bef18385.mov

#### Expected behaviour
Textbox resizing should be smooth.

#### Steps To Reproduce
```python


@app('/')
async def serve(q: Q):
    """
    App function.
    """

    q.page['card'] = ui.form_card(
        box='1 1 4 4',
        items=[ui.textbox(name='t', label='Text', value='Sample', multiline=True, height='100%', width='100%')]
    )

    await q.page.save()
```

#### Comments
1 - What does height=100% mean or do? Even when I use long text, the box doesn't fit the text and height remains same.
2 - When setting height=100%, when dragged to increase height, it is impossible to drag it back to reduce size.
3 - When setting height=100%, the textbox continues increasing in height after clicking even though the mouse doesn't move.

marek-mihok:
Also fixes additional 40px padding ([declared here](https://github.com/microsoft/fluentui/blob/c5c4ca05e04672bb488fa22dbb75cbde67678321/packages/react/src/components/TextField/TextField.styles.tsx#L347)) on textfield even when icon is not set. 

Closes #1490

### DIFF
--- a/py/h2o_wave/types.py
+++ b/py/h2o_wave/types.py
@@ -1042,7 +1042,7 @@ class Textbox:
         self.trigger = trigger
         """True if the form should be submitted when the text value changes."""
         self.height = height
-        """The height of the text box, e.g. '100px'. Applicable only if `multiline` is true."""
+        """The height of the text box, e.g. '100px'. Percentage values not supported. Applicable only if `multiline` is true."""
         self.width = width
         """The width of the text box, e.g. '100px'. Defaults to '100%'."""
         self.visible = visible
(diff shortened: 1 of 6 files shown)


mturoci (dismissed):
Thanks @marek-mihok. LGTM, just a single comment.

mturoci (approved):

PR #1539, opened 2022-07-13
fix: Add icon to masked textfield component #1536

### LINKED ISSUE #1536: Icon Bug For Textbox Component When Setting Mask Parameter
#### Wave SDK Version, OS

Ecnountered on Wave 0.22.0, Mac OSX 11.3.1

#### Actual behavior

Adding a mask to a textbox component disables the ability to specify an icon.

#### Expected behavior

Be able to add both a mask and icon to a textbox component.

aalencar:
closes #1536

### DIFF
--- a/ui/src/textbox.tsx
+++ b/ui/src/textbox.tsx
@@ -70,12 +70,30 @@ export interface Textbox {
 const DEBOUNCE_TIMEOUT = 500
 export const
   XTextbox = ({ model: m }: { model: Textbox }) => {
-    const onChange = ({ target }: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>, v?: S) => {
-      v = v || (target as HTMLInputElement).value
+    const
+      onChange = ({ target }: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>, v?: S) => {
+        v = v || (target as HTMLInputElement).value
 
-      wave.args[m.name] = v ?? (m.value || '')
-      if (m.trigger) wave.push()
-    }
+        wave.args[m.name] = v ?? (m.value || '')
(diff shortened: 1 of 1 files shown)


mturoci (commented):
Thanks @aalencar. Seems like there are more missing props though. Let's create a single `Fluent.ITextFieldProps` that will be passed into both `MaskedTextField` and `TextField` to have this resolved once and for all.

ui/src/textbox.tsx
```diff
@@ -70,12 +70,30 @@ export interface Textbox {
...
+      onChange = ({ target }: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>, v?: S) => {
+        v = v || (target as HTMLInputElement).value
 
-      wave.args[m.name] = v ?? (m.value || '')
-      if (m.trigger) wave.push()
-    }
+        wave.args[m.name] = v ?? (m.value || '')
+        if (m.trigger) wave.push()
+      },
+      textFieldProps: Fluent.ITextFieldProps = {
```
mturoci, line 80: Can we also add `data-test` here?

mturoci (changes requested):
Can we also include `data-test` within `textFieldProps`?

aalencar:
> Can we also include `data-test` within `textFieldProps`?

Let's use `textFieldProps` only for `Fluent.ITextFieldProps` properties.

mturoci:
>Let's use textFieldProps only for Fluent.ITextFieldProps properties.

Why?

mturoci:
That's fine. The goal here is to have a single source of truth for common props.

mturoci (approved):

PR #1555, opened 2022-07-25
feat: Add explicit refresh token callable. #1230

### LINKED ISSUE #1230: Provide a way to get fresh access tokens even "in request"

Access token are usually short-lived. In order or minutes but even sub minute time to live is possible.
App can use `access_token` property of available `h2o_wave.Q.auth`. This token seems to guaranteed to be fresh on the invocation of the handler but can expire during the handling when the handling takes longer the the token's TTL. This can happen when (e.g.) remote service takes too long to respond or big data are being process within the handler.

`h2o_wave.Q.auth` exposes also `refresh_token` but no means for using it for obtaining a new access token.

Extend `Auth` class in the so that the `access_token` is fact `@property` that ensures that the access token is fresh even withing the request handling. 

Or...

provide the callable that ensures the same.

Or...

provide the callable that forces the refresh of the access token.

Exposing OIDC client details so that the app can do the `refresh_token` exchange by its own.

This would expose client secret to the app and that's not desirable.

#### Additional context

N/A

mturoci:
Adds `force_token_refresh` callable to `q.auth`. Upon calling, the `q.auth.access_token` is updated, and waved session is synced.

Since this is server-to-server communication, no cookies are stored (in comparison to server-to-browser) so I had to keep the session ID in the app to make sure I could identify and sync the correct session during refresh. Open to better ideas/suggestions.

Open question: Shall we purge the session in case of an unsuccessful refresh attempt?

Closes #1230

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -494,6 +494,8 @@ func (h *RefreshHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	session.token = token
+	h.auth.set(session)
+
 	w.WriteHeader(http.StatusOK)
 	w.Header().Set("Wave-Access-Token", token.AccessToken)
 	w.Header().Set("Wave-Refresh-Token", token.RefreshToken)
--- a/auth.go
+++ b/auth.go
@@ -489,7 +489,7 @@ func (h *RefreshHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		// Purge session and reload clients if refresh not successful?
 		echo(Log{"t": "refresh_session", "error": err.Error()})
-		http.Error(w, http.StatusText(http.StatusUnauthorized), http.StatusUnauthorized)
+		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
 		return
 	}
 
--- a/auth.go
+++ b/auth.go
@@ -487,6 +487,7 @@ func (h *RefreshHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	token, err := h.auth.ensureValidOAuth2Token(r.Context(), session.token)
 	if err != nil {
+		// Purge session and reload clients if refresh not successful?
 		echo(Log{"t": "refresh_session", "error": err.Error()})
 		http.Error(w, http.StatusText(http.StatusUnauthorized), http.StatusUnauthorized)
 		return
--- a/auth.go
+++ b/auth.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/coreos/go-oidc"
 	"github.com/google/uuid"
+	"github.com/h2oai/wave/pkg/keychain"
 	"golang.org/x/oauth2"
 )
 
@@ -458,4 +459,41 @@ func (h *LogoutHandler) redirect(w http.ResponseWriter, r *http.Request, idToken
 	redirectURL.RawQuery = query.Encode()
 
 	http.Redirect(w, r, redirectURL.String(), http.StatusFound)
+}
+
+// Handles token refreshes.
+type RefreshHandler struct {
+	auth     *Auth
+	keychain *keychain.Keychain
+}
+
+func newRefreshHandler(auth *Auth, keychain *keychain.Keychain) http.Handler {
+	return &RefreshHandler{auth, keychain}
+}
+
+func (h *RefreshHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if !h.keychain.Guard(w, r) { // API
+		return
+	}
+
+	sessionID := r.Header.Get("Wave-Session-ID")
+	session, ok := h.auth.get(sessionID)
+
+	if !ok {
+		echo(Log{"t": "refresh_session", "error": "session unavailable"})
+		http.Error(w, http.StatusText(http.StatusUnauthorized), http.StatusUnauthorized)
+		return
+	}
+
+	token, err := h.auth.ensureValidOAuth2Token(r.Context(), session.token)
+	if err != nil {
+		echo(Log{"t": "refresh_session", "error": err.Error()})
+		http.Error(w, http.StatusText(http.StatusUnauthorized), http.StatusUnauthorized)
+		return
+	}
+
+	session.token = token
+	w.WriteHeader(http.StatusOK)
+	w.Header().Set("Wave-Access-Token", token.AccessToken)
+	w.Header().Set("Wave-Refresh-Token", token.RefreshToken)
 }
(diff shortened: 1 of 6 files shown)


py/h2o_wave/server.py
```diff
@@ -73,6 +73,21 @@ def __init__(self, username: str, subject: str, access_token: str, refresh_token
         """The access token of the user."""
         self.refresh_token = refresh_token
         """The refresh token of the user."""
+        self._session_id = session_id
+        """Session identifier. Do not access, internal use only."""
```
zoido, line 77: nit: I don't know how the doc generation works so I just mention that this is internal attribute  so it should not (IMHO) appear in the documentation.

py/h2o_wave/server.py
```diff
@@ -73,6 +73,21 @@ def __init__(self, username: str, subject: str, access_token: str, refresh_token
...
+    
+    async def force_token_refresh(self):
+        """
+        Implictly refresh OIDC tokens when needed, e.g. during long-running background jobs.
```
zoido, line 81: Explicitly IMHO.

Implicit is something that is implied, explicit is something that us stated directly.

We need to call the endpoint directly, therefore I think the refresh is done "explicitly".

auth.go
```diff
@@ -459,3 +460,41 @@ func (h *LogoutHandler) redirect(w http.ResponseWriter, r *http.Request, idToken
...
+
+	token, err := h.auth.ensureValidOAuth2Token(r.Context(), session.token)
+	if err != nil {
+		// Purge session and reload clients if refresh not successful?
+		echo(Log{"t": "refresh_session", "error": err.Error()})
+		http.Error(w, http.StatusText(http.StatusUnauthorized), http.StatusUnauthorized)
+		return
+	}
+
+	session.token = token
```
zoido, line 496: I might be wrong but I'd expect to call a `h.auth.set` after changing the session after modifying it. Is it missing here?

py/h2o_wave/server.py
```diff
@@ -73,6 +73,21 @@ def __init__(self, username: str, subject: str, access_token: str, refresh_token
         """The access token of the user."""
         self.refresh_token = refresh_token
         """The refresh token of the user."""
+        self._session_id = session_id
+        """Session identifier. Do not access, internal use only."""
+    
+    async def force_token_refresh(self):
```
zoido, line 79: Plea: can we also return the obtained `access_token`? It would make this method usable as the async token provider used by client libraries.

e.g.:
```python
from typing_extensions import Protocol

class TokenProvider(Protocol):
    async def __call__(self) -> str:
        pass
```

zoido, line 79: Naming: IMHO this does necessarily not _forces_ the refresh of the token. I cannot confirm what implementation of the `TokenSource` is used server-side but default implementation IMHO refreshes the token only if it's expired. Prefixing it with `force_` might implied functionality that is not effectively implemented.

Ideas:
 - `ensure_fresh_token`: reflects what it really does on the back end
 - `token_provider`: see the plea comment

zoido (commented):
Thank you. Just some minor comments with discussion potential and one question:

What are potential performance implications?

Apps that will required this will need to call this with every outgoing request. This is IMHO ok for the internal channel to `waved`. But I am not sure what `TokenSource` implementation is effectively used so my concern is whether there is a potential of spamming the identity with a lot of token requests.

py/h2o_wave/server.py
mturoci: These are ignored during docs build. The docstring is for people that like to explore code manually, in which case I guess it's fine.

py/h2o_wave/server.py
mturoci: Agree.

auth.go
mturoci: Good point, done.

py/h2o_wave/server.py
mturoci: I see no problem with that. How shall it be returned? In header or body?

zoido: I meant just `return access_token` at the end of the method.

mturoci: I agree with `force_` prefix not being the best. I would go with `ensure_fresh_token` as it seems to be most flexible for both using it as explicit refresh and as a token provider, wdyt?

zoido: Agree. Thanks.

mturoci:
>But I am not sure what TokenSource implementation is effectively used so my concern is whether there is a potential of spamming the identity with a lot of token requests.

Good point. According to [docs](https://pkg.go.dev/golang.org/x/oauth2#section-readme) - _TokenSource returns a TokenSource that returns t until t expires, automatically **refreshing it as necessary** using the provided context._ which seems good to me, but feel free to correct me if I am missing something.

> Open question: Shall we purge the session in case of an unsuccessful refresh attempt?

wdyt @zoido?

auth.go
```diff
@@ -459,3 +460,43 @@ func (h *LogoutHandler) redirect(w http.ResponseWriter, r *http.Request, idToken
...
+		echo(Log{"t": "refresh_session", "error": "session unavailable"})
+		http.Error(w, http.StatusText(http.StatusUnauthorized), http.StatusUnauthorized)
+		return
+	}
+
+	token, err := h.auth.ensureValidOAuth2Token(r.Context(), session.token)
+	if err != nil {
+		// Purge session and reload clients if refresh not successful?
+		echo(Log{"t": "refresh_session", "error": err.Error()})
+		http.Error(w, http.StatusText(http.StatusUnauthorized), http.StatusUnauthorized)
```
zoido, line 492: Is this always unauthorized?

zoido:
> Good point. According to [docs](https://pkg.go.dev/golang.org/x/oauth2#section-readme) - _TokenSource returns a TokenSource that returns t until t expires, automatically **refreshing it as necessary** using the provided context._ which seems good to me, but feel free to correct me if I am missing something.

Agree. Clicked trough it once again and got to the same conclusion. Thanks.

> 
> > Open question: Shall we purge the session in case of an unsuccessful refresh attempt?
> 
> wdyt @zoido?

I don't think it's necessary for the explicit endpoint. Reasoning: you explicitly call something that can fail. If it fails you should be able to recover. 

But I do not see deep into the interaction between the app and waved so I cannot foresee all of the potential side-effects.

zoido (approved):
LGTM 🥗

Thanks.

auth.go
mturoci: I believe most of the time, but the AP server can also be down etc. Other options could be:
* propagating the err - probably not ideal security-wise.
* returning generic 500 server error.

zoido: I vote for default on 500 unless there is a way to distinguish the cause.

mturoci:
Thanks for the review @zoido!

lo5 (approved):
Thank you @mturoci, @zoido!

zoido (approved):

PR #1566, opened 2022-08-01
fix: Prevent ui.wide_gauge_stat from moving when textual value changes

### LINKED ISSUE #1446: Wide gauge stat should not move when textual values change
via @itsxaos
### Discussed in https://github.com/h2oai/wave/discussions/1443

<sup>Originally posted by **itsxaos** May 25, 2022</sup>
The alignment of the gauge on the wide gauge card doesn't seem right, it moves when the value / auxiliary value changes, and generally doesn't look like it does in the [docs](https://wave.h2o.ai/docs/examples/stat-wide-gauge).

Here's what it looks like, on the left in Vivaldi (Chromium) and Firefox on the right:

[video](https://user-images.githubusercontent.com/33079230/170277424-57ecad0d-1f11-41f3-b0a8-8c1ccf3e9483.mp4)

## Implementation notes

The moving is kind of expected when one moves between order of 10s (e.g. going from 99 to 100) or the precision changes (e.g. 0,1 to 0,09). However, this should not happen when the change is within the same order (1 to 2).

marek-mihok:
**Before:**

https://user-images.githubusercontent.com/23740173/182230440-8ab0174f-2522-4dca-aa1b-1fe1d0302bea.mov

**After:**

https://user-images.githubusercontent.com/23740173/183668215-859a1a54-06bf-4269-9c7e-fc7c98f9ee84.mov

Closes #1446


mturoci (changes requested):
Thanks @marek-mihok. Let's change the percentage ratio to a precomputed px min width instead so that the stat is somewhat responsive. Currently, the circle is way too small even if there is a lot of free space. Instead, we would want the circle to take up all the free space and keep the minimal space needed for the value part.

Original:
image

This PR:
image

ui/src/wide_gauge_stat.tsx
```diff
@@ -76,8 +73,19 @@ interface State {
 
 export const
   View = bond(({ name, state: s, changed }: Model<State>) => {
+    const
+      valueNonDigitCharCount = substitute(s.value, s.data).replace(/[0-9]/g, '').length,
+      auxValueNonDigitCharCount = substitute(s.aux_value, s.data).replace(/[0-9]/g, '').length
```
mturoci, line 78: Is this really ***Non*** digit count? Seems like the regex counts the digits or maybe I am wrong?

ui/src/wide_gauge_stat.tsx
```diff
@@ -76,8 +73,19 @@ interface State {
...
-      const data = unpack<Rec>(s.data)
+      const
+        data = unpack<Rec>(s.data),
+        /*
+         * Numbers 14 and 9 are representing maximum width of the single character.
+         * The minimum width of the value containers is the width of 4 characters.
+         * This prevents the jumping layout in the most common scenarios when numbers are changing between ones to tens. 
+        */
+        valueContainerWidth = 14 * Math.max(substitute(s.value, s.data).length, 4 + valueNonDigitCharCount),
+        auxValueContainerWidth = 9 * Math.max(substitute(s.aux_value, s.data).length, 4 + auxValueNonDigitCharCount)
```
mturoci, line 88: Let's substitute only once, put the values into variables and use the variables for both digitcounts and containerwidths.

ui/src/wide_gauge_stat.tsx
```diff
@@ -76,8 +73,19 @@ interface State {
...
   View = bond(({ name, state: s, changed }: Model<State>) => {
+    const
+      valueNonDigitCharCount = substitute(s.value, s.data).replace(/[0-9]/g, '').length,
+      auxValueNonDigitCharCount = substitute(s.aux_value, s.data).replace(/[0-9]/g, '').length
     const render = () => {
-      const data = unpack<Rec>(s.data)
+      const
+        data = unpack<Rec>(s.data),
+        /*
+         * Numbers 14 and 9 are representing maximum width of the single character.
```
mturoci, line 83: use properly named constants instead of this comment. Something like `MAX_AUX_WIDTH` etc.

mturoci (changes requested):
Thanks @marek-mihok last small comments.

ui/src/wide_gauge_stat.tsx
marek-mihok: Yes, it really is. `replace` function replaces digits matched by regex with empty string `''` so the length of the resulting string is the count of non-digit characters.

mturoci: Could we use a negative regex (all non-numerical characters) `match` instead of `replace`? Would be less mental gymnastics that way

ui/src/wide_gauge_stat.tsx
marek-mihok: The problem is that in that case digitcounts, especially `.replace(/[0-9]/g, '').length` part will be computed on every render what is not necessary. Are you sure you want it like this?

ui/src/wide_gauge_stat.tsx
marek-mihok: Yeah, I was deciding about this option as well. I'll do it!

ui/src/wide_gauge_stat.tsx
marek-mihok: Sure. That's a good idea. What I was trying to achieve was to prevent something like this: `valueNonDigitCharCount = substitute(s.value, s.data).length - substitute(s.value, s.data).replace(/[^0-9]/g, '').length

mturoci: Yes, but match solves this.

ui/src/wide_gauge_stat.tsx
mturoci: Sure. Wouldn't doing otherwise cause a bug? After updating the value, the digit counts would not be updated thus be out of date.

marek-mihok: Nope, because `valueNonDigitCharCount` and `auxValueNonDigitCharCount` says about the number of non-digit characters which is always the same. Let's say you have value defined in this format: `value='=${{intl foo minimum_fraction_digits=2 maximum_fraction_digits=2}} (USD)'`
So the real example would be: `$43.02 (USD)`
That means you have 8 non-digits character here: `$.[space](USD)`
Even if value changes to `$143.02 (USD)`or `$3.12 (USD)` the number of non-digit characters does not change. Therefore it is fine to have it computed just once.

mturoci: What about changing the value from `1 share` to `2 shares`?

mturoci (approved):
Looking great, thanks @marek-mihok!

PR #1568, opened 2022-08-02
feat: Do not freeze Wave dependencies version.

### LINKED ISSUE #1460: ``h2o_wave` Python package should not pin `httpx` version.

Cannot use (at least) semi-arbitrary versions of `httpx` (and effectively also `starlette` and `uvicorn`) and dependencies that has require different (even slightly) `httpx`.
Installation via `requirements.txt` leads to the error.

```
httpx==0.22
h2o_wave==0.21
```

```
The conflict is caused by:
    The user requested httpx==0.22
    h2o-wave 0.21.0 depends on httpx==0.16.1
```

Don't pin `httpx` version or at least allow range of versions.
I understand the motivation behind the pining. However the part of the `h2o_wave` is used as library/framework and pining dependencies is not considered as a good practice for libraries.

mturoci:
This PR sets the minimum required version for Wave python dependencies. Both unit and manual testing pass with the latest versions.

This PR also adds python unit tests to be run as part of our CI releases. This way we get notified if something breaks due to a new dependency version thanks to our nightly job that runs every day.

I believe this is the most flexible (and automated) way for dealing with this since it allows for using basically any dependency version, thus solving #1460. Another option can be to allow a version range that we guarantee Wave can handle, but it's more maintenance-demanding compared to the proposed one.

Open to other suggestions @lo5

Closes #1460


lo5 (approved):
Thanks @mturoci - looks great!

PR #1590, opened 2022-08-17
Fix: Prevent regression caused by non-interactive plots

### LINKED ISSUE #1586: Regression caused by non-interactive plots
Regression introduced by #1331.

Repro:

```py
from h2o_wave import main, data, ui, site

page = site['/demo']

rows = [["2022-08-10 08:20:09", 0.0], ["2022-08-10 08:25:09", 0.63], ["2022-08-10 08:50:09", 0.88]]

marks = [
    ui.mark(
        type="line",
        x="=time",
        y="=feature_drift",
        color="#FEC925",
        x_title="x_title",
        y_title="y_title",
        y_min=0,
        label_overlap="hide",
    ),
    ui.mark(
        y=0.2,
        label=str(0.2),
        label_fill_color="#F44336",
        ref_stroke_color="#F44336",
    )
]

page.add('example', ui.form_card(
    box='1 1 4 4', items=[
        ui.visualization(
            plot=ui.plot(marks),
            data=data("time feature_drift", rows=rows, pack=True),
            interactions=["brush"],
        )
    ]
))

page.save()
```

Rendering blank page with a console error:

image

The problem is at `plot.tsx:1111`.

marek-mihok:
Filters out annotation marks before setting geometry interactivity to prevent regression.

Closes #1586

### DIFF
--- a/ui/src/plot.tsx
+++ b/ui/src/plot.tsx
@@ -1112,7 +1112,10 @@ export const
           })
           currentChart.current = chart
           chart.data(data)
-          model.plot.marks.forEach(({ interactive = true }, idx) => chart.geometries[idx].customInfo({ interactive }))
+          // filter out annotation marks
+          model.plot.marks.filter(({ x, y }) => x == null && y == null).forEach(({ interactive = true }, idx) =>
+            chart.geometries[idx].customInfo({ interactive })
+          )
           if (model.events) {
             for (const event of model.events) {
               switch (event) {


mturoci:
@marek-mihok can you elaborate on what caused the regression and why you think this is the correct fix?

marek-mihok:
@mturoci sure. `marks` defined by user are transformed into `geometries` and `annotations` arrays by our `makeGeom` function:
```
for (const m of marks) {
  const { x, y } = m
  if (x != null || y != null) {
    annotations.push(...makeNote(m))
  } else {
    geometries.push(makeGeom(m))
  }
}
```
These arrays are later passed to the chart `options` when creating a chart.
So assuming that for every `mark` there is exactly one `geometry` object is wrong since `mark` can be either `geometry` or an `annotation` where geometries are stored in `chart.geometries` and annotations in `chart.options.annotations`:
```
model.plot.marks.forEach(({ interactive = true }, idx) =>
  chart.geometries[idx].customInfo({ interactive })
)
```
When talking about interactivity, this is the property of chart geometries, not chart annotations.
Therefore the solution here is to filter out marks which are not geometries before:
```
model.plot.marks.filter(({ x, y }) => x == null && y == null).forEach(({ interactive = true }, idx) =>
  chart.geometries[idx].customInfo({ interactive })
)
```

mturoci:
Got it, makes sense.

ui/src/plot.tsx
```diff
@@ -1112,7 +1112,10 @@ export const
           })
           currentChart.current = chart
           chart.data(data)
-          model.plot.marks.forEach(({ interactive = true }, idx) => chart.geometries[idx].customInfo({ interactive }))
+          // filter out annotation marks
+          model.plot.marks.filter(({ x, y }) => x == null && y == null).forEach(({ interactive = true }, idx) =>
+            chart.geometries[idx].customInfo({ interactive })
+          )
```
mturoci, line 1118: It seems like this is logically correct, but took me a while to wrap my head around whether the indexes are indeed correct. Let's go for something like

```suggestion
          model.plot.marks.forEach(({ x, y, interactive = true }, idx) => {
            if(isGeometry(x,y)) chart.geometries[idx].customInfo({ interactive })
          })
```
With a helper function (with proper types)
```ts
const isGeometry = (x,y) => x == null || y === null
```

And also refactor the existing `makeMarks` function to use the helper one so that the it's clear what's going on.

mturoci (changes requested):

ui/src/plot.tsx
marek-mihok: No problem. I totally understand it. My original implementation was almost the same when I've realised the indexes are wrong. So I came up with the filter solution to keep indexes correct.

PR #1591, opened 2022-08-17
fix: Prevent memory leak caused by unclosed file handles during site.upload.

mturoci:
Prevents a memory leak.

HTTPX (our HTTP client) lib accepts `BufferedReader` for multipart (file) uploads, but somehow doesn't close file handles when finished so it's necessary to do it ourselves - this PR.


lo5 (approved):
Good catch! Thanks @mturoci!

PR #1593, opened 2022-08-17
Fix: Prevent ui.wide_series_stat moving when textual values change

### LINKED ISSUE #1592: Wide series stat should not move when textual values change
This is the same issue as #1446 but for `ui.wide_series_stat`:

https://user-images.githubusercontent.com/23740173/185189539-505274cf-ab2b-4091-a554-0317bae05cfb.mov

### Implementation notes
Provide similar fix for `ui.wide_series_stat` as the fix provided for `ui.wide_gauge_stat` in #1566 and prevent the value moving when the value change is within the same numeric order.

marek-mihok:
This PR makes the behavior consistent with `ui.wide_gauge_stat` (#1566) and prevents disturbing and unnecessary moving of  numeric/textual value when it changes.

https://user-images.githubusercontent.com/23740173/191020785-bca76f88-d989-4eed-b13b-673bd614c3e1.mov

Closes #1592

### DIFF
--- a/ui/src/wide_gauge_stat.tsx
+++ b/ui/src/wide_gauge_stat.tsx
@@ -83,8 +83,8 @@ export const
         data = unpack<Rec>(s.data),
         value = substitute(s.value, s.data),
         auxValue = substitute(s.aux_value, s.data),
-        valueNonDigitCharCount = value.match(/[^0-9]/g).length,
-        auxValueNonDigitCharCount = auxValue.match(/[^0-9]/g).length,
+        valueNonDigitCharCount = value.match(/[^0-9-]/g).length,
+        auxValueNonDigitCharCount = auxValue.match(/[^0-9-]/g).length,
         // This prevents the jumping layout in the most common scenarios when numbers are changing between ones to tens. 
         valueContainerWidth = MAX_CHAR_WIDTH * Math.max(value.length, MIN_CHAR_COUNT + valueNonDigitCharCount),
         auxValueContainerWidth = MAX_CHAR_AUX_WIDTH * Math.max(auxValue.length, MIN_CHAR_COUNT + auxValueNonDigitCharCount)
--- a/ui/src/wide_series_stat.tsx
+++ b/ui/src/wide_series_stat.tsx
@@ -15,10 +15,10 @@
 import { Data, F, Model, Rec, S, unpack } from 'h2o-wave'
 import React from 'react'
 import { stylesheet } from 'typestyle'
-import { cards, Format, grid } from './layout'
+import { cards, Format, grid, substitute } from './layout'
 import { MicroArea } from './parts/microarea'
 import { MicroBars } from './parts/microbars'
-import { clas, cssVar, pc } from './theme'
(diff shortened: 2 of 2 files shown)


ui/src/wide_gauge_stat.tsx
```diff
@@ -83,8 +83,8 @@ export const
         data = unpack<Rec>(s.data),
         value = substitute(s.value, s.data),
         auxValue = substitute(s.aux_value, s.data),
-        valueNonDigitCharCount = value.match(/[^0-9]/g).length,
-        auxValueNonDigitCharCount = auxValue.match(/[^0-9]/g).length,
+        valueNonDigitCharCount = value.match(/[^0-9-]/g).length,
+        auxValueNonDigitCharCount = auxValue.match(/[^0-9-]/g).length,
```
mturoci, line 87: Why is this needed?

ui/src/wide_series_stat.tsx
```diff
@@ -28,12 +28,12 @@ const
       padding: grid.gap,
     },
     plot: {
-      display: 'flex',
-      width: pc(50),
+      position: 'relative',
```
mturoci, line 31: What is this position for? IIRC plots are not positioned absolutely or maybe I am missing something?

ui/src/wide_series_stat.tsx
```diff
@@ -28,12 +28,12 @@ const
...
+      width: '100%',
     },
     text: {
-      width: pc(50),
       paddingLeft: grid.gap,
+      alignSelf: 'center',
```
mturoci, line 36: I am thinking that centering may not be the best idea here. Wdyt?

image

vs 

image

mturoci (changes requested):
Thanks @marek-mihok. Couple of small questions from me.

ui/src/wide_gauge_stat.tsx
marek-mihok: The intention was to make wide gauge stat implementation consistent with wide series stat and to prevent value moving with change from positive to negative (-) values. Or do you ask whether it makes sense for a gauge stat since it can't display minus values, e.g. a percentage loss?

mturoci: I wasn't sure what it has to do with `NonDigitCharCount` but now I see it's for the `-` sign 👍

ui/src/wide_series_stat.tsx
marek-mihok: I've just added it for implementation consistency with wide gauge stat, but now I see this was not necessary.

ui/src/wide_series_stat.tsx
marek-mihok: Why not? It fills the available space and also it is consistent with a gauge stat. Do you find the first one more elegant or is there any issue with that?

mturoci: No issue, just cosmetics. Let's leave it as is. We can revisit in future if needed.

mturoci (approved):
Thanks @marek-mihok!

PR #1609, opened 2022-08-30
Feat: ui.time_picker component

marek-mihok:
This PR introduces a new `ui.time_picker` component with the following API:
```ts
export interface TimePicker {
  name: Id
  /** Text to be displayed alongside the component. */
  label?: S
  /** A string that provides a brief hint to the user as to what kind of information is expected in the field. */
  placeholder?: S
  /** The time value in hh:mm format. E.g. '10:30', '14:25', '23:59', '00:00' */
  value?: S
  /** True if this field is disabled. */
  disabled?: B
  /** The width of the time picker, e.g. '100px'. Defaults to '100%'. */
  width?: S
  visible?: B
  /** True if the form should be submitted when the time is selected. */
  trigger?: B
  /** True if this is a required field. Defaults to False. */
  required?: B
  /** Specifies 12-hour or 24-hour time format. One of `12` or `24`. Defaults to `24`. */
  hour_format?: S
  /** The minimum allowed time value in hh:mm format. E.g.: '08:00', '13:30' */
  min?: S
  /** The maximum allowed time value in hh:mm format. E.g.: '15:30', '00:00' */
  max?: S
  /** Limits the available minutes to select from. One of `1`, `5`, `10`, `15`, `20`, `30` or `60`. Defaults to `1`. */
  minutes_step?: U
}
```

https://user-images.githubusercontent.com/23740173/189341542-910f1376-5be7-42ab-a5d9-664c0dd9df16.mov


lo5:
The part I don't like is that AM/PM is part of the API, and that `time_format_12h` affects the return value, not just the display of hours.

I would keep it simple and restrict `value`, `min`, `max` and the return value consistently to 24-hr `hh:mm` time format, similar to how the date picker restricts it to the ISO 8601 date format. This will simplify the front-end code (no AM/PM parsing), and is also one less thing to reason about when processing the submitted date in Python. 

Notice that the examples in the PR are confusing at the moment - should there be a space before AM/PM, or not? The examples / docs use both.

The `time_format_12h` can be renamed to `hour_cycle` (one of `12` default, or `24`), similar to the naming used by `Intl.DateTimeFormat`. It should _only_ affect the hour display, not the return value. For example, allow the user to pick `6:45` and `PM`, but submit `18:45`.

marek-mihok:
Thanks @lo5 for making this simple and consistent! Just one note - the plan was to have hour cycle defined as you proposed but then there is an issue after generating a `types.py` where 
```
class TimePickerHourCycle:
    12 = '12'
    24 = '24'
```
which is not a valid python code since `Expression cannot be assignment target`.
Is it worth handling in `wavegen.ts` or can we stick with an already implemented `time_format_12h` boolean?

lo5:
> Is it worth handling in wavegen.ts or can we stick with an already implemented time_format_12h boolean?

Define `hour_cycle` as `string` instead of `'12' | '24'`. I don't like the boolean `time_format_12h` because `11` and `23` are valid hour cycles (we don't have to support them today, but someone might bring this up in the future).

ui/src/time_picker.tsx
```diff
@@ -0,0 +1,269 @@
...
+      {label && <Fluent.Label className={css.toolbarLabel}>{label}</Fluent.Label>}
+      <Fluent.Text className={css.toolbarText}>
+        <Fluent.Text className={css.toolbarTime} onClick={() => setOpenView('hours')}>{time?.substring(0, 2) || '--'}</Fluent.Text>{':'}
+        <Fluent.Text className={css.toolbarTime} onClick={() => setOpenView('minutes')}>{time?.substring(3, 5) || '--'}</Fluent.Text>{' '}
+        <Fluent.Text className={css.toolbarTime} onClick={switchAmPm}>{time?.substring(6, 8) || ''}</Fluent.Text>
+      </Fluent.Text>
+    </div>
+
+  },
+  TextInput = ({ props: { onClick, onChange, placeholder, value, label, required, hour_cycle = '12', errorMsg, error, disabled } }: TextInputProps) =>
```
mturoci, line 132: No need for a separate component if it's just a single root anyway. Remove both props obj and propsType.

ui/src/time_picker.tsx
```diff
@@ -0,0 +1,269 @@
...
+      required={required}
+      styles={{ field: { cursor: 'pointer' }, icon: { bottom: 7 } }}
+      errorMessage={error ? errorMsg : undefined}
+    />
+
+
+export const
+  XTimePicker = ({ model: m }: { model: TimePicker }) => {
+    const
+      { hour_cycle = '12' } = m,
```
mturoci, line 151: It's better to define defaults in a single place if possible.

```suggestion
            { hour_cycle ='24', label = 'Select a time', disabled, min, max, placeholder, required, minutes_step = 1 } = m,,
```

ui/src/time_picker.tsx
```diff
@@ -0,0 +1,269 @@
...
+      defaultVal = m.value ? parseTimeStringToDate(m.value,) : null,
+      [value, setValue] = React.useState(defaultVal),
```
mturoci, line 153: Is defaultVal really necessary? If not:
```suggestion
      [value, setValue] = React.useState(m.value ? parseTimeStringToDate(m.value) : null),
```

ui/src/time_picker.tsx
```diff
@@ -0,0 +1,269 @@
...
+        hour_cycle: hour_cycle,
+        disabled: m.disabled,
+        errorMsg: `Wrong input. Please enter the time in range from 
+          ${formatDateToTimeString(parseTimeStringToDate(m.min || '00:00'), hour_cycle)} to 
+          ${formatDateToTimeString(parseTimeStringToDate(m.max || '00:00'), hour_cycle)}.
+        `
+      }
+
+    React.useEffect(() => {
+      wave.args[m.name] = defaultVal ? formatDateToTimeString(defaultVal, '24') : null
```
mturoci, line 220: Is there any reason why we cannot use `value` instead of `defaultVal` here?

ui/src/time_picker.tsx
```diff
@@ -0,0 +1,269 @@
...
+      <React.Suspense fallback={<LazyLoadPlaceholder />}>
+        {theme && AdapterDateFns
+          ? <ThemeProvider theme={theme}>
+            <LocalizationProvider dateAdapter={AdapterDateFns}>
+              <TimePicker
+                value={value}
+                label={m.label}
+                open={isDialogOpen}
+                onChange={(value, _keyboardInputValue) => setValue(value as Date)}
+                onAccept={(value) => onSelectTime(value as Date)}
```
mturoci, line 245: Nit
```suggestion
                onAccept={value => onSelectTime(value as Date)}
```

ui/src/time_picker.tsx
```diff
@@ -0,0 +1,269 @@
...
+    return <>
+      <React.Suspense fallback={<LazyLoadPlaceholder />}>
+        {theme && AdapterDateFns
+          ? <ThemeProvider theme={theme}>
+            <LocalizationProvider dateAdapter={AdapterDateFns}>
+              <TimePicker
+                value={value}
+                label={m.label}
+                open={isDialogOpen}
+                onChange={(value, _keyboardInputValue) => setValue(value as Date)}
```
mturoci, line 244: Nit

```suggestion
                onChange={value => setValue(value as Date)}
```

ui/src/time_picker.tsx
```diff
@@ -0,0 +1,269 @@
...
+                onAccept={(value) => onSelectTime(value as Date)}
+                onClose={() => setIsDialogOpen(false)}
+                ampm={hour_cycle === '12'}
+                showToolbar={true}
+                ToolbarComponent={params => <Toolbar params={params} label={m.label} switchAmPm={switchAmPm} />}
+                PopperProps={{ anchorEl: () => textInputRef.current as VirtualElement, ...popoverProps }}
+                minTime={m.min ? parseTimeStringToDate(m.min) : undefined}
+                maxTime={m.max ? parseTimeStringToDate(m.max) : undefined}
```
mturoci, line 252: Nit
```suggestion
                ToolbarComponent={params => <Toolbar params={params} label={m.label} switchAmPm={switchAmPm} />}
                PopperProps={{ anchorEl: () => textInputRef.current as VirtualElement, ...popoverProps }}
                minTime={min ? parseTimeStringToDate(min) : undefined}
                maxTime={max ? parseTimeStringToDate(max) : undefined}
```

ui/src/time_picker.tsx
```diff
@@ -0,0 +1,269 @@
...
+                minutesStep={[1, 5, 10, 15, 20, 30, 60].includes(m.minutes_step || 1) ? m.minutes_step : 1}
+                disabled={m.disabled}
```
mturoci, line 254: Nit
```suggestion
                disabled={disabled}
```

ui/src/time_picker.tsx
```diff
@@ -0,0 +1,269 @@
...
+                onChange={(value, _keyboardInputValue) => setValue(value as Date)}
+                onAccept={(value) => onSelectTime(value as Date)}
+                onClose={() => setIsDialogOpen(false)}
+                ampm={hour_cycle === '12'}
+                showToolbar={true}
+                ToolbarComponent={params => <Toolbar params={params} label={m.label} switchAmPm={switchAmPm} />}
+                PopperProps={{ anchorEl: () => textInputRef.current as VirtualElement, ...popoverProps }}
+                minTime={m.min ? parseTimeStringToDate(m.min) : undefined}
+                maxTime={m.max ? parseTimeStringToDate(m.max) : undefined}
+                minutesStep={[1, 5, 10, 15, 20, 30, 60].includes(m.minutes_step || 1) ? m.minutes_step : 1}
```
mturoci, line 253:
```suggestion
                minutesStep={allowedMinutesSteps[minutes_step]}
```

where `allowedMinutesSteps` is a constant:

```ts
  allowedMinutesSteps: { [key: U]: U } = { 1: 1, 5: 5, 10: 10, 15: 15, 20: 20, 30: 30, 60: 60 }
```

ui/src/time_picker.tsx
```diff
@@ -0,0 +1,269 @@
...
+                disabled={m.disabled}
+                onOpen={onOpen}
+                renderInput={({ inputProps, error }: TextFieldProps) =>
+                  <div ref={textInputRef} data-test={m.name}>
+                    <TextInput props={{ onChange: inputProps?.onChange, error, ...textInputProps }} />
```
mturoci, line 258:
```suggestion
                    <Fluent.TextField
                      iconProps={{ iconName: 'Clock', }}
                      onClick={() => setIsDialogOpen(true)}
                      onChange={inputProps?.onChange}
                      placeholder={placeholder}
                      disabled={disabled}
                      readOnly
                      value={value ? formatDateToTimeString(value, time_format_12h) : ''}
                      label={label}
                      required={required}
                      styles={{ field: { cursor: 'pointer' }, icon: { bottom: 7 } }}
                      errorMessage={error ? getErrMsg(time_format_12h, min, max) : undefined}
                    />
```

ui/src/time_picker.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+    expect(wave.args[name]).toBeNull()
+  })
+
+  it('Sets args - init - value specified', async () => {
+    render(<XTimePicker model={{ ...timepickerProps, value: '10:30' }} />)
+    await waitForIdleEventLoop()
+    expect(wave.args[name]).toBe('10:30')
+  })
+
+  it('Show correct input value in 12 hour time format', async () => {
```
mturoci, line 56: Is this correct? Shouldn't the default be `24`?

ui/src/time_picker.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const { getByDisplayValue } = render(<XTimePicker model={{ ...timepickerProps, value: '14:30' }} />)
+    await waitForIdleEventLoop()
+    expect(getByDisplayValue('02:30 PM')).toBeInTheDocument()
+  })
+
+  it('Shows midnight correctly in 12 hour time format', async () => {
```
mturoci, line 62: I would also add a an expect for `wave.args` to all these kind of tests to check if `wave.args` are still submitted in 24h format despite visual format being 12.

mturoci (changes requested):
Thanks @marek-mihok!

Note that when you create a docs page for a new component (`time_picker.md`), it needs to be added [here](https://github.com/h2oai/wave/blob/master/website/sidebars.js).

marek-mihok:
Thanks for a deep review @mturoci!
> Note that when you create a docs page for a new component (time_picker.md), it needs to be added [here](https://github.com/h2oai/wave/blob/master/website/sidebars.js).

Should it be also present in [website/widgets/overview.md](https://github.com/h2oai/wave/blob/master/website/widgets/overview.md) and [py/examples/form.py](https://github.com/h2oai/wave/blob/master/py/examples/form.py)?

mturoci:
> Should it be also present in [website/widgets/overview.md](https://github.com/h2oai/wave/blob/master/website/widgets/overview.md) and [py/examples/form.py](https://github.com/h2oai/wave/blob/master/py/examples/form.py)?

Yes, thanks.

ui/src/time_picker.tsx
```diff
@@ -0,0 +1,269 @@
...
+    disabled?: B
+  }
+}
+
+const
+  ThemeProvider = React.lazy(() => import('@mui/material/styles').then(({ ThemeProvider }) => ({ default: ThemeProvider }))),
+  LocalizationProvider = React.lazy(() => import('@mui/x-date-pickers').then(({ LocalizationProvider }) => ({ default: LocalizationProvider }))),
+  TimePicker = React.lazy(() => import('@mui/x-date-pickers').then(({ TimePicker }) => ({ default: TimePicker }))),
+  parseTimeStringToDate = (time: S) => new Date(`2000-01-01T${time.slice(0, 5)}:00`),
+  formatDateToTimeString = (date: Date, time_format_12h: B = true) => date.toLocaleString('en-US', { hour: '2-digit', minute: '2-digit', hourCycle: time_format_12h ? 'h12' : 'h23' }),
```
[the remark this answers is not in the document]
marek-mihok: Times like 24:00, 24:56 instead of 00:00, 00:56 look definitely odd to me and I've read it is used just rarely. In time of writing this code I was even thinking it is a bug but then I found out that this is called 24 hour cycle. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/hourCycle 
So I've used 23 hour life cycle to show the time in the most common way.

Even in the library we are using for time picker we can't select number 24 since there is 00 instead. Do we really want to support 24:xx times? I don't think it will be a common use case.

But you are right there is an inconsistency. According to current implementation of the time picker API should accept `12` and `23`. But now there's a question - which one to support/use? `24` or `23`?

ui/src/time_picker.test.tsx
marek-mihok: Yes, but @lo5 mentioned `12` as the default: https://github.com/h2oai/wave/pull/1609#issuecomment-1241159949 
I agree since most of our customers are based in the US where 12 hour time format is used.

mturoci: Oh I thought we have 24h default, ignore my comment then :) Thanks.

ui/src/time_picker.tsx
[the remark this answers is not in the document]
mturoci: > Times like 24:00, 24:56 instead of 00:00, 00:56 look definitely odd to me

It seems like there is a difference between 24h **cycle** and 24h **time format**.

If we want to expose the cycle, not the format, IMO we should follow the implementation meaning both exposing and using the same hour cycle (exposing `23` and using `23` and so on). Otherwise, I am not sure how we would handle future requests for the `24h` cycle if we already used it for the `23` if that makes sense.

wdyt @lo5 ^^ ?

Btw https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/hourCycle has quite [low support](https://caniuse.com/?search=Intl.Locale) so I would suggest looking at different way of getting the right output (maybe newly imported DateFns has something with better support).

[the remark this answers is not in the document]
lo5: Use `hour_format='12'` and `hour_format='24'` - that should keep things simple.

ui/src/time_picker.tsx
```diff
@@ -0,0 +1,259 @@
...
+  /** The width of the time picker, e.g. '100px'. Defaults to '100%'. */
+  width?: S
+  /** True if the component should be visible. Defaults to True. */
+  visible?: B
+  /** True if the form should be submitted when the time is selected. */
+  trigger?: B
+  /** True if this is a required field. Defaults to False. */
+  required?: B
+  /** Specifies 12-hour or 24-hour time format. One of `12` or `24`. Defaults to `24`. */
+  hour_format?: S
```
mturoci, line 52: I guess default here should be `12`?
```suggestion
  /** Specifies 12-hour or 24-hour time format. One of `12` or `24`. Defaults to `12`. */
  hour_format?: S
```

ui/src/time_picker.tsx
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+type ToolbarProps = { time: S, setOpenView: (view: CalendarOrClockPickerView) => void, label: S | undefined, switchAmPm: () => void }
+
+const
+  ThemeProvider = React.lazy(() => import('@mui/material/styles').then(({ ThemeProvider }) => ({ default: ThemeProvider }))),
+  LocalizationProvider = React.lazy(() => import('@mui/x-date-pickers').then(({ LocalizationProvider }) => ({ default: LocalizationProvider }))),
+  TimePicker = React.lazy(() => import('@mui/x-date-pickers').then(({ TimePicker }) => ({ default: TimePicker }))),
+  parseTimeStringToDate = (time: S) => new Date(`2000-01-01T${time.slice(0, 5)}:00`),
+  useMuiTheme = (themeObj: ThemeOptions) => {
```
mturoci, line 100: Can this be within a single hook rather than 3 separate ones?

ui/src/time_picker.tsx
```diff
@@ -0,0 +1,259 @@
...
+    const [theme, setTheme] = React.useState<Theme>()
+    React.useEffect(() => {
+      import('@mui/material/styles')
```
mturoci, line 103: Can we import a single function instead of whole style module here?
```suggestion
      import('@mui/material/styles/createTheme')
```

ui/src/time_picker.tsx
```diff
@@ -0,0 +1,259 @@
...
+      }
+    }
+  }
+
+type ToolbarProps = { time: S, setOpenView: (view: CalendarOrClockPickerView) => void, label: S | undefined, switchAmPm: () => void }
+
+const
+  ThemeProvider = React.lazy(() => import('@mui/material/styles').then(({ ThemeProvider }) => ({ default: ThemeProvider }))),
+  LocalizationProvider = React.lazy(() => import('@mui/x-date-pickers').then(({ LocalizationProvider }) => ({ default: LocalizationProvider }))),
+  TimePicker = React.lazy(() => import('@mui/x-date-pickers').then(({ TimePicker }) => ({ default: TimePicker }))),
```
mturoci, line 98: It would be better if we could import just the stuff we need instead of whole modules. Would something like this work?

```suggestion
  ThemeProvider = React.lazy(() => import('@mui/material/styles/ThemeProvider').then(({ ThemeProvider }) => ({ default: ThemeProvider }))),
  LocalizationProvider = React.lazy(() => import('@mui/x-date-pickers/LocalizationProvider').then(({ LocalizationProvider }) => ({ default: LocalizationProvider }))),
  TimePicker = React.lazy(() => import('@mui/x-date-pickers/TimePicker').then(({ TimePicker }) => ({ default: TimePicker }))),
```

ui/src/time_picker.tsx
```diff
@@ -0,0 +1,259 @@
...
+    React.useEffect(() => {
+      import('@mui/x-date-pickers/AdapterDateFns')
+        .then(({ AdapterDateFns }) => { setAdapterDateFns(() => AdapterDateFns) })
+    }, [])
+    return [AdapterDateFns]
+  },
+  useDateFns = () => {
+    const [dateFns, setDateFns] = React.useState<any>()
+    React.useEffect(() => {
+      import('date-fns')
```
mturoci, line 120: Can we import a single function instead of whole style module here?

```suggestion
      import('date-fns/esm/format')
```

ui/src/time_picker.tsx
```diff
@@ -0,0 +1,259 @@
...
+  useAdapterDateFns = () => {
+    const [AdapterDateFns, setAdapterDateFns] = React.useState<typeof DateFnsUtils | null>()
+    React.useEffect(() => {
+      import('@mui/x-date-pickers/AdapterDateFns')
+        .then(({ AdapterDateFns }) => { setAdapterDateFns(() => AdapterDateFns) })
+    }, [])
+    return [AdapterDateFns]
+  },
+  useDateFns = () => {
+    const [dateFns, setDateFns] = React.useState<any>()
```
mturoci, line 118: Typing here would be great.

```suggestion
    const [dateFns, setDateFns] = React.useState<typeof format>()
```

mturoci (changes requested):
Thanks @marek-mihok! A few final comments.

ui/src/time_picker.tsx
marek-mihok: My fault, thanks for noticing!

ui/src/time_picker.tsx
marek-mihok: We can, but why not `import('date-fns/format')`?

ui/src/time_picker.tsx
marek-mihok: It works, but not with jest. It looks like jest cannot read these files.

```
console.error node_modules/jest-environment-jsdom-fourteen/node_modules/jsdom/lib/jsdom/virtual-console.js:29
    Error: Uncaught [SyntaxError: Unexpected token 'export']
        at reportException (/Users/mmihok/Documents/h2o-ai/projects/wave/repo/wave/ui/node_modules/jest-environment-jsdom-fourteen/node_modules/jsdom/lib/jsdom/living/helpers/runtime-script-errors.js:62:24)
        at innerInvokeEventListeners (/Users/mmihok/Documents/h2o-ai/projects/wave/repo/wave/ui/node_modules/jest-environment-jsdom-fourteen/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:331:9)
        at invokeEventListeners (/Users/mmihok/Documents/h2o-ai/projects/wave/repo/wave/ui/node_modules/jest-environment-jsdom-fourteen/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:266:3)
        at HTMLUnknownElementImpl._dispatch (/Users/mmihok/Documents/h2o-ai/projects/wave/repo/wave/ui/node_modules/jest-environment-jsdom-fourteen/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:212:11)
        at HTMLUnknownElementImpl.dispatchEvent (/Users/mmihok/Documents/h2o-ai/projects/wave/repo/wave/ui/node_modules/jest-environment-jsdom-fourteen/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:87:17)
        at HTMLUnknownElement.dispatchEvent (/Users/mmihok/Documents/h2o-ai/projects/wave/repo/wave/ui/node_modules/jest-environment-jsdom-fourteen/node_modules/jsdom/lib/jsdom/living/generated/EventTarget.js:144:23)
        at Object.invokeGuardedCallbackDev (/Users/mmihok/Documents/h2o-ai/projects/wave/repo/wave/ui/node_modules/react-dom/cjs/react-dom.development.js:3994:16)
        at invokeGuardedCallback (/Users/mmihok/Documents/h2o-ai/projects/wave/repo/wave/ui/node_modules/react-dom/cjs/react-dom.development.js:4056:31)
        at beginWork$1 (/Users/mmihok/Documents/h2o-ai/projects/wave/repo/wave/ui/node_modules/react-dom/cjs/react-dom.development.js:23964:7)
        at performUnitOfWork (/Users/mmihok/Documents/h2o-ai/projects/wave/repo/wave/ui/node_modules/react-dom/cjs/react-dom.development.js:22779:12) Jest encountered an unexpected token
    
    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.
    
    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".
    
    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.
    
    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html
    
    Details:
    
    /Users/mmihok/Documents/h2o-ai/projects/wave/repo/wave/ui/node_modules/@mui/material/styles/ThemeProvider.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){export { ThemeProvider as default } from '@mui/system';
                                                                                             ^^^^^^
    
    SyntaxError: Unexpected token 'export'
        at new Script (vm.js:102:7)
        at ScriptTransformer._transformAndBuildScript (/Users/mmihok/Documents/h2o-ai/projects/wave/repo/wave/ui/node_modules/@jest/transform/build/ScriptTransformer.js:537:17)
        at ScriptTransformer.transform (/Users/mmihok/Documents/h2o-ai/projects/wave/repo/wave/ui/node_modules/@jest/transform/build/ScriptTransformer.js:579:25)
        at Runtime._execModule (/Users/mmihok/Documents/h2o-ai/projects/wave/repo/wave/ui/node_modules/jest-runtime/build/index.js:842:53)
        at Runtime._loadModule (/Users/mmihok/Documents/h2o-ai/projects/wave/repo/wave/ui/node_modules/jest-runtime/build/index.js:577:12)
        at Runtime.requireModule (/Users/mmihok/Documents/h2o-ai/projects/wave/repo/wave/ui/node_modules/jest-runtime/build/index.js:433:10)
        at Runtime.requireModuleOrMock (/Users/mmihok/Documents/h2o-ai/projects/wave/repo/wave/ui/node_modules/jest-runtime/build/index.js:598:21)
        at /Users/mmihok/Documents/h2o-ai/projects/wave/repo/wave/ui/src/time_picker.tsx:96:36
        at processTicksAndRejections (internal/process/task_queues.js:95:5)

  console.error node_modules/react-dom/cjs/react-dom.development.js:20085
    The above error occurred in one of your React components:
    
        at Lazy
        at Suspense
        at XTimePicker (/Users/mmihok/Documents/h2o-ai/projects/wave/repo/wave/ui/src/time_picker.tsx:132:27)
    
    Consider adding an error boundary to your tree to customize error handling behavior.
    Visit https://reactjs.org/link/error-boundaries to learn more about error boundaries.
```

ui/src/time_picker.tsx
mturoci: If the bundle stays the same size, then `date-fns/format` is fine as well.

ui/src/time_picker.tsx
mturoci: The problem is not in reading, but in parsing the files - need to be transformed first via `babel` (https://stackoverflow.com/questions/60714101/how-to-setup-jest-with-node-modules-that-use-es6). However, seems like this is only a problem for ThemeProvider, the rest of lazy components work fine.

Please add TODO to the ThemeProvider then so it's not forgotten and we can move ahead with the release today.

ui/src/time_picker.tsx
mturoci: Although `esm` shall be smaller, I assume you will hit the same problem as [here](https://github.com/h2oai/wave/pull/1609#discussion_r970393299) so just add TODO.

ui/src/time_picker.tsx
marek-mihok: UPDATE: The problem causes just `import('@mui/material/styles/ThemeProvider')` and even the bundle is little bit larger. Other two imports `import('@mui/x-date-pickers/LocalizationProvider')` and  `import('@mui/x-date-pickers/TimePicker')` work perfect and are reducing bundle size (`40+ kB`).

ui/src/time_picker.tsx
marek-mihok: It causes troubles with jest. The bundle size is the same though.

ui/src/time_picker.tsx
marek-mihok: Sorry, your message above was not displayed while writing last message. I'll add TODO.

ui/src/time_picker.tsx
marek-mihok: Seems there is a `6kB` difference. I'll add TODO as well.

mturoci (approved):
Thanks @marek-mihok, good job!

PR #1611, opened 2022-09-02
Fix: Success message without file names when ui.file_upload inside ui.dialog

### LINKED ISSUE #1607: ui.file_upload sucess message bar doesn't show correct file names
#### Wave SDK Version, OS

wave 0.22.0

#### Actual behavior

`ui.file_upload` doesn't show proper file names in success message
grafik
(Shows "." if 1 file was uploaded and adds a "," per additional file)

#### Expected behavior

`ui.file_upload` shows proper file names in success message

#### Steps To Reproduce

Run the following app and upload 1 or more files:

```python
from h2o_wave import Q, app, main, ui  # noqa F401

@app("/")
async def serve(q: Q):
    if not q.client.init:
        q.page["meta"] = ui.meta_card(box="")
        await q.page.save()
        q.page["meta"].dialog = ui.dialog(
            title="Import Datasets",
            primary=True,
            items=[
                ui.file_upload(
                    name="file_import",
                    label="Import",
                    multiple=True,
                    file_extensions=["csv", "gz", "zip", "tar"],
                ),
                ui.button(name="close_dialog", icon="Cancel"),
            ],
        )
        await q.page.save()
        q.client.init = True
```

grafik

marek-mihok:
Closes #1607

### DIFF
--- a/ui/src/file_upload.tsx
+++ b/ui/src/file_upload.tsx
@@ -138,7 +138,7 @@ export const
           wave.args[name] = files
 
           if (!compact) wave.push()
-          setSuccessMsg(`Successfully uploaded files: ${files.map(({ name }: File) => name).join(',')}.`)
+          setSuccessMsg(`Successfully uploaded files: ${fileNames}.`)
         }
         catch (err: unknown) {
           if (err) setError(err instanceof XMLHttpRequest ? err.responseText : 'There was an error when uploading file.')


ui/src/file_upload.tsx
```diff
@@ -138,7 +138,7 @@ export const
           wave.args[name] = files
 
           if (!compact) wave.push()
-          setSuccessMsg(`Successfully uploaded files: ${files.map(({ name }: File) => name).join(',')}.`)
+          setSuccessMsg(`Successfully uploaded files: ${fileNames}.`)
```
mturoci, line 141: Previously, we displayed paths to uploaded files on the server (`/f/<hash>/filename`), but showing the file names that the user actually uploaded might be better. The question is why has this broken? Is there any problem with the server response?

Note: @marek-mihok it's good practice to always include a short description of what's the root cause when fixing bugs.

marek-mihok: @mturoci  sure. There is a difference between `files` as a state, which look like this:
```
[
  {
  lastModified: 1655915055755
  lastModifiedDate: Wed Jun 22 2022 18:24:15 GMT+0200 (Central European Summer Time) {}
  name: "app.zip"
  size: 2806
  type: "application/zip"
  webkitRelativePath: ""
  },
...
]
```
and `files` which were declared in the try block of the upload function (`const { files } = JSON.parse(responseText)` which has the format of paths as you mentioned: `['/_f/233223fe-9b04-4035-9d25-0c3563c7da62/app.zip', '/_f/ed827fe9-dd9a-4e77-965c-40e4c4a13131/smtg.zip']`

The problem is that in `setSuccessMsg` there was an attempt to iterate through the `files` from second case 
```
setSuccessMsg(`Successfully uploaded files: ${files.map(({ name }: File) => name).join(',')}.`)
``` 
as it was the `files` from the state.

Could server response change? Was it really working before?

In our demo app we are showing what's uploaded into wave server directly after upload
```
@app('/demo')
async def serve(q: Q):
    if 'file_upload' in q.args:
            ui.text(f'file_upload={q.args.file_upload}'),
            ui.button(name='show_upload', label='Back', primary=True),
        ])
    else:
            box='1 1 4 10',
            items=[
                ui.file_upload(name='file_upload', label='Upload!', multiple=True,
                               file_extensions=["csv", "gz", "zip", "tar"])
            ]
        )
    await q.page.save()
```
and therefore user can see the paths to uploaded files on the server.

However this is missing [here](https://github.com/h2oai/wave/issues/1607#issue-1350686936) in the app for the issue reproduction and the app relies on the `ui.file_upload` success message implementation, which is wrong.

mturoci: So are files uploaded correctly and 200 is returned with correct server paths as expected?

marek-mihok: It seems yes:
```
status: 200 
responseText: {"files":["/_f/e13c0991-7567-45a8-9259-23536832fd49/app.zip","/_f/26ee3d9d-9765-45be-97bb-92303956bf85/smtg.zip"]}
```

mturoci (approved):

PR #1612, opened 2022-09-02
Feat: Make ui.text_annotator to respect \n and \r in text

### LINKED ISSUE #1358: Let ui.text_annotator respect \n and \r in text

Currently, `ui.text_annotator` strips off any `\n` `\r` in the text prior to rendering. This makes formatted texts to appear in a single line and additionally results in wrong character level offsets for the annotations (possible bug). 

Let `ui.text_annotator` respect `\n` and `\r` in text for rendering and character offset calculations. To help the annotation process, this could be made visible by showing acutal `\n` or invisible by adding newlines. 

The issue probably extends to all invisible characters.

marek-mihok:
Javascript's newline escape sequences `\n` and `\r` are now transformed into line breaks in **ui.text_annotator**:
```py
ui.text_annotator_item(text=' is a member of the hip\rhop super\ngroup ')
```
image

These escape characters are only included in selection when user spans the selection through multiple lines.

Closes #1358

### DIFF
--- a/py/examples/text_annotator.py
+++ b/py/examples/text_annotator.py
@@ -23,7 +23,7 @@ async def serve(q: Q):
                 ],
                 items=[
                     ui.text_annotator_item(text='Killer Mike', tag='p'),
-                    ui.text_annotator_item(text=' is a member, of the hip hop supergroup '),  # no tag
+                    ui.text_annotator_item(text=' is a member of the hip hop supergroup '),  # no tag
                     ui.text_annotator_item(text='Run the Jewels', tag='o'),
                 ],
             ),
--- a/py/examples/text_annotator.py
+++ b/py/examples/text_annotator.py
@@ -25,7 +25,9 @@ async def serve(q: Q):
                     ui.text_annotator_item(text='Killer Mike', tag='p'),
                     ui.text_annotator_item(text=' is a member of the hip hop supergroup '),  # no tag
                     ui.text_annotator_item(text='Run the Jewels', tag='o'),
-                    ui.text_annotator_item(text='.\nIt is also known by the initials RTJ.')
+                    ui.text_annotator_item(text='.\nIt is also known by the initials '),
+                    ui.text_annotator_item(text='RTJ', tag='o'),
+                    ui.text_annotator_item(text='.')
                 ],
             ),
             ui.button(name='submit', label='Submit', primary=True)
--- a/py/examples/text_annotator.py
+++ b/py/examples/text_annotator.py
@@ -25,6 +25,7 @@ async def serve(q: Q):
                     ui.text_annotator_item(text='Killer Mike', tag='p'),
                     ui.text_annotator_item(text=' is a member of the hip hop supergroup '),  # no tag
                     ui.text_annotator_item(text='Run the Jewels', tag='o'),
+                    ui.text_annotator_item(text='.\nIt is also known by the initials RTJ.')
                 ],
             ),
             ui.button(name='submit', label='Submit', primary=True)
--- a/ui/src/text_annotator.test.tsx
+++ b/ui/src/text_annotator.test.tsx
@@ -252,5 +252,35 @@ describe('TextAnnotator.tsx', () => {
         { text: 'ood', tag: 'tag1' },
         { text: ' day' }])
     })
+
+    it('Sets correct args on annotate when escape characters are included - single line selection', () => {
+      const customItems = [{ text: 'I am\nmultiline\ntext.' }]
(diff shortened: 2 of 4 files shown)


mturoci (approved):
Looking great @marek-mihok! Please also think about how to best showcase this behavior in our docs/example.

marek-mihok:
@mturoci wdyt about these examples?

- **docs example:**

image


- **tour example:**

```py
items=[
    ui.text_annotator_item(text='Killer Mike', tag='p'),
    ui.text_annotator_item(text=' is a member of the hip hop supergroup '),  # no tag
    ui.text_annotator_item(text='Run the Jewels', tag='o'),
    ui.text_annotator_item(text='.\nIt is also known by the initials RTJ.')
],
```
image

mturoci:
^^ This may be one usecase, but the more common one would I think be:

```py
# Form / TextAnnotator
# Use text annotator when you need to highlight text phrases.
# #form #annotator
# ---


@app('/demo')
async def serve(q: Q):
    if q.args.annotator:
        q.page['example'].items = [
            ui.text(f'annotator={q.args.annotator}'),
            ui.button(name='show_form', label='Back', primary=True),
        ]
    else:
        txt = '''
I am
multiline
text
''' 
            ui.text_annotator(
                name='annotator',
                title='Select text to annotate',
                tags=[
                    ui.text_annotator_tag(name='p', label='Person', color='#F1CBCB'),
                    ui.text_annotator_tag(name='o', label='Org', color='#CAEACA'),
                ],
                items=[
                    ui.text_annotator_item(text=txt),
                ],
            ),
            ui.button(name='submit', label='Submit', primary=True)
        ])
    await q.page.save()
```
Looking at the example above, we should probably ignore newline if it's the first character of text.

Showing both usecases in the docs will be probably the best.

pascal-pfeiffer:
Thanks a lot @marek-mihok for implementing this feature!

The implementation looks good to me. Unfortunately, I wasn't able to thoroughly test the functionality with our repository, yet. In your example above (`Okay.\nExcellent work...`), is the trailing `\n` added to the annotated text span? I would argue, that any trailing `\n` should not be included in the selection unless specifically marked (e.g. by draging the mouse to the start of the next line while annotating). It should not be included when doing a "smart selection" click on one word. 


> Looking at the example above, we should probably ignore newline if it's the first character of text.

What exactly is meant with the first character of text? It is important that all characters are preserved in the annotation step.

mturoci:
>What exactly is meant with the first character of text

Running my example above with text

```py
txt = '''
I am
multiline
text
''' 
```

renders

image

whereas I would expect it to render

image

that is only achievable via

```py
txt = '''I am
multiline
text
''' 
```
which is somewhat ugly.

mturoci:
>trailing \n should not be included in the selection unless specifically marked

Agree, good point.

pascal-pfeiffer:
Thanks @mturoci . Still the former (including the first newline and a trailing \n) is the correct way. In my opinion there should be no arbitrarily removal of characters in the process of annotation and it would only make debugging character offsets more difficult. And usually text is loaded from a file or a database anyway, and would not be visible in static code.

mturoci:
> it would only make debugging character offsets more difficult. And usually text is loaded from a file or a database anyway

Fair points, no need for special-casing then.

marek-mihok:
> is the trailing \n added to the annotated text span?
@pascal-pfeiffer, the idea was, that this is the only way how `\n` can be added to the selection - to include it in the code explicitly.

> any trailing \n should not be included in the selection unless specifically marked (e.g. by draging the mouse to the start of the next line while annotating)
Dragging the mouse to the next line is currently not possible. There is no way to add `\n` into selection right now while annotating. Is it necessary? What would be the use case for this? I'm not an expert here :)

marek-mihok:
@pascal-pfeiffer I've checked the common office tools and I found out that newline characters are included in selection only if there is a line break character (`\r`, `\n`) in the end of the line and when selection is spanning multiple lines. It is not included in case when text is braked automatically in the end of the line. Basically, it is what you proposed. Therefore I'll implement it like this.

mturoci (commented):
Thanks @marek-mihok! LGTM. Can we also add some unit tests that verify correct line breaks are also returned according to [pascal's comment](https://github.com/h2oai/wave/pull/1612#issuecomment-1238310220)?

marek-mihok:
> Can we also add some unit tests that verify correct line breaks are also returned... ?
@mturoci, done ✅

mturoci (approved):
Thanks @marek-mihok! Looking great!

PR #1616, opened 2022-09-08
test: Add test for single-valued Combobox

### LINKED ISSUE #1615: ui.combobox returns different types of values
sometimes the return value is string (when selected by dropdown), sometimes, if you type some text in it, list is returned instead, for the same instance. 

You can also use this tutorial to reproduce https://wave.h2o.ai/docs/examples/combobox

aalencar:
Add test to ensure a new typed value is a string instead of an array like in Wave 0.22.

closes #1615

### DIFF
--- a/ui/src/combobox.test.tsx
+++ b/ui/src/combobox.test.tsx
@@ -85,6 +85,14 @@ describe('Combobox.tsx', () => {
     
         expect(pushMock).toHaveBeenCalled()
       })
+
+      it('Sets wave args as string when a new valued is typed after init', () => {
+        const { getByRole } = render(<XCombobox model={{ ...comboboxProps, value: 'A' }} />)
+    
+        userEvent.type(getByRole('combobox'), '{backspace}D{enter}')
+    
+        expect(wave.args[name]).toBe('D')
+      })
     })
 
     describe('Prop changes - update values dynamically from Wave app', () => {
--- a/ui/src/combobox.test.tsx
+++ b/ui/src/combobox.test.tsx
@@ -86,11 +86,36 @@ describe('Combobox.tsx', () => {
         expect(pushMock).toHaveBeenCalled()
       })
 
-      it('Sets wave args as string when a new valued is typed after init', () => {
+      it('Sets wave args as string when a new valued is typed and enter is pressed - after init', () => {
+        const { getByRole } = render(<XCombobox model={{ ...comboboxProps, value: 'A' }} />)
+        expect(wave.args[name]).toBe('A')
+        userEvent.type(getByRole('combobox'), '{backspace}D{enter}')
+        expect(wave.args[name]).toBe('D')
+      })
+
+      it('Sets wave args as string when a new valued is typed and user clicks away - after init', () => {
         const { getByRole } = render(<XCombobox model={{ ...comboboxProps, value: 'A' }} />)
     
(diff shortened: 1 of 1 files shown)


ui/src/combobox.test.tsx
```diff
@@ -76,15 +78,21 @@ describe('Combobox.tsx', () => {
...
         fireEvent.click(getByRole('presentation', { hidden: true }))
         fireEvent.click(getByText('A'))
     
         expect(pushMock).toHaveBeenCalled()
       })
+
+      it('Sets wave args as string when a new valued is typed after init', () => {
+        const { getByRole } = render(<XCombobox model={{ ...comboboxProps, value: 'A' }} />)
+    
+        userEvent.type(getByRole('combobox'), '{backspace}D{enter}')
```
mturoci, line 92: Enter shouldn't be required.
```suggestion
        userEvent.type(getByRole('combobox'), '{backspace}D')
```

aalencar: We need to trigger `onChange`, that's where the args are set. We either press enter or click outside the combobox (blur).

mturoci: Tests should mimic user behavior. Thus, end goal is to set `wave.args` without the need for enter (same as textbox).

aalencar: > Tests should mimic user behavior
It's doing exactly that.

> Thus, end goal is to set wave.args without the need for enter (same as textbox).
I don't get it. We want the wave args to represent the selected value. We have not selected any value yet as we type. Textbox has a different use case. Why would I want to set wave args on every keystroke if what matters is when the value is selected? That's why Fluent triggers onChange only on Enter or Blur.

mturoci: >It's doing exactly that.

It's not. Wave args are populated even without hitting the enter, see video below:

https://user-images.githubusercontent.com/64769322/189288169-34c6389a-59f4-4f7e-9c52-70d8ba18a3cd.mov

However, looking at the code, it seems like this is just a coincidence/bug causing correct behavior.

> Why would I want to set wave args on every keystroke if what matters is when the value is selected?

Users have no idea that enter/blur triggers the selection (it's an implementation detail). All they can see is that they can either type or select an option. The most common scenario for typing is depicted in the video above where enter is not used after typing is finished which should be also depicted in the unit test.

aalencar: Thanks for clarifying your point and for the screencast.

> It's not. Wave args are populated even without hitting the enter, see video below:
However, looking at the code, it seems like this is just a coincidence/bug causing correct behavior.

It's not a bug nor coincidence. You might not be considering that when you click on submit the combobox loses focus which in turn triggers onChange.

> Users have no idea that enter/blur triggers the selection (it's an implementation detail). 

Yes, they don't need to know the implementation. They will type and do something else (e.g. press enter, press tab, click submit, select a different form) and wave.args will be set because onChange will be triggered.

> The most common scenario for typing is depicted in the video above where enter is not used after typing is finished which should be also depicted in the unit test.

I will add a test for the use case in the screencast as well.

mturoci: >when you click on submit the combobox loses focus which in turn triggers onChange

Yep, that's it. Thanks for clarification.

mturoci (approved):
Thanks @aalencar! LGTM.

PR #1621, opened 2022-09-14
fix: Add support to dynamically change Dropdown value

### LINKED ISSUE #1574: Dropdown value clearing
#### Wave SDK Version, OS

0.21.1

#### Actual behavior

I have two dropdowns and I need to clear the selected value of the second dropdown when the value of the first dropdown changes. 

This works fine with form by setting the second dropdown value as `None`.

```python
from h2o_wave import app, main, Q, ui


@app("/")
async def serve(q: Q) -> None:
    if q.args.dropdown1:
        if q.args.dropdown1 != q.client.dropdown1:
            q.client.dropdown1_changed = True
        else:
            q.client.dropdown1_changed = False
        q.client.dropdown1 = q.args.dropdown1
    if q.args.dropdown2:
        if q.client.dropdown1_changed:
            q.client.dropdown2 = None
        else:
            q.client.dropdown2 = q.args.dropdown2

    q.page["example"] = ui.form_card(
        box="1 1 2 2",
        items=[
            ui.dropdown(
                name="dropdown1",
                label="Dropdown1",
                value=q.client.dropdown1,
                trigger=True,
                choices=[
                    ui.choice(name="1", label="Choice 1"),
                    ui.choice(name="2", label="Choice 2"),
                    ui.choice(name="3", label="Choice 3"),
                ],
            ),
            ui.dropdown(
                name="dropdown2",
                label="Dropdown2",
                value=q.client.dropdown2,
                trigger=True,
                choices=[
                    ui.choice(name="a", label="Choice a"),
                    ui.choice(name="b", label="Choice b"),
                    ui.choice(name="c", label="Choice c"),
                ],
            ),
        ],
    )

    await q.page.save()
```

https://user-images.githubusercontent.com/29954570/183110581-37e9c894-6e90-415c-9391-51fdd2db78c3.mov

When the dropdowns are in a side panel, you need to set the value of the second dropdown as a value except `None`.

Scenario 1: Value as `None` - Not working.

``` python
from h2o_wave import app, main, Q, ui

@app("/")
async def serve(q: Q) -> None:
    if not q.client.initialized:
        q.page['example'] = ui.form_card(box='1 1 2 1', items=[
            ui.button(name='show_side_panel', label='Open side panel', primary=True)
        ])
        q.client.initialized = True
    else:
        if q.args.show_side_panel:
            show_side_panel(q)

        if q.args.dropdown1:
            if q.args.dropdown1 != q.client.dropdown1:
                q.client.dropdown2 = None
            q.client.dropdown1 = q.args.dropdown1
            show_side_panel(q)
        if q.args.dropdown2:
            q.client.dropdown2 = q.args.dropdown2
            show_side_panel(q)
        if q.args.cancel:
            q.page['meta'].side_panel = None
            q.client.dropdown1 = None
            q.client.dropdown2 = None

    await q.page.save()


def show_side_panel(q):
    q.page["meta"].side_panel = ui.side_panel(
        title="Side Panel",
        items=[
            ui.dropdown(
                name="dropdown1",
                label="Dropdown1",
                value=q.client.dropdown1,
                trigger=True,
                choices=[
                    ui.choice(name="1", label="Choice 1"),
                    ui.choice(name="2", label="Choice 2"),
                    ui.choice(name="3", label="Choice 3"),
                ],
            ),
            ui.dropdown(
                name="dropdown2",
                label="Dropdown2",
                value=q.client.dropdown2,
                trigger=True,
                choices=[
                    ui.choice(name="a", label="Choice a"),
                    ui.choice(name="b", label="Choice b"),
                    ui.choice(name="c", label="Choice c"),
                ],
            ),
            ui.button(name='cancel', label='Cancel')
        ],
        events=["dismissed"],
        closable=True,
    )
```

https://user-images.githubusercontent.com/29954570/183111407-bde3cb22-b764-495d-91f4-e2fb001288b4.mov

Scenario 2: Value as an empty string - working.

```  python
from h2o_wave import app, main, Q, ui

@app("/")
async def serve(q: Q) -> None:
    if not q.client.initialized:
        q.page['example'] = ui.form_card(box='1 1 2 1', items=[
            ui.button(name='show_side_panel', label='Open side panel', primary=True)
        ])
        q.client.initialized = True
    else:
        if q.args.show_side_panel:
            show_side_panel(q)

        if q.args.dropdown1:
            if q.args.dropdown1 != q.client.dropdown1:
                q.client.dropdown2 = ""
            q.client.dropdown1 = q.args.dropdown1
            show_side_panel(q)
        if q.args.dropdown2:
            q.client.dropdown2 = q.args.dropdown2
            show_side_panel(q)
        if q.args.cancel:
            q.page['meta'].side_panel = None
            q.client.dropdown1 = None
            q.client.dropdown2 = None

    await q.page.save()


def show_side_panel(q):
    q.page["meta"].side_panel = ui.side_panel(
        title="Side Panel",
        items=[
            ui.dropdown(
                name="dropdown1",
                label="Dropdown1",
                value=q.client.dropdown1,
                trigger=True,
                choices=[
                    ui.choice(name="1", label="Choice 1"),
                    ui.choice(name="2", label="Choice 2"),
                    ui.choice(name="3", label="Choice 3"),
                ],
            ),
            ui.dropdown(
                name="dropdown2",
                label="Dropdown2",
                value=q.client.dropdown2,
                trigger=True,
                choices=[
                    ui.choice(name="a", label="Choice a"),
                    ui.choice(name="b", label="Choice b"),
                    ui.choice(name="c", label="Choice c"),
                ],
            ),
            ui.button(name='cancel', label='Cancel')
        ],
        events=["dismissed"],
        closable=True,
    )
```

https://user-images.githubusercontent.com/29954570/183111658-97f40304-f07e-4fd1-85d2-bd9cfed24ace.mov

#### Expected behavior

Expect to clear the value of a dropdown in a side panel, when it is set as `None`.

aalencar:
## Multi-valued dialog Dropdown Refactor notes

On line 189, since `filteredItems` is intialized with `items`, `setFilteredItems` was mutating `items` when it set `i.checked = checked` which is definitely an antipattern, plus `items` uses `useMemo` which should only change when its dependencies change.
`189: selectAll = (checked = true) => () => setFilteredItems(filteredItems.map(i => { i.checked = checked; return i })),`

`filteredItems` should represent the displayed items and should rely on `items` as single source of truth. The refactoring makes it so that we change `items` and the changes are reflected to `filteredItem` through `useEffect`. Same for searching.

 I believe this is a better mental model.

## Unrelated bugs

I fixed a bug where multi-valued dialog Dropdown was setting wave args to string when it only had a single value selected and it was setting wave args to array for more than one value. Now it's always an array. No one has reported this issue but if they handled this in their code, the fix might break it.

## Performance concerns

We are using `toString()` to compare array content and it can be an expensive computation.

I tested dialog with 100k options and it was fine.

## Updating `choices` prop

This PR solves the issue which was concerned about changing `value` and `values`. Updating `choices` dynamically could be done in a separate PR due to the number of changes and complexity already present in this one.

## Further improvements

The code is still quite complex. We could split each component into single-valued and multi-valued like we did for dropdown. Maybe it would be easier to follow.

Any suggestion is welcomed.

## Handy repro code

```py
from h2o_wave import app, main, Q, ui

combobox_choices = ['A', 'B', 'C']

@app("/demo")
async def serve(q: Q) -> None:
    if not q.client.initialized:
        q.client.initialized = True
        q.page["example"] = ui.form_card(
            box="1 1 4 6",
            items=[
                ui.dropdown(
                    name='dp1',
                    label='base single',
                    choices=[
                        ui.choice(name="a", label="choice a"),
                        ui.choice(name="b", label="choice b"),
                        ui.choice(name="c", label="choice c"),
                    ],
                    value="a",
                ),
                 ui.dropdown(
                    name='dp2',
                    label='base multi',
                    choices=[
                        ui.choice(name="a", label="choice a"),
                        ui.choice(name="b", label="choice b"),
                        ui.choice(name="c", label="choice c"),
                    ],
                    values=["a"],
                ),
                ui.dropdown(
                    name='dp3',
                    label='dialog single',
                    choices=[
                        ui.choice(name="1", label="Choice 1"),
                        ui.choice(name="2", label="Choice 2"),
                        ui.choice(name="3", label="Choice 3"),
                    ],
                    popup="always",
                    value="1"
                ),
                ui.dropdown(
                    name='dp4',
                    label='dialog multi',
                    choices=[ui.choice(name=str(i), label=f'Choice {i}') for i in range(100000)],
                    values=["1"],
                    popup="always",
                ),
                ui.button(name="change", label="change"),
            ],
        )
    if q.args.change:
        q.page["example"].items[0].dropdown.value = 'b'
        q.page["example"].items[1].dropdown.values = ['b']
        q.page["example"].items[2].dropdown.value = '2'
        q.page["example"].items[3].dropdown.values = ['2']
        
    await q.page.save()
```

closes #1574

### DIFF
--- a/ui/src/dropdown.tsx
+++ b/ui/src/dropdown.tsx
@@ -81,12 +81,16 @@ const
       marginTop: 16
     }
   }),
-  BaseDropdown = ({ name, label, required, disabled, value, values, choices, trigger, placeholder }: Dropdown) => {
-    const
-      isMultivalued = !!values,
-      selection = React.useMemo(() => isMultivalued ? new Set<S>(values) : null, [isMultivalued, values]),
+  BaseDropdown = ({ model: m}: { model: Dropdown}) => {
+  const
+      { name, label, required, disabled, choices, trigger, placeholder } = m,
+      isMultivalued = !!m.values,
+      [stateValue, setStateValue] = React.useState(m.value ?? null),
+      selection = React.useMemo(() => isMultivalued ? new Set<S>(m.values) : null, [isMultivalued, m.values]),
       [selectedOptions, setSelectedOptions] = React.useState(Array.from(selection || [])),
       options = (choices || []).map(({ name, label, disabled }): Fluent.IDropdownOption => ({ key: name, text: label || name, disabled })),
+      previousValues = React.useRef<S[] | undefined>(),
+      skipUseEffect = React.useRef(false),
       onChange = (_e?: React.FormEvent<HTMLElement>, option?: Fluent.IDropdownOption) => {
         if (option) {
           const optionKey = option.key as S
@@ -94,9 +98,18 @@ const
             option.selected ? selection.add(optionKey) : selection.delete(optionKey)
 
             const selectedOpts = Array.from(selection)
+            previousValues.current = selectedOpts
             wave.args[name] = selectedOpts
             setSelectedOptions(selectedOpts)
           } else {
+            setStateValue(optionKey)
+            // HACK: The useEffect is only triggered if the above setter sets a different value and "m.value" changed.
+            // The next line does two things: 
+            // 1. Skip the useEffect so that it doesn't set wave args to same value if trigger is provided
+            // 2. m.value !== optionKey prevents the bug where user selects the same value twice and then the wave app tries to change the "value" prop
+            // e.g. choices = ['a', 'b', 'c'], selects 'c', selects 'c', wave app changes "value" to 'b'
+            skipUseEffect.current = m.value !== optionKey
+            m.value = optionKey
             wave.args[name] = optionKey
           }
         }
@@ -110,19 +123,40 @@ const
 
         const selectionArr = Array.from(selection)
         setSelectedOptions(selectionArr)
+        previousValues.current = selectionArr
         wave.args[name] = selectionArr
 
-        onChange()
+        if (trigger) wave.push()
       },
       deselectAll = () => {
         if (!selection) return
 
         selection.clear()
         setSelectedOptions([])
+        previousValues.current = []
         wave.args[name] = []
 
-        onChange()
+        if (trigger) wave.push()
       }
+
+    React.useEffect(() => {
+      if (!isMultivalued && !skipUseEffect.current) {
+        setStateValue(m.value ?? null)
+        wave.args[name] = m.value ?? null
+      }
+      skipUseEffect.current = false
+    }, [m.value, isMultivalued, name])
+
+    // Since "m.values" is an array, every time it's set from Wave App we create a new array reference and the useEffect bellow is triggered.
+    // Because of that we don't have to use the hack where "m.values" is set in "onChange" like we do for "m.value".
+    React.useEffect(() => {
+      if (isMultivalued ) {
+        setSelectedOptions(Array.from(m.values || []))
+        const isDifferent = previousValues.current?.toString() !== m.values?.toString()
+        if (isDifferent) wave.args[name] = m.values ?? []
+        previousValues.current = m.values
+      }
+    }, [m.values, isMultivalued, name])
 
     return (
       <>
@@ -135,7 +169,7 @@ const
           required={required}
           disabled={disabled}
           multiSelect={isMultivalued || undefined}
-          defaultSelectedKey={!isMultivalued ? value : undefined}
+          selectedKey={!isMultivalued ? stateValue : undefined}
           selectedKeys={isMultivalued ? selectedOptions : undefined}
           onChange={onChange}
         />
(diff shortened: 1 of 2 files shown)


mturoci:
Thanks for the great work @aalencar! I refactored your changes to make the components simpler. Please check the changes and apply additional refactors as we discussed.

aalencar:
I didn't see any major performance improvement on using Map over a more functional style, recreating the array every time and setting `items` props as needed.

ui/src/dropdown.tsx
```diff
@@ -118,12 +123,24 @@ const
...
         selection.clear()
-        setSelectedOptions([])
+        setMultiValues([])
         wave.args[name] = []
 
         onChange()
       }
 
+    React.useEffect(() => {
+      if (!isMultivalued) return
```
mturoci, line 133: Would this work if `isMultivalued` changes? It's not in the dep array. Checking `values` for `undefined` seems like a safer way here.

ui/src/dropdown.tsx
```diff
@@ -118,12 +123,24 @@ const
...
+      setMultiValues(values || [])
+      wave.args[name] = values ?? null
+    }, [name, values])
+
+    React.useEffect(() => {
+      if (isMultivalued) return
```
mturoci, line 139: Same as above.

ui/src/dropdown.tsx
```diff
@@ -152,119 +169,160 @@ const
...
+      : null,
 
-        if (values?.length) return values.map(v => itemsMap.get(v) || '').filter(Boolean).join(', ')
-        if (value) return itemsMap.get(value)
-      }),
-      toggleDialog = React.useCallback(() => setIsDialogHidden(!isDialogHidden), [isDialogHidden]),
-      cancelDialog = React.useCallback(() => {
+  DialogDropdownSingle = ({ model }: { model: Dropdown}) => {
+    const
+      { name, choices = [], disabled, required, trigger, placeholder, label } = model,
```
mturoci, line 196: nit: maybe we can use destructuring assignment for `value` here in the same way as in `BaseDropdown`?

ui/src/dropdown.tsx
```diff
@@ -152,119 +169,160 @@ const
...
+            </Fluent.DialogContent>
+          </Fluent.Dialog>
+        </>
+      )
+  },
+  DialogDropdownMulti = ({ name, choices = [], values = [], disabled, required, trigger, placeholder, label }: Dropdown) => {
+    const
+      [isDialogHidden, setIsDialogHidden] = React.useState(true),
+      [items, setItems, onSearchChange] = useItems(choices, values),
+      auxItems = React.useRef([...items]),
```
mturoci, line 249: Wouldn't this be enough? Also naming needs work. Something like initial selection or something. Aux is too general.

```suggestion
      auxItems = React.useRef(items),
```

ui/src/dropdown.tsx
```diff
@@ -152,119 +169,160 @@ const
...
+      openDialog = () => {
+        setIsDialogHidden(false)
+        auxItems.current = [...items]
```
mturoci, line 252: Same question as in initialization.

ui/src/dropdown.tsx
mturoci: No, I am referering to 
image

if it can be the same as BaseDropdown - your most recent impl.

Screen Shot 2022-09-26 at 11 10 04 AM

ui/src/dropdown.tsx
aalencar: > Wouldn't this be enough? 

It's enough because I am always creating a new array by using `map` in `setItems` and `auxItems`.
But semantically, `auxItems` is a copy so it makes sense to create a new array so that in the case someone (by mistake or not) mutates `items`, it won't be affected.

In terms of memory, I profiled `useRef(items)` and `useRef([...items])` and for some reason `useRef([...items])`  consistently performed slightly better. I don't know if using Chrome Devtools is enough to evaluate such a situation though.

3 and 7 are `useRef(items)` the rest is `useRef([...items])`.
image


>Also naming needs work. Something like initial selection or something. Aux is too general.

True. I'd not use `initialSelection` because it can imply that was it the first selection when the component was created. I'm going to use `itemsBackup` since it's used to revert to a previous state.

ui/src/dropdown.tsx
aalencar: [Discussion here ](https://github.com/h2oai/wave/pull/1621#discussion_r979763603)

ui/src/dropdown.tsx
mturoci: > so that in the case someone (by mistake or not) mutates items, it won't be affected.

This is a shallow copy, not a deep copy so this doesn't hold true - it should be affected in both scenarios.

> I don't know if using Chrome Devtools is enough to evaluate such a situation though.

AFAIK it should allow you to explore the complete memory snapshot item by item so you can see what's the root cause of the extra memory.

> it can imply that was it the first selection when the component was created. I'm going to use itemsBackup since it's used to revert to a previous state

What is previous state in this case? I thought it's the initial selection that is seen after opening the dialog, but maybe I am wrong?

ui/src/dropdown.tsx
aalencar: TBH, I was not even considering that we want to allow to dynamically switch between single and multi, but it worked by simply adding `isMultivalued` to dep array. One caveat: `values` needs to be set to `None` to switch from multi to single.

Repro code

```py
from h2o_wave import app, main, Q, ui


@app("/demo")
async def serve(q: Q) -> None:
    if not q.client.initialized:
        q.client.initialized = True
        q.page["example"] = ui.form_card(
            box="1 1 4 6",
            items=[
                ui.dropdown(
                    name='dropdown',
                    label='dropdown',
                    choices=[
                        ui.choice(name="1", label="Choice 1"),
                        ui.choice(name="2", label="Choice 2"),
                        ui.choice(name="3", label="Choice 3"),
                    ],
                    value="1",
                ),
                ui.button(name="single", label="single (value = 2)"),
                ui.button(name="multi", label="multi (values = [2, 3])"),
                ui.button(name="popup_on", label="popup on"),
                ui.button(name="popup_off", label="popup off"),
            ],
        )
    if q.args.single:
        q.page["example"].items[0].dropdown.value = '2'
        q.page["example"].items[0].dropdown.values = None
    if q.args.multi:
        q.page["example"].items[0].dropdown.values = ['2', '3']
    if q.args.popup_on:
        q.page["example"].items[0].dropdown.popup = 'always'
    if q.args.popup_off:
        q.page["example"].items[0].dropdown.popup = 'never'

    await q.page.save()

```

Also, checked whether changing `popup` dynamically it would work but found some bugs.

aalencar:
> The only thing that bothers me is a lot of introduced duplication, especially in the markup part, but we can leave that for future.

It also bothered me. As you said, we can leave it for another moment.

ui/src/dropdown.tsx
mturoci: I see, thanks for clarification.

ui/src/dropdown.tsx
aalencar: > This is a shallow copy, not a deep copy so this doesn't hold true - it should be affected in both scenarios.

Good point, I overlooked that. I guess we can simply pass the `items` reference.

> AFAIK it should allow you to explore the complete memory snapshot item by item so you can see what's the root cause of the extra memory.

If I wanted a detailed explanation, yes, but in this case, I just want to see the overall change to memory since the only thing changed in my code was `useRef([...items])` -> `useRef(items)`. I'm not very familiar with the profiler tool though.

> What is previous state in this case? I thought it's the initial selection that is seen after opening the dialog, but maybe I am wrong?

You are right, you are saying that it's "initial" at the moment we open the dialog. I'm saying that the word "initial" can also mean "when the component is initialized and these values will not change" and that can be confusing. Maybe rename to `snapshot`?
I believe that anyone will be able to follow the logic anyway. I'm fine with `initialSelection` as well.

mturoci: >  just want to see the overall change to memory since the only thing changed in my code was useRef([...items]) -> useRef(items)

I see. I was just wondering about the cause because IMO it should be the other way around - using reference should result in less memory.

> when the component is initialized and these values will not change" and that can be confusing

Got it. `snapshot` is still too broad. Maybe something in the lines of `itemsOnOpen` / `snapshotOnOpen` or maybe even `onDialogOpen` which is the most crystal clear possible probably :D

aalencar: `onDialogOpen` sounds like an event handler. I went with `itemsOnDialogOpen`.

ui/src/dropdown.tsx
aalencar: No problem.

mturoci (approved):
Thanks @aalencar!

PR #1628, opened 2022-09-18
fix: Multiline selection support for ui.text_annotator

marek-mihok:
The new character level annotations feature implementation has caused the inability of multiline selections. To be more precise - the text was automatically annotated when mouse went outside of text while annotating (even in the case while dragging the mouse between the lines because of the margin in between them).

This fix ensures that no annotation is made when mouse leaves the text, instead the annotation is made after `onMouseUp` event.

https://user-images.githubusercontent.com/23740173/190928824-977ca065-b9bd-4db5-bbee-1230aad2707d.mov

### DIFF
--- a/ui/src/text_annotator.tsx
+++ b/ui/src/text_annotator.tsx
@@ -284,9 +284,13 @@ export
       },
       handleMouseDown = (startElProps: TokenMouseEventProps) => () => startElPropsRef.current = startElProps,
       handleMouseUp = (endElProps: TokenMouseEventProps) => () => annotate(endElProps),
-      handleMouseLeave = (endElProps: TokenMouseEventProps) => (ev: React.MouseEvent<HTMLSpanElement>) => {
-        const nodeName = (ev.relatedTarget as any)?.nodeName
-        if (startElPropsRef.current !== undefined && nodeName !== 'P' && nodeName !== 'MARK') annotate(endElProps) // nodeName prop is not typed yet
+      // handles the selection in case mouse up event is fired when cursor is not above text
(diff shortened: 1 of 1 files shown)


ui/src/text_annotator.tsx
```diff
@@ -315,17 +319,18 @@ export
...
-        <div className={clas(css.content, 'wave-s16 wave-t7 wave-w3')}>{
+        <div className={clas(css.content, 'wave-s16 wave-t7 wave-w3')} onMouseUp={handleSelectionFinish}>{
           tokens.map(({ start, end, text, tag }, idx) => {
             const activeColor = activeTag ? tagColorMap.get(activeTag) : undefined
             return (
               // Use paragraph instead of span to support text highlight.
               <p
                 key={idx}
+                // data-keys keeps state in DOM to access it via window.getSelection() API in case mouse up event is fired when cursor is not above <p/> element
+                data-keys={`{ "key": ${idx}, "start": ${start}, "end": ${end} }`}
```
mturoci, line 330: Wouldn't this work?

```suggestion
                data-keys={{ idx, start, end }}
```

mturoci (approved):
Thanks @marek-mihok! Just 1 question.

ui/src/text_annotator.tsx
marek-mihok: It does not since dataset can be only string - https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset
The solution here is: 
- to use string representation of JSON and later parse it
- use separate `data-*` attributes for `idx`, `start` and `,end`

Issue #1049, opened 2021-10-06
Upload Component ( Enhancement ) 


sandruparo:
![image](https://user-images.githubusercontent.com/9942408/136232379-0a47e33c-4426-4148-96d2-93ae849b5436.png)

sandruparo:
@mturoci 

We have revamped the upload component also in the wave style guide. We need to enhance this component and include this in the sdk . Thanks Martin. 

cc - @shihan007 @shanaka-rajitha @dulajra @VijithaEkanayake

shihan007:
@mturoci I'll update those states. X is a close icon use to remove file if user want.

mturoci:
Overall UX is odd. Ideally we should prevent user from experiencing what he can or cannot do himself. It's better to restrict actions so that it's impossible to do the wrong thing. If such restrictions are not possible though, at least inform the user in advance.

Having all the info in a single simple label feels too much to me + this assumes app devs will handle that themselves which I doubt. Let's try a different approach - try to think about placing these restrictions somewhere, separate from the label, so that they are visible and don't require dev interaction at all - they will be auto-populated based on chosen attributes.

>We can add " select one file at a time" text to above text. And if user still chooses multiple file to upload lets display an error msg as shown " Error uploading multiple files, please select one file".

As mentioned above, it's best to restrict invalid action completely so that user doesn't need to learn "the hard way". What about hiding whole drag&drop box after a single file was uploaded?

Another issue is that displaying uploaded files like this can cause layout shifts or even scrollbars which is usually unwanted.

PR #1633, opened 2022-09-21
feat: Add required prop to ui.file_upload

### LINKED ISSUE #1503: file_upload required parameter

No

Support for `required` parameter for `ui.file_upload()`, for example like `ui.dropdown()` or `ui.combobox()`

Responding if file upload is not done by checking `q.args` **after** the form is submitted.

marek-mihok:
Adds `required` prop to `ui.file_upload` component. Updated API:
```ts
export interface FileUpload {
  name: Id
  /** Text to be displayed in the bottom button or as a component title when the component is displayed compactly. Defaults to "Upload". */
  label?: S
  /** True if the component should allow multiple files to be uploaded. */
  multiple?: B
  /** List of allowed file extensions, e.g. `pdf`, `docx`, etc. */
  file_extensions?: S[]
  /** Maximum allowed size (Mb) per file. No limit by default. */
  max_file_size?: F
  /** Maximum allowed size (Mb) for all files combined. No limit by default. */
  max_size?: F
  /** The height of the file upload, e.g. '400px', '50%', etc. Defaults to 300px. */
  height?: S
  /** The width of the file upload, e.g. '100px'. Defaults to '100%'. */
  width?: S
  /** True if the component should be displayed compactly (without drag-and-drop capabilities). Defaults to False. */
  compact?: B
  visible?: B
  tooltip?: S
  /** True if this is a required field. Defaults to False. */
  required?: B
}
```

Full-sized:
image

Compact (with label):
image

Compact (without label):
image

Closes #1503

### DIFF
--- a/ui/src/file_upload.tsx
+++ b/ui/src/file_upload.tsx
@@ -15,7 +15,7 @@
 import * as Fluent from '@fluentui/react'
 import { B, F, Id, S, U, xid } from 'h2o-wave'
 import React from 'react'
-import { stylesheet } from 'typestyle'
+import { style, stylesheet } from 'typestyle'
 import { centerMixin, clas, cssVar, dashed, padding } from './theme'
 import { wave } from './ui'
 
@@ -46,6 +46,8 @@ export interface FileUpload {
   visible?: B
   /** An optional tooltip message displayed when a user clicks the help icon to the right of the component. */
   tooltip?: S
+  /** True if this is a required field. Defaults to False. */
+  required?: B
 }
 
 const
@@ -72,9 +74,16 @@ const
       minWidth: 80,
       boxSizing: 'border-box',
       height: 32,
+      display: 'inline-block'
+    },
+    asterisk: {
       $nest: {
-        '&:hover': {
-          cursor: 'pointer'
+        '&::after': {
+          content: "'*'",
+          verticalAlign: 'top',
+          paddingLeft: '2px',
+          lineHeight: '12px',
+          position: 'absolute',
         }
       }
     },
@@ -103,13 +112,14 @@ const convertMegabytesToBytes = (bytes: F) => bytes * 1024 * 1024
 export const
   XFileUpload = ({ model }: { model: FileUpload }) => {
     const
-      { name, label, file_extensions, max_file_size, compact, height, max_size, multiple } = model,
+      { name, label, file_extensions, max_file_size, compact, height, max_size, multiple, required } = model,
       [isDragging, setIsDragging] = React.useState(false),
       [files, setFiles] = React.useState<File[]>([]),
       [fileNames, setFileNames] = React.useState<S>(''),
       [percentComplete, setPercentComplete] = React.useState(0.0),
       [error, setError] = React.useState(''),
       [successMsg, setSuccessMsg] = React.useState(''),
+      { semanticColors: { errorText: errorTextColor } } = Fluent.useTheme(),
       maxFileSizeBytes = max_file_size ? convertMegabytesToBytes(max_file_size) : 0,
       maxSizeBytes = max_size ? convertMegabytesToBytes(max_size) : 0,
       fileExtensions = file_extensions ? file_extensions.map(e => e.startsWith('.') ? e : `.${e}`) : null,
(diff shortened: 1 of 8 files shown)


ui/src/file_upload.tsx
```diff
@@ -103,13 +112,14 @@ const convertMegabytesToBytes = (bytes: F) => bytes * 1024 * 1024
...
     const
-      { name, label, file_extensions, max_file_size, compact, height, max_size, multiple } = model,
+      { name, label, file_extensions, max_file_size, compact, height, max_size, multiple, required } = model,
       [isDragging, setIsDragging] = React.useState(false),
       [files, setFiles] = React.useState<File[]>([]),
       [fileNames, setFileNames] = React.useState<S>(''),
       [percentComplete, setPercentComplete] = React.useState(0.0),
       [error, setError] = React.useState(''),
       [successMsg, setSuccessMsg] = React.useState(''),
+      { semanticColors: { errorText: errorTextColor } } = Fluent.useTheme(),
```
mturoci, line 122: We could maybe make this a CSS var (`--errorText`) and set it on theme change. Would make also future code simpler.

mturoci (changes requested):
Thanks @marek-mihok! Let's put the asterisk in compact mode next to the label (similar to how textbox/dropdown works. As for the large version, let's go with the current implementation, can be revisited in https://github.com/h2oai/wave/issues/1049.

ui/src/file_upload.tsx
marek-mihok: It is possible to achieve in two ways. Do you mean to add `semanticColors: {errorText: '#xxxxxx'}` into each theme in the `const themes: Dict<PredefinedTheme>` object in `theme.ts`? Or just get the `semanticColors.errorText` from theme after it loads/changes and then set it as css variable? 
Because if we go with second option I need to do it in `theme.ts` where theme is changed and in `index.ts` where theme is first loaded.

mturoci: >Or just get the semanticColors.errorText from theme after it loads/changes and then set it as css variable?

this one

ui/src/file_upload.tsx
```diff
@@ -72,9 +74,17 @@ const
       minWidth: 80,
       boxSizing: 'border-box',
       height: 32,
+      display: 'inline-block'
+    },
+    asterisk: {
       $nest: {
-        '&:hover': {
-          cursor: 'pointer'
```
mturoci, line 77: This should stay in `uploadLabel` class to get the proper UX.

ui/src/file_upload.tsx
```diff
@@ -326,9 +338,9 @@ export const
               )
               : (
                 <>
-                  {label && <Fluent.Label style={{ paddingTop: 6 }}>{label}</Fluent.Label>}
+                  {label && <Fluent.Label style={{ paddingTop: 6 }} required={required}>{label}</Fluent.Label>}
                   <div className={css.compact}>
-                    <Fluent.TextField data-test={`textfield-${name}`} readOnly value={fileNames} errorMessage={error} />
+                    <Fluent.TextField data-test={`textfield-${name}`} readOnly value={fileNames} errorMessage={error} required={label ? undefined : required} />
```
mturoci, line 343: What does `required` do in `Textfield`? Isn't including it on `Label` enough?

mturoci (changes requested):
2 more questions and we are good to go.

ui/src/file_upload.tsx
marek-mihok: Thanks for noticing!

ui/src/file_upload.tsx
marek-mihok: @mturoci it is not enough, since `label` is just an optional prop. If label is not present, the asterisk shoud be displayed right to the input. The same behavior is implemented for other form components so I can't see any reason to make this component inconsistent with others. 
image

mturoci: > The same behavior is implemented for other form components

Didn't know that! Agree then, let's leave it as is.

mturoci (approved):

PR #1635, opened 2022-09-22
feat: Add proper err handling to site.download function and allow extensionless file downloads. #1627

### LINKED ISSUE #1627: Uploads fail if file has no extension
#### Wave SDK Version, OS

0.23.0, Mac OSX

#### Actual behavior

Uploaded an AWS credentials file using `ui.file_upload`, file gets created on wave server with text "Not Found".

Logs show when I try to download the file:
```
2022/09/16 17:07:33 # {"error":"not found","path":"/_f/80b79c0e-5685-42df-8c2c-450a19f60689/credentials","t":"file_download"}
```

#### Expected behavior

Ability to upload files without extensions.

mturoci:
* Raise an err if there is any - same behavior like the rest of alike functions.
* Create a file for the stream to write into only in case of 200 status code.

Needs discussion: https://github.com/h2oai/wave/issues/1627#issuecomment-1254702592

Closes #1627

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -40,19 +40,6 @@
       ]
     },
     {
-      "name": "Debug PY App",
-      "type": "python",
-      "request": "launch",
-      "program": "${workspaceFolder}/py/venv/bin/wave",
-      "python": "${workspaceFolder}/py/venv/bin/python",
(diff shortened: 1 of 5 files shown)


lo5 (approved):
Thanks @mturoci.

PR #1640, opened 2022-09-28
feat: Add vertical alignment to ui.inline

### LINKED ISSUE #1535: Allow vertical alignment for ui.inline
via @azim-b

Add `align` attr to `ui.inline` to allow for vertical alignment. Currently, we [center everything by default](https://github.com/h2oai/wave/blob/989d2c808f0bb60f7e179bb2a7a9ccb44ad6064f/ui/src/form.tsx#L201), which does not suit every case, e.g. displaying 2 choice groups side by side with different choice counts:

image

## Implementation notes

* Add a new `align` prop to `Inline` interface with all the possible values that flexbox offers for vertical alignment.
* Implement the necessary functionality.
* Add a python example demonstrating the behavior.

NevilleMthw:
Hi, this resolves #1535.

Thanks,
Neville


mturoci:
Thanks for the PR @NevilleMthw! Unfortunately, this is not what is needed to be done. I have updated the issue description, so please have a look. I also suggest going through contribution guidelines that describe what needs to be done when adding new props to cards/components.

NevilleMthw:
Oh okay. Thanks for the feedback. I will go through the issue again. @mturoci.

mturoci (changes requested):
Thanks for the PR @NevilleMthw! Can you also please include python examples showcasing your feature + update PR description with the screenshots so that we can see how does the feature works in practice.

NevilleMthw:
> Thanks for the PR @NevilleMthw! Can you also please include python examples showcasing your feature + update PR description with the screenshots so that we can see how does the feature works in practice.

Hi, I will definitely look at trying to create a Python example. Will let you know. Thanks.

mturoci:
Closing due to inactivity.

PR #1658, opened 2022-10-05
docs: Add social media icons to header RHS.

### LINKED ISSUE #1653: Add social icons to docs page header
Replace the current Github link with an icon and include icons for Twitter and Discord. Replace the theme-changing icon with a more compact one.

## Current

image

## Desired result

image

inspiration from - https://www.frontendinterviewhandbook.com/

## Links

Github - https://github.com/h2oai/wave
Twitter - https://twitter.com/intent/follow?screen_name=h2o_wave
Discord - https://discord.gg/V8GZFAy3WM

i-m-karanagarwal:
@mturoci Please review my PR.
Added Social Icons to Docs Page header.

Closes #1653


website/docusaurus.config.js
```diff
@@ -41,7 +41,25 @@ module.exports = {
         { to: 'https://github.com/h2oai/wave/discussions', label: 'Discuss', position: 'left' },
         { to: 'https://h2oai.github.io/h2o-ai-cloud/', label: 'Enterprise', position: 'left', },
         { to: 'https://github.com/h2oai/wave/releases', label: 'Download', position: 'right', },
-        { href: 'https://github.com/h2oai/wave', label: 'GitHub', position: 'right', },
+        { href: 'https://github.com/h2oai/wave',
+          position: 'right',
+          className: 'navbar-icon navbar-icon-github',
```
mturoci, line 46: Are `navbar-icon-<social-media>` defined somewhere? If not, let's remove those. Seems like only `navbar-icon` class is used.

Same for the rest of the icons.

```suggestion
          className: 'navbar-icon',
```

website/docusaurus.config.js
```diff
@@ -41,7 +41,25 @@ module.exports = {
...
+          'aria-label': 'GitHub repository',
+          html: `<img src="img/github.svg"/>`,},
```
mturoci, line 48: We should use absolute URLs instead of relative ones. Notice how the icons break when you navigate to `Get started` page for example.

Also please add an `alt` attribute to the icons so that they are accessible.

mturoci (changes requested):
Thanks @i-m-karanagarwal! Looks great.

A few comments from me + please remove all the changes to `package-lock.json` - these should not be there.

website/docusaurus.config.js
```diff
@@ -41,7 +41,25 @@ module.exports = {
         { to: 'https://github.com/h2oai/wave/discussions', label: 'Discuss', position: 'left' },
         { to: 'https://h2oai.github.io/h2o-ai-cloud/', label: 'Enterprise', position: 'left', },
         { to: 'https://github.com/h2oai/wave/releases', label: 'Download', position: 'right', },
-        { href: 'https://github.com/h2oai/wave', label: 'GitHub', position: 'right', },
+        { href: 'https://github.com/h2oai/wave',
+          position: 'right',
+          className: 'navbar-icon',
+          'aria-label': 'GitHub repository',
+          html: `<img src="https://img.icons8.com/ios-glyphs/30/000000/github.png" alt='Github Logo' />`,},
```
mturoci, line 48: Although this solves the problem, what I meant by absolute URLs was to be absolute within our docs so all there is needed to be done was to prefix your image src with `/`. See [third example here](https://developer.mozilla.org/en-US/docs/Learn/Common_questions/What_is_a_URL#examples_of_absolute_urls). Please revert and only prefix the urls.

Having the SVG icons served from our own server is better performance-wise since we do not need to make another TLS handshake.

mturoci (approved):

Issue #1599, opened 2022-08-19
Carousel and Image Gallery / Image Grid


dott1718:
Current implementations to nicely show multiple images or galleries in wave apps involve quite a bit of javascript and/or HTML & CSS usage. It would be helpful to further abstract that functionality to the wave framework and to add common image displays to speed up the app development process. It would also help to prevent speed issues and to align UX across wave apps.

We are looking for either a Carousel/Lightbox style implementation (https://www.joomforest.com/joomla-extensions/jf-responsive-carousel) or an Image Grid (https://www.w3schools.com/howto/howto_js_image_grid.asp) 

Requirements:
- Customizable in image width/height, number of images per row/column
- Should take a list of image paths as input (open for discussion)
- Automatic resizing for speed improvement
- Responsive to clicks, e.g. trigger a reload and generate an args id or event
- Add mouse hover tooltips per image
- (lower priority) ability to show samples from image annotator with annotations (with mouse hover tooltips to show the corresponding classes to the boxes)
- (optional) pre-caching

mturoci:
Some designs can be found [here](https://xd.adobe.com/view/acfb7ef1-5b99-4129-97b2-b30f8086c239-f9a8/). Let's discuss the API prior to implementation. cc @lo5

aalencar:
### Proposal

Develop two separate components, Carousel Card and Image Grid Card.
Both open a lightbox (modal with image) when an image is clicked. We could also somehow integrate Image Grid with Carousel Card by using a "View All" button that opens a modal with Image Grid.

### Proposed API for Carousel Card

```ts
export interface State {
  /** An identifying name for this component. */
  name: Id
  /** List of images to be displayed. */
  images: Image[]
  /** True if the current image should be submitted when the image changes. */
  trigger?: B
}

// Similar to Image Card but with description
/** Create an image. */
export interface Image {
  /** The image title, displayed under the image. */
  title: S
  /** The image description, displayed under the title. */
  description: S
  /** The image MIME subtype. One of `apng`, `bmp`, `gif`, `x-icon`, `jpeg`, `png`, `webp`. Required only if `image` is set. */
  type?: S
  /** Image data, base64-encoded. */
  image?: S
  /** The path or URL or data URL of the image, e.g. `/foo.png` or `http://example.com/foo.png` or `data:image/png;base64,???`. */
  path?: S
  /** The width of the image, e.g. '100px'. */
  width?: S
  /** True if the component should be visible. Defaults to true. */
[...]

PR #1662, opened 2022-10-06
feat: Image lightbox component for image zoom and image grid

### LINKED ISSUE #1542: Image zoom when clicking on an image
On some websites, if you click on an image, it will blow up to fullscreen mode (with an "x" to exit the view). Maybe we can add this functionality to Wave?

marek-mihok:
New image lightbox component (not public facing API) for (almost) full-screen image view. 
Appears when **ui.image** is clicked.

```tsx
interface LightboxProps {
  visible: B,
  onDismiss: () => void,
  images: Image[],
  defaultImageIdx?: U
}
```
**For multi-image view:**

https://user-images.githubusercontent.com/23740173/194434539-6eb4d8c9-0fac-4e7a-82ad-bf749bf4ea29.mov

To be implemented with **ui.image_grid** component (#1599)

**For single image:**

https://user-images.githubusercontent.com/23740173/194434533-7edcdaf9-6555-4521-86a4-61461ee9de2a.mov

Closes #1542

### DIFF
--- a/ui/src/parts/lightbox.tsx
+++ b/ui/src/parts/lightbox.tsx
@@ -40,11 +40,11 @@ const
         img: {
             // flexGrow: 1,
             maxWidth: '100%', // TODO: 3000x2000 vertical scrollbar
-            alignSelf: 'center',
-            // objectFit: 'scale-down',
-            transformOrigin: 'top left',
-            transition: 'transform 0.25s ease',
-            cursor: 'pointer',
+            // alignSelf: 'center',
+            objectFit: 'scale-down',
+            // transformOrigin: 'top left',
+            // transition: 'transform 0.25s ease',
+            // cursor: 'pointer',
         },
         header: {
             width: '100%',
@@ -108,7 +108,7 @@ const
             overflow: 'hidden',
         },
         title: { fontWeight: 500 },
-        description: { opacity: 0.85 }, // TODO: use proper color instead of opacity
+        description: { color: '#bbbbbb' }, // TODO: theme colors
         navImgPlaceholder: { display: 'inline-block', width: '124px' }
     })
 
@@ -125,49 +125,44 @@ interface Props {
     defaultImageIdx?: U
 }
 
-const
-    zoomLevels: { [key: U]: U } = { 0: 1, 1: 1.5, 2: 2, 3: 3, 4: 4 },
-    lazyImageObserver = new IntersectionObserver(entries => {
-        entries.forEach(entry => {
-            if (entry.isIntersecting) {
-                const lazyImage = entry.target as HTMLImageElement
-                lazyImage.src = lazyImage.dataset.src!
-                lazyImage.classList.remove("lazy")
-                lazyImageObserver.unobserve(lazyImage)
-            }
-        })
+const lazyImageObserver = new IntersectionObserver(entries => {
+    entries.forEach(entry => {
+        if (entry.isIntersecting) {
+            const lazyImage = entry.target as HTMLImageElement
+            lazyImage.src = lazyImage.dataset.src!
+            lazyImage.classList.remove("lazy")
+            lazyImageObserver.unobserve(lazyImage)
+        }
     })
+})
 
 export const Lightbox = ({ visible, onDismiss, images, defaultImageIdx }: Props) => {
     const
         [activeImageIdx, setActiveImageIdx] = React.useState(defaultImageIdx || 0),
-        [zoomLevel, setZoomLevel] = React.useState(0),
         imageNavRef = React.useRef<HTMLDivElement | undefined>(),
         handleClickLeft = () => {
             const nextImgIdx = (activeImageIdx === 0) ? (images.length - 1) : activeImageIdx - 1
             setActiveImageIdx(nextImgIdx)
-            setZoomLevel(0)
             if (imageNavRef.current) {
                 const isLeft = activeImageIdx <= Math.floor(imageNavRef.current.scrollLeft / 124)
                 if (nextImgIdx === images.length - 1) imageNavRef.current.scrollLeft = imageNavRef.current.scrollWidth - imageNavRef.current?.clientWidth
-                else if (isLeft) imageNavRef.current.scrollLeft = (activeImageIdx * 124) - 124
+                else if (isLeft) imageNavRef.current.scrollBy({ left: (activeImageIdx * 124) - 124 - imageNavRef.current.scrollLeft, top: 0, behavior: 'smooth' })
             }
         },
         handleClickRight = () => {
             const nextImgIdx = (activeImageIdx === images.length - 1) ? 0 : activeImageIdx + 1
             setActiveImageIdx(nextImgIdx)
-            setZoomLevel(0)
             if (imageNavRef.current) {
                 const pageImageCount = Math.floor(imageNavRef.current?.clientWidth / 124)
                 const isRight = activeImageIdx >= Math.floor(imageNavRef.current.scrollLeft / 124) + (pageImageCount - 1)
                 if (nextImgIdx === 0) imageNavRef.current.scrollLeft = 0
-                else if (isRight) imageNavRef.current.scrollLeft = ((activeImageIdx + 1 - pageImageCount) * 124) + 124
+                else if (isRight) imageNavRef.current.scrollBy({ left: ((activeImageIdx + 1 - pageImageCount) * 124) + 124 - imageNavRef.current.scrollLeft, top: 0, behavior: 'smooth' })
             }
         },
         onClose = () => {
-            setZoomLevel(0)
             onDismiss()
             setActiveImageIdx(defaultImageIdx || 0)
+            if (imageNavRef.current) imageNavRef.current.scrollLeft = 0
         }
 
     React.useLayoutEffect(() => {
@@ -200,14 +195,8 @@ export const Lightbox = ({ visible, onDismiss, images, defaultImageIdx }: Props)
                     alt={title}
                     src={src}
                     style={{
-                        transform: `scale(${zoomLevels[zoomLevel]})`, // TODO:
-                        cursor: zoomLevel < 4 ? 'zoom-in' : 'zoom-out',
-                        maxHeight: 'inherit',
+                        // maxHeight: 'inherit',
                         // objectFit: 'none'
-                    }}
-                    onClick={(_ev) => {
-                        const zoom = zoomLevel === 4 ? 0 : zoomLevel + 1
-                        setZoomLevel(zoom)
                     }}
                 />
                 {images.length > 1 &&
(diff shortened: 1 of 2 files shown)


mturoci (changes requested):
Thanks @marek-mihok the functionality looks great!

In addition to my comments:

* Does this work cross-browser?
* Since we can assume there is always going to be just a single lightbox at time, let's treat it in the similar fashion as Dialog, SidePanel etc. - single component per app which is displayed if it gets data.
* All `_setActiveImageIdx` calls should be handled via callback since the next value is dependent on the prev one. See https://reactjs.org/docs/hooks-reference.html#functional-updates.
* Add a simple unit test that verifies the dialog is open after clicking `ui.image`.

ui/src/parts/lightbox.tsx
```diff
@@ -0,0 +1,229 @@
...
+      _setActiveImageIdx(idx)
+      activeImageIdxRef.current = idx
+    },
+    onClose = () => {
+      onDismiss()
+      setActiveImageIdx(defaultImageIdx || 0)
+      if (imageNavRef.current) imageNavRef.current.scrollLeft = 0
+    },
+    handleKeyDown = (ev: KeyboardEvent) => {
+      if (ev.keyCode === keys.esc) onClose()
```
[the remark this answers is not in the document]
marek-mihok: mdn website [suggests](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) using `KeyboardEvent.code` instead which is new (implemented in 2020 for Edge), that was the reason for keeping a `keyCode`. But the alternative you suggests is fine.

ui/src/parts/lightbox.tsx
```diff
@@ -0,0 +1,229 @@
...
+    if (visible) window.addEventListener("keydown", handleKeyDown)
+    else window.removeEventListener("keydown", handleKeyDown)
+    return () => window.removeEventListener("keydown", handleKeyDown)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [visible])
+
+  // Handle image navigation scroll.
+  React.useEffect(() => {
+    if (isGallery) navImgRefs[activeImageIdx].current!.scrollIntoView({
+      behavior: (activeImageIdx === 0 || activeImageIdx === images.length - 1) ? 'auto' : 'smooth',
```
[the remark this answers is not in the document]
marek-mihok: Yes, everything what's implemented is tested in recent versions of Chrome, Firefox and Safari and it works.

ui/src/parts/lightbox.tsx
```diff
@@ -0,0 +1,229 @@
...
+      height: '100%',
+      zIndex: 1,
+      backgroundColor: 'rgba(0, 0, 0, 0.9)',
+    },
+    img: {
+      maxWidth: '100%',
+      objectFit: 'scale-down',
+    },
+    header: {
+      width: '100%',
```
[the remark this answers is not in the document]
marek-mihok: Not in this point, but it was with previous implementations. Thanks for noticing.

ui/src/parts/lightbox.tsx
```diff
@@ -0,0 +1,229 @@
...
+    },
+    arrow: {
+      cursor: "pointer",
+      position: "absolute",
+      top: "50%",
+      width: "auto",
+      padding: "16px",
+      marginTop: "-50px",
+      color: "#fff",
+      userSelect: "none",
```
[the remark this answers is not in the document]
marek-mihok: Not necessary anymore since I am using fluent icons.

ui/src/parts/lightbox.tsx
```diff
@@ -0,0 +1,229 @@
...
+      height: '120px',
+      width: '120px',
+      margin: '0px 2px',
+      cursor: 'pointer',
+      filter: 'brightness(30%)',
+      border: '2px solid black',
+      $nest: { '&:hover': { filter: 'unset', border: '2px solid red' } }
+    },
+    arrow: {
+      cursor: "pointer",
```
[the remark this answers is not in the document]
marek-mihok: The same situation here. A leftover from non-fluent implementation.

ui/src/parts/lightbox.tsx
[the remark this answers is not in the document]
mturoci: >that was the reason for keeping a keyCode

This is not a good approach since this would stop working with the new browsers in the future.

ui/src/parts/lightbox.tsx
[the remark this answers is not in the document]
mturoci: image

The smooth behavior is only supported in Safari version from last month which is not enough. Any alternatives?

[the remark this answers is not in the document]
mturoci: The most modern feature we use without polyfills is CSS variables, which are supported ~ 5 years back across all browsers so it's good to comply with that in general. However, since this particular feature is about the `smooth` scroll which is more like a nice-to-have rather than a hard requirement, your `scrollBy` support matrix is fine as long as the main feature (the element shall be scrolled into view) will work even if snappy instead of smooth.

marek-mihok:
> Since we can assume there is always going to be just a single lightbox at time, let's treat it in the similar fashion as Dialog, SidePanel etc. - single component per app which is displayed if it gets data.

Do you mean to use component pre-loading in `meta.tsx` similar to Dialog and SidePanel?

mturoci:
> do you mean to use component pre-loading in meta.tsx similar to Dialog and SidePanel?

Nope, since this will have nothing in common with `meta_card`. I want to have a single Lightbox component per app same as [dialog/sidepanel/notification bar](https://github.com/h2oai/wave/blob/master/ui/src/app.tsx#L117-L125).

ui/src/parts/lightbox.tsx
```diff
@@ -0,0 +1,224 @@
...
+      width: '120px',
+      margin: '0px 2px',
+      cursor: 'pointer',
+      filter: 'brightness(30%)',
+      border: '2px solid #000'
+    },
+    imgCaptions: {
+      whiteSpace: 'nowrap',
+      padding: '10px 40px',
+      height: IMAGE_CAPTIONS_HEIGHT - 20
```
mturoci, line 73: What's this magic number `20`?

ui/src/parts/lightbox.tsx
```diff
@@ -0,0 +1,224 @@
...
+    },
+    title: {
+      fontWeight: 500
+    },
+    description: {
+      color: '#bbb'
+    },
+    navImgContainer: {
+      display: 'inline-block',
+      width: '124px'
```
mturoci, line 87: Why `124`?

ui/src/parts/lightbox.tsx
```diff
@@ -0,0 +1,224 @@
...
+      </div>
+      <div className={css.content} style={{ height: `calc(100% - ${HEADER_HEIGHT + FOOTER_HEIGHT}px)` }}>
+        <img className={css.img} alt={title} src={getImageSrc(images[activeImageIdx])} />
+        {isGallery &&
+          <>
+            <Fluent.ActionButton
+              styles={iconStyles}
+              className={css.iconStylesRootArrow}
+              style={{ left: 0 }}
+              onClick={() => setActiveImageIdx(activeImageIdx === 0 ? images.length - 1 : activeImageIdx - 1)}
```
mturoci, line 188: Since next state for `activeImageIdx` depends on the previous one, let's use the value provided by the setState callback rather than the closure one as we discussed. Same for the other button.

ui/src/parts/lightbox.tsx
```diff
@@ -0,0 +1,224 @@
...
+          lazyImage.src = lazyImage.dataset.src!
+          lazyImage.classList.remove("lazy")
+          lazyImageObserver.unobserve(lazyImage)
+        }
+      })
+    ),
+    handleKeyDown = (ev: KeyboardEvent) => {
+      if (['Escape', 'ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown'].indexOf(ev.key) >= 0) ev.preventDefault()
+      if (ev.key === 'Escape') lightboxB(null)
+      if (ev.key === 'ArrowRight') setActiveImageIdx(prevIdx => prevIdx === images.length - 1 ? 0 : prevIdx + 1)
```
mturoci, line 138: We can probably use `else if` here, right?
```suggestion
      else if (ev.key === 'ArrowRight') setActiveImageIdx(prevIdx => prevIdx === images.length - 1 ? 0 : prevIdx + 1)
```

ui/src/parts/lightbox.tsx
```diff
@@ -0,0 +1,224 @@
...
+      else if (ev.key === 'ArrowLeft') setActiveImageIdx(prevIdx => prevIdx === 0 ? images.length - 1 : prevIdx - 1)
+    }
+
+  React.useEffect(() => {
+    // Initialize intersection observer for lazy images.
+    const lazyImages = [].slice.call(document.querySelectorAll(".lazy"))
+    lazyImages.forEach(lazyImage => lazyImageObserver.observe(lazyImage))
```
mturoci, line 145: Would this work?
```suggestion
    document.querySelectorAll(".lazy").forEach(lazyImage => lazyImageObserver.observe(lazyImage))
```

ui/src/parts/lightbox.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+    // Add keyboard events listener.
+    window.addEventListener("keydown", handleKeyDown)
+    return () => window.removeEventListener("keydown", handleKeyDown)
```
mturoci, line 149: Let's try to avoid registering global event listeners. Can these be moved to the component root (wrapper) element?

mturoci (changes requested):
Thanks @marek-mihok, a few more comments.

ui/src/parts/lightbox.tsx
marek-mihok: Isn't it self-explanatory when there is a top/bottom 10px padding?

mturoci: You see it's not since I am asking about it :D Have a look at `box-sizing` CSS prop, that's the proper way to do this instead of calculations.

ui/src/parts/lightbox.tsx
marek-mihok: Works! Thanks

ui/src/parts/lightbox.tsx
marek-mihok: I was not sure if you don't know or you just want to make code even better. Thanks for the `box-sizing` prop, that is what I was looking for.

ui/src/parts/lightbox.tsx
marek-mihok: I was trying to achieve that during the initial implementation, but I had troubles with keeping focus on the component root element. That's why I've sticked with global handlers in the end. But I am open to suggestions. Or we can have a short call.

mturoci (approved):
LGTM!

Issue #1659, opened 2022-10-05
Alignment of table values


vopani:
#### Is your feature request related to a problem? Please describe
Allowing different alignments of values in a table can beautify UI in some use-cases.

#### Describe the solution you'd like
Allow user to left / center / right align the values of a column (something like `justify` in `ui.table_column` with default='left' so nothing breaks / changes in UI of existing apps).

#### Additional context
This is purely for nicer UI, especially in scenarios when some of the table cell elements are markdown (where content can be center-aligned), it'll make it more consistent visually.

Pooja444:
@mturoci I would like to work on this issue, can it be assigned to me, please? I have worked on the issue https://github.com/h2oai/wave/issues/1577 and now I have a good knowledge of how to develop UI changes so it'll probably help with this issue 😁

mturoci:
Please go ahead @Pooja444!

RedJOe-0608:
I'd like to contribute to this issue! Can I work on it?

mturoci:
Hi @RedJOe-0608, this issue is currently taken by @Pooja444 but feel free to pick some other issue - there are still many left.

Pooja444:
Thanks for assigning this to me @mturoci, I have raised a PR for this issue - https://github.com/h2oai/wave/pull/1670
This issue was awesome and I had fun contributing to it! Please review and let me know if I need to make any changes, thank you!

PR #1670, opened 2022-10-14
feat: align table column values

Pooja444:
## Github Issue

Closes https://github.com/h2oai/wave/issues/1659

## Description

- Added an option to the `ui.table` component to allow users to specify alignment per column.
- Alignment values could be one of ['left', 'center', 'right']
- Added new example `table_column_alignment.py` so that original `table.py` remains unchanged
- Added this new feature to `table.md`
- Ran `make generate` before raising PR

## Successful local testing

Screen Shot 2022-10-13 at 6 11 55 PM

### DIFF
--- a/ui/src/table.tsx
+++ b/ui/src/table.tsx
@@ -203,6 +203,7 @@ type PaginationProps = {
 const
   // TODO: Clean up into correct Fluent style slots.
   css = stylesheet({
+    // HACK: Put sorting icon on right (same as filter).
     sortableHeader: {
       $nest: {
         '.ms-DetailsHeader-cellName': {
@@ -462,16 +463,15 @@ const
         wave.args[m.name] = [item.key as S]
         wave.push()
       },
-      getItemColumnStyle = (item?: Fluent.IObjectWithKey & Dict<any>, _idx?: U, col?: WaveColumn) => {
+      getCellComponent = (item?: Fluent.IObjectWithKey & Dict<any>, _idx?: U, col?: WaveColumn) => {
         if (!item || !col) return <span />
 
         const TooltipWrapper = ({ children }: { children: S }) => {
-          const align = col.align || 'left'
           if (col.cellOverflow === 'tooltip') return (
             <Fluent.TooltipHost
               id={item.key as S}
               // HACK: prevent Safari from showing a default tooltip - https://github.com/microsoft/fluentui/issues/13868
-              styles={{ root: { '::after': { content: '', display: 'block' }, display: 'flex', justifyContent: align, textAlign: align } }}
+              styles={{ root: { '::after': { content: '', display: 'block' } } }}
               content={children}
               overflowMode={Fluent.TooltipOverflowMode.Parent}
               title={children}
@@ -508,14 +508,8 @@ const
         return <TooltipWrapper>{v}</TooltipWrapper>
       },
       onRenderItemColumn = (item?: Fluent.IObjectWithKey & Dict<any>, _idx?: U, col?: WaveColumn) => {
-        if (!item || !col) return <span />
-
-        const AlignmentWrapper = ({ children }: { children: any }) => {
-          const align = col.align || 'left'
-          return <div style={{ display: 'flex', justifyContent: align, textAlign: align }}>{children}</div>
-        }
-
-        return <AlignmentWrapper>{getItemColumnStyle(item, _idx, col)}</AlignmentWrapper>
+        const align = col?.align || 'left'
+        return <div style={{ display: 'flex', justifyContent: align, textAlign: align }}>{getCellComponent(item, _idx, col)}</div>
       },
       // HACK: fixed jumping scrollbar issue when scrolling into the end of list with all groups expanded - https://github.com/microsoft/fluentui/pull/5204 
       getGroupHeight = (group: Fluent.IGroup) => {
--- a/ui/src/table.tsx
+++ b/ui/src/table.tsx
@@ -73,7 +73,7 @@ interface TableColumn {
   /** List of values to allow filtering by, needed when pagination is set. Only applicable to filterable columns. */
   filters?: S[]
   /** Defines how to align values in a column. */
-  alignment?: 'left' | 'center' | 'right' | 'justify'
+  alignment?: 'left' | 'center' | 'right'
 }
 
 /** Create a table row. */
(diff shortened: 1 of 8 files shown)


mturoci (changes requested):
Thanks @Pooja444! A few comments from me.

Pooja444:
@mturoci thanks a lot for these suggestions! I have updated the PR as per your comments, kindly review and let me know if the changes look good

ui/src/table.tsx
```diff
@@ -486,16 +490,33 @@ const
...
       },
+      onRenderItemColumn = (item?: Fluent.IObjectWithKey & Dict<any>, _idx?: U, col?: WaveColumn) => {
+        if (!item || !col) return <span />
+
+        const AlignmentWrapper = ({ children }: { children: any }) => {
+          const align = col.align || 'left'
+          return <div style={{ display: 'flex', justifyContent: align, textAlign: align }}>{children}</div>
+        }
+
+        return <AlignmentWrapper>{getItemColumnStyle(item, _idx, col)}</AlignmentWrapper>
```
mturoci, line 518: Almost there, no need for separate component if it's used just once anyway. The simpler the code the better.

```suggestion
        const align = col.align || 'left'
        return <div style={{ display: 'flex', justifyContent: align, textAlign: align }}>{getItemColumnStyle(item, _idx, col)}</div>
```

ui/src/table.tsx
```diff
@@ -459,15 +462,16 @@ const
         wave.args[m.name] = [item.key as S]
         wave.push()
       },
-      onRenderItemColumn = (item?: Fluent.IObjectWithKey & Dict<any>, _idx?: U, col?: WaveColumn) => {
+      getItemColumnStyle = (item?: Fluent.IObjectWithKey & Dict<any>, _idx?: U, col?: WaveColumn) => {
         if (!item || !col) return <span />
```
mturoci, line 466: We know that both `item` and `col` are truthy at this point thanks to a check in `onRenderItemColumn`. No need for an additional check.

Nit: Naming could be better since we are not returning the style but component.

```suggestion
      getCellComponent = (item: Fluent.IObjectWithKey & Dict<any>,col: WaveColumn,  _idx?: U) => {
```

ui/src/table.tsx
```diff
@@ -200,7 +203,6 @@ type PaginationProps = {
 const
   // TODO: Clean up into correct Fluent style slots.
   css = stylesheet({
-    // HACK: Put sorting icon on right (same as filter).
```
mturoci, line 203: Let's put this comment back.

ui/src/table.tsx
```diff
@@ -486,16 +490,33 @@ const
...
             wave.push()
           }
-          return <Fluent.Link onClick={onClick}><TooltipWrapper>{v}</TooltipWrapper></Fluent.Link>
+          const align = col.align || 'left'
+          return (
+            <Fluent.Link onClick={onClick}>
+              <div style={{ display: 'flex', justifyContent: align, textAlign: align }}>
+                <TooltipWrapper>{v}</TooltipWrapper>
+              </div>
+            </Fluent.Link>)
```
mturoci, line 505: Is the `align` here necessary?

ui/src/table.tsx
```diff
@@ -459,15 +462,16 @@ const
...
         if (!item || !col) return <span />
 
         const TooltipWrapper = ({ children }: { children: S }) => {
+          const align = col.align || 'left'
           if (col.cellOverflow === 'tooltip') return (
             <Fluent.TooltipHost
               id={item.key as S}
               // HACK: prevent Safari from showing a default tooltip - https://github.com/microsoft/fluentui/issues/13868
-              styles={{ root: { '::after': { content: '', display: 'block' } } }}
+              styles={{ root: { '::after': { content: '', display: 'block' }, display: 'flex', justifyContent: align, textAlign: align } }}
```
mturoci, line 474: Let's keep the tooltip content always left-aligned. Please revert.

mturoci (changes requested):
Thanks @Pooja444! Slowly getting there. A few more minor comments from me.

ui/src/table.tsx
Pooja444: Thanks for this suggestion, I removed the undefined check for `item` and `col` from `onRenderItemColumn` instead of removing it from `getCellComponent`. If I remove it from `getCellComponent` I would have to keep doing null checks for both `item` and `col` throughout `getCellComponent`, how does that sound?

ui/src/table.tsx
Pooja444: Yes, if I remove this, then there are some values in the 1st column (I checked this using the `table.py` example) that are not correctly aligned.

For example, if I remove this styling, then the center alignment looks something like this on the 1st column (the red ones are incorrect) - 

Screen Shot 2022-10-18 at 8 55 26 PM


Right alignment would look like this (the red ones are incorrect) - 

Screen Shot 2022-10-18 at 8 55 42 PM

Pooja444:
Thanks for your review @mturoci! I have addressed your review comments, please let me know if the changes look good.

ui/src/table.tsx
mturoci: Sounds good.

> I would have to keep doing null checks for both item and col throughout getCellComponent

Not necessarily. You would just need to remove optional types in `getCellComponent` and make them required (since you already checked them in `onRenderItemColumn`). However, leaving them optional and checking just in `getCellComponent` is fine too. Thanks!

ui/src/table.tsx
mturoci: Got it. In that case, we probably need to do the same also for `return <TooltipWrapper>{v}</TooltipWrapper>`, right?

Pooja444: I don't think we need to do that because once `<TooltipWrapper>{v}</TooltipWrapper>` is returned from `getCellComponent`, it is getting wrapped with the alignment logic in `onRenderItemColumn`. Let me know if I need to change this

mturoci: I see. The reason why it's needed here is that `Fluent.Link` has a style of `textAlign: 'left'` which overrides your flexbox solution. I fixed it there instead of adding an additional wrapper.

The `return <TooltipWrapper>{v}</TooltipWrapper>` has no such style thus it's not necessary as you correctly pointed out. Thanks!

mturoci (approved):
Thanks @Pooja444, great job! LGTM.

PR #1674, opened 2022-10-21
fix: Add support for unicode in table export

### LINKED ISSUE #1626: Download table content with icons as CSV

I was using default download feature in the wave tables. In my table it contains some icons which I need to display in downloaded CSV file as it is. But it shows some other characters in cells where those icons placed. 

Table UI:
image

Expected cell in downloaded CSV:
image

Currently receiving output:
image

Contain icons used in wave table cells as it is in the downloaded CSV file. 

Use separate button to download CSV after including meaningful text instead of icons. But this not includes icons as it is. Just replacing icons with meaningful names. 

#### Additional context

Discussion is in https://github.com/h2oai/wave/discussions/1624
cc: @mturoci

JacksonIsaac:
Fixes #1626 

### Example Output

image

### DIFF
--- a/ui/src/table.tsx
+++ b/ui/src/table.tsx
@@ -584,6 +584,8 @@ const
           dataRows = (m.groups ? m.groups.flatMap(({ rows }) => rows) : m.rows)?.map(({ cells }) => cells) || [],
           data = toCSV([m.columns.map(({ label, name }) => label || name), ...dataRows]),
           a = document.createElement('a'),
+          // Add BOM prefix to data. This is required to export unicode characters in the table correctly
+          // Reference: https://stackoverflow.com/a/18251283/1970068
           blob = new Blob(['\uFEFF', data], { type: "text/csv;charset=utf-8," }),
           url = window.URL.createObjectURL(blob)
 
--- a/ui/src/table.tsx
+++ b/ui/src/table.tsx
@@ -584,7 +584,7 @@ const
           dataRows = (m.groups ? m.groups.flatMap(({ rows }) => rows) : m.rows)?.map(({ cells }) => cells) || [],
           data = toCSV([m.columns.map(({ label, name }) => label || name), ...dataRows]),
           a = document.createElement('a'),
-          blob = new Blob([data], { type: "text/csv;charset=utf-8," }),
+          blob = new Blob(['\uFEFF', data], { type: "text/csv;charset=utf-8," }),
           url = window.URL.createObjectURL(blob)
 
         a.href = url
--- a/ui/src/table.tsx
+++ b/ui/src/table.tsx
@@ -584,7 +584,7 @@ const
           dataRows = (m.groups ? m.groups.flatMap(({ rows }) => rows) : m.rows)?.map(({ cells }) => cells) || [],
           data = toCSV([m.columns.map(({ label, name }) => label || name), ...dataRows]),
           a = document.createElement('a'),
-          blob = new Blob([data], { type: "octet/stream" }),
+          blob = new Blob([data], { type: "text/csv;charset=utf-8," }),
           url = window.URL.createObjectURL(blob)
 
         a.href = url


mturoci (changes requested):
Thanks for the PR @JacksonIsaac! Unfortunately, it doesn't work as proposed. I am getting this after table download (status col):

image

instead of

image

JacksonIsaac:
> Thanks for the PR @JacksonIsaac! Unfortunately, it doesn't work as proposed. I am getting this after table download (status col):
> 
> image
> 
> instead of
> 
> image

@mturoci Can you please describe the steps to reproduce this exact scenario ?

I followed https://github.com/h2oai/wave/discussions/1624 but what I did was 
- set the unicode \u2705 and \u274c in https://github.com/h2oai/wave/blob/master/py/examples/table_download.py#L33
- In the project root directory, I run 'make all' and 'make run'
- Then I run 'cd ./py && ./venv/bin/wave run examples.tour'
- Load http://localhost:10101/tour#table_download in my browser
- Click download and verify in the csv file that the unicode is correctly parsed i.e., \u2705 and so on.

What I am not aware of to add the symbol itself as a string to a table. How do I do that in the UI, what do I run, what do I change and where ?

If you can let me know this, that might help already to have a direction to reproduce the exact scenario.

mturoci:
> set the unicode \u2705 and \u274c

This only sets unicode as plain strings. You need to set ✅ for example instead.

> What I am not aware of to add the symbol itself as a string to a table. How do I do that in the UI, what do I run, what do I change and where ?

Just set this in table_download.py code temporarily. Your testing flow is correct (verification via Tour), but you didn't test the right input. To cut time, here is a possible example you were supposed to test against:

```py
# Table / Download
# Allow downloading a table's data as CSV file.
# #table #download
# ---
import random
from faker import Faker

fake = Faker()

_id = 0


class Issue:
    def __init__(self, text: str, status: str, progress: float, icon: str, notifications: str):
        global _id
        _id += 1
        self.id = f'I{_id}'
        self.text = text
        self.status = status
        self.views = 0
        self.progress = progress
        self.icon = icon
        self.notifications = notifications


# Create some issues
issues = [
    Issue(
        text=fake.sentence(),
        status=('✅' if i % 2 == 0 else '❌'),
        progress=random.random(),
        icon=('BoxCheckmarkSolid' if random.random() > 0.5 else 'BoxMultiplySolid'),
        notifications=('Off' if random.random() > 0.5 else 'On')) for i in range(100)
]

# Create columns for our issue table.
columns = [
    ui.table_column(name='text', label='Issue'),
    ui.table_column(name='status', label='Status'),
    ui.table_column(name='notifications', label='Notifications'),
    ui.table_column(name='done', label='Done', cell_type=ui.icon_table_cell_type()),
    ui.table_column(name='views', label='Views'),
    ui.table_column(name='progress', label='Progress', cell_type=ui.progress_table_cell_type()),
]


@app('/demo')
async def serve(q: Q):
    q.page['form'] = ui.form_card(box='1 1 -1 11', items=[
        ui.table(
            name='issues',
            columns=columns,
            rows=[ui.table_row(
                name=issue.id, cells=[issue.text, issue.status, issue.notifications, issue.icon, str(issue.views),
                                      str(issue.progress)]) for issue in issues],
            downloadable=True,
        )
    ])
    await q.page.save()
```

ui/src/table.tsx
```diff
@@ -584,7 +584,7 @@ const
           dataRows = (m.groups ? m.groups.flatMap(({ rows }) => rows) : m.rows)?.map(({ cells }) => cells) || [],
           data = toCSV([m.columns.map(({ label, name }) => label || name), ...dataRows]),
           a = document.createElement('a'),
-          blob = new Blob([data], { type: "octet/stream" }),
+          blob = new Blob(['\uFEFF', data], { type: "text/csv;charset=utf-8," }),
```
mturoci, line 587: Can you please also add a comment explaining why the prefix is necessary + link to the SO question for future reference? Thanks!

mturoci (changes requested):
Thanks @JacksonIsaac! The solution is now correct, let's add a small comment for future and we are good to merge.

mturoci (approved):

JacksonIsaac:
@mturoci can you please add hacktoberfest-accepted, then it will be my first contribution of this year :)

mturoci:
No worries @JacksonIsaac, this repo has a `hacktoberfest` label which means every merged PR counts as valid Hacktoberfest contribution. No need for additional label.

Congrats on your first contribution!

PR #1679, opened 2022-10-27
feat: Add dialog prompt for download filename

JacksonIsaac:
## Example output

image


mturoci:
Thanks @JacksonIsaac. Unfortunately, I have to close this since this feature needs to be discussed and agreed upon first. Then we will create a GH issue with a detailed description to make it ready to be worked on. If you would like to contribute, please pick one of our existing GH issues.

JacksonIsaac:
Ah, that's indeed unfortunate. Who needs to discuss, and how long does it take to become a GH Issue ? Still trying to make it within the next few days for hacktoberfest :D

But definitely will keep contributing small fixes around later.

PR #1680, opened 2022-10-28
fix: Remove vertical scroll overflow in todo example

### LINKED ISSUE #776: Improvements to tour.py
Via @mturoci  #753

- [x] Split nav menu into subgroups (form / plots etc.) to make it easier for user when searching something.
- [x] Implement global search to quickly go to the example that user wants. This could be achieved by using either picker (not so great) or by #303 (better).
- [ ] Clean up all the examples and make them use only the minimum necessary card space (edit `box` attribute) so that it doesn't cause overflow within tour iframe even though content is short. In other words, prevent scrollbars where possible.
- [x] Distant future - integrate our online IDE and make the tour examples editable with live changes.

JacksonIsaac:
# Github issue

Fixes #776 

## Before output

before-tour-wave-todo

before-tour-wave-todo-add

## After output

after-tour-wave-todo

after-tour-wave-todo-add

### DIFF
--- a/py/examples/background.py
+++ b/py/examples/background.py
@@ -15,7 +15,7 @@ def blocking_function(secs) -> str:
 @app('/demo')
 async def serve(q: Q):
     if q.args.start:
-        q.page['form'] = ui.form_card(box='1 1 12 10', items=[ui.progress('Running...')])
+        q.page['form'] = ui.form_card(box='1 1 6 2', items=[ui.progress('Running...')])
         await q.page.save()
 
         seconds = random.randint(1, 6)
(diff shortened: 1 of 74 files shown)


JacksonIsaac:
@mturoci The latest commit should fix all the examples, both horizontal and vertical scroll wherever I could notice improvement opportunities.

mturoci (changes requested):
Thanks @JacksonIsaac, a few more adjustments are necessary. I recommend completing this on your laptop screen rather than an external monitor so that you can see the scrollbars directly + it's how most of our users will do.

The following is a list of examples that need adjustments:

* wide plot
* wide article preview
* form
* textbox
* table
* table tags
* image
* image annotator
* frame path
* plot schema
* plot form
* plot pandas
* vegalite form
* vegalite flex
* bokeh callbacks

mturoci (approved):
Thanks @JacksonIsaac!

PR #1681, opened 2022-10-29
fix: Reduce space between searchbar and table

### LINKED ISSUE #1492: Table with Search Bar has wastage of space
A table with a search bar leads to a large wasted space between the bar and the table.

A much smaller gap to utilize maximum space.

Screenshot 2022-06-09 at 6 26 12 PM

JacksonIsaac:
# Github Issue
Fixes #1492 

## Before 
image

## After
image

#### With pixel spacing dimension

image

image

#### Side by side (groupby and search)

image
image

### DIFF
--- a/ui/src/table.tsx
+++ b/ui/src/table.tsx
@@ -860,7 +860,7 @@ export const
         if (items.length > 10) return 500
 
         const
-          topToolbarHeight = searchableKeys.length || groupable ? 80 : 0,
+          topToolbarHeight = searchableKeys.length || groupable ? (groupable ? 80 : 46) : 0,
           headerHeight = 50,
           rowHeight = m.columns.some(c => c.cell_type?.progress) ? 76 : 48,
           footerHeight = m.downloadable || m.resettable || searchableKeys.length || m.columns.some(c => c.filterable) ? 46 : 0,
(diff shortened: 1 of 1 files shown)


mturoci:
How big is the gap? It would be great if it was exactly `15px`. Also can you also include screenshots of:

* Just searchbar - already included.
* Just groupby
* Searchbar + groupby together.

We need to make sure all these are consistent and have the same vertical spacing.

JacksonIsaac:
I have added screenshots for the pixel dimension between search/groupby bar and the table. The gap for searchbar is ~16px and for groupby is ~15px. I tried few different parameters, but this was the best I could get for them.

mturoci:
> The gap for searchbar is ~16px and for groupby is ~15px

Unfortunately, that's not good enough since they can be also displayed side by side in which case, it will not look good.

JacksonIsaac:
@mturoci how can I try that scenario where both are used side by side ? Do you perhaps have an example maybe ?

mturoci:
`table.py` should have them both IIRC.

JacksonIsaac:
When put together, in table.py example both search and groupby had exact space of 15px

image

mturoci (approved):
Thanks @JacksonIsaac!

PR #1694, opened 2022-11-13
feat: allow adjustable height of copyable text

cvanvoorhis:
This change adds a height attribute to ui.copyable text so that copyable textbox height can stay consistent with textbox height, as specified in issue #1302.

See example use case:
- text box of certain height
- user enters text and hits Submit
- if the query was successfully, user gets back a "copyable text box" of same height (i.e. ui.copyable_text() of same height)

Also includes updated docs and example.

### DIFF
--- a/ui/src/copyable_text.tsx
+++ b/ui/src/copyable_text.tsx
@@ -51,11 +51,13 @@ export interface CopyableText {
   name?: S
   /** True if the component should allow multi-line text entry. */
   multiline?: B
+  /** The adjustable height of the textbox. */
+  height?: S
 }
 
 export const XCopyableText = ({ model }: { model: CopyableText }) => {
   const
-    { name, multiline, label, value } = model,
+    { name, multiline, label, value, height } = model,
     ref = React.useRef<Fluent.ITextField>(null),
     timeoutRef = React.useRef<U>(),
     [copied, setCopied] = React.useState(false),
@@ -80,7 +82,14 @@ export const XCopyableText = ({ model }: { model: CopyableText }) => {
 
   return (
     <div data-test={name} className={multiline ? css.multiContainer : css.compactContainer}>
-      <Fluent.TextField componentRef={ref} value={value} label={label} multiline={multiline} styles={{ root: { width: pc(100) } }} readOnly />
+      <Fluent.TextField 
+        componentRef={ref}
+        value={value}
+        label={label}
+        multiline={multiline}
+        styles={ multiline ? { root: { width: pc(100), height: height} } : { root: { width: pc(100) } } }
+        readOnly
+      />
       <Fluent.PrimaryButton
         title='Copy to clipboard'
         onClick={onClick}
--- a/website/widgets/form/copyable_text.md
+++ b/website/widgets/form/copyable_text.md
@@ -32,3 +32,19 @@ q.page['form'] = ui.form_card(
     items=[ui.copyable_text(label='Copyable text', value=multiline_content, multiline=True)]
 )
 ```
+
+## Height of copyable text
+
+If you want to adjust the height of the copyable textbox, set the `height` attribute to your desired height.
+Note: The set height will only take effect if `multiline` is set to `True`.
+
+```py
+multiline_content = '''Wave is truly awesome.
+You should try all the features!
+Like setting the height to 200px!'''
+
+q.page['form'] = ui.form_card(
+    box='1 1 2 2',
+    items=[ui.copyable_text(label='Copyable text', value=multiline_content, multiline=True, height='200px')]
+)
+```
--- a/py/examples/copyable_text.py
+++ b/py/examples/copyable_text.py
@@ -13,6 +13,7 @@ page = site['/demo']
 page['hello'] = ui.form_card(box='1 1 3 3', items=[
     ui.copyable_text(label='Copyable text', value='Hello world!'),
     ui.copyable_text(label='Copyable text', value=multiline_content, multiline=True),
+    ui.copyable_text(label='Copyable text', value=multiline_content, multiline=True, height='200px'),
 ])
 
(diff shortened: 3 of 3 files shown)


mturoci:
Thanks for the PR @cvanvoorhis!

> See example use case

Can you please update the PR description with a video/screenshots that verify your changes work?

cvanvoorhis:
These changes alone crash the program, so I made the changes to ui.py and types.py that allows height to be an attribute of CopyableText. However, the height of the textbox does not actually change in the display, as you can see in the attached screenshots.
Screenshot_20221116_023512
Screenshot_20221116_023417
Screenshot_20221116_023425
Despite setting the height to 1000px and then 500px, the height of each textbox is the same, and they display a scrollbar. I believe this because of the TextField.styles.js in the @fluentui module, which has the relevant lines in an attached screenshot. If multiline is true, the height of the textbox is set to auto.
Screenshot_20221116_022934
So my question is, do we want to change this implementation? My changes won't show up visually because of this implementation, but changing it would change all TextFields in the wave app I believe, and we do not want to do that. 

What are your thoughts?

mturoci:
Thanks @cvanvoorhis!

> These changes alone crash the program, so I made the changes to ui.py and types.py that allows height to be an attribute of CopyableText.

When adding a new attribute to a widget, you need to call `make generate` to generate the type definitions automatically after editing Typescript definitions. See [contributing guidelines](https://wave.h2o.ai/docs/contributing/#development-setup).

> I believe this because of the TextField.styles.js in the @FluentUI module

That's possible. In that case, you shall overwrite the default Fluent styles if your new attribute (`height`) is set. If it's not set, the component should render as before to prevent breaking changes for existing apps. Here is a [small example](https://github.com/h2oai/wave/blob/master/ui/src/textbox.tsx#L111-L115) of how something similar is done for our textbox component.

cvanvoorhis:
Apologies for the lack of communication lately.

Thank you for the help! When I run make generate the page now updates, however the height is not showing up as intended. It displays as spacing in between the textboxes themselves, as shown below.
Screenshot_20221206_053625
Screenshot_20221206_053638

When I changed copyable_text.tsx to have `field: { height: height}, fieldGroup: { minHeight: height }` like the textbox.tsx example, the textboxes display the same as the images before. 

Personally I think the expandable textboxes of copyable text already allow for adjustable height, what do you all think about changing this feature to be called **spacing** instead of **height**?

mturoci:
> however the height is not showing up as intended

Yes, you need to make the styling right :) You need to first identify the HTML node that needs the height to be set (I assume it's going to be some textarea) and then find out how to style it in the FluentUI way.

> Personally I think the expandable textboxes of copyable text already allow for adjustable height, what do you all think about changing this feature to be called spacing instead of height?

Spacing would be of no use. The goal of this issue is to provide a way to render bigger area of copyabletext in case there is simply too text for copying and you would like to avoid scrollbars.

As noted previously, you should focus more on the correct CSS styling.

cvanvoorhis:
Just pushed commits that implement this change. The height is still adjustable after the textbox is made. You can see the height is set in the images below:

Screenshot_20221207_112134
Screenshot_20221207_112143
Screenshot_20221207_112154

Let me know if I should add more documentation and/or make the example better somehow?

ui/src/copyable_text.tsx
```diff
@@ -51,11 +51,13 @@ export interface CopyableText {
   name?: S
   /** True if the component should allow multi-line text entry. */
   multiline?: B
+  /** The adjustable height of the textbox. */
```
mturoci, line 54:
```suggestion
  /** Custom height in px, e.g. '200px'. Requires `multiline` to be set. */
```

Don't forget to run `make generate` after this change.

ui/src/copyable_text.tsx
```diff
@@ -80,7 +82,16 @@ export const XCopyableText = ({ model }: { model: CopyableText }) => {
 
   return (
     <div data-test={name} className={multiline ? css.multiContainer : css.compactContainer}>
-      <Fluent.TextField componentRef={ref} value={value} label={label} multiline={multiline} styles={{ root: { width: pc(100) } }} readOnly />
+      <Fluent.TextField 
+        componentRef={ref}
+        value={value}
+        label={label}
+        multiline={multiline}
+        styles={ multiline && height && !height.endsWith('%')
```
mturoci, line 90: Restricting the `height` to only be defined in `px` in docstring should be enough.
```suggestion
        styles={ multiline && height
```

website/widgets/form/copyable_text.md
```diff
@@ -32,3 +32,19 @@ q.page['form'] = ui.form_card(
     items=[ui.copyable_text(label='Copyable text', value=multiline_content, multiline=True)]
 )
 ```
+
+## Height of copyable text
+
+If you want to adjust the height of the copyable textbox, set the `height` attribute to your desired height.
```
mturoci, line 38:
```suggestion
If you want to adjust the height of the copyable textbox, set the `height` attribute to your desired height in pixels, e.g. `'200px'`.
```

mturoci (changes requested):
Thanks @cvanvoorhis! The changes look much better. One thing that needs fixing though is the styling during/after resizing.


https://user-images.githubusercontent.com/64769322/206400037-c3181c25-65fe-402d-9741-390d2d6f4c5c.mov

cvanvoorhis:
Thank you for your suggestions! I made those changes as well as kept the styling of expanding the textbox that was available before in the most recent commit.

Here is a vid showing that off:
https://user-images.githubusercontent.com/113397293/206731912-ddbb61f3-3dfe-4470-b3a4-d24d140a5f00.mp4

mturoci (approved):
Perfect. Thanks @cvanvoorhis!

PR #1709, opened 2022-11-19
feat: Adjust position of table searchbar to always be on the left, groupby always on the right #1671

### LINKED ISSUE #1671: Adjust Wave table search bar position

By default the search bar in a wave table is located in left hand corner. When the group by feature enabled it goes to right hand corner as shown in below images. 
image
image

I need a control to locate the search bar from our own. (left hand side or right hand side). I am looking to maintain consistency over the pages which are shown above and I like to keep search bar in right hand corner even I am not using group by feature. 

Enable an API to adjust the positions of those search bar and group by dropdown.  

--

#### Additional context

If there is any alternative solution available happy to know it. Thank you.

buhaytza2005:
Adjust position of searchbar to always be on the left, groupby always on the right

Closes #1671


ui/src/table.tsx
```diff
@@ -943,16 +943,22 @@ export const
...
+          <Fluent.Stack.Item align='end' styles={{ root: { width:300,}}}  >
+          {groupable && !m.pagination && <Fluent.Dropdown data-test='groupby' label='Group by' selectedKey={groupByKey} onChange={onGroupByChange} options={groupByOptions}/>}
+           </Fluent.Stack.Item>
+       </Fluent.Stack >
         <Fluent.ScrollablePane
           componentRef={contentRef}
           scrollbarVisibility={Fluent.ScrollbarVisibility.auto}
           styles={{
-            root: { top: groupable || searchableKeys.length ? (groupable ? 74 : 48) : 0, bottom: shouldShowFooter ? 46 : 0 },
+            /* root: { top: groupable || searchableKeys.length ? (groupable ? 74 : 48) : 0, bottom: shouldShowFooter ? 46 : 0 },*/
```
mturoci, line 960: chore: This commented out code can be removed.

ui/src/table.tsx
```diff
@@ -943,16 +943,22 @@ export const
...
+            root: { top: groupable || searchableKeys.length ? (groupable ? 120 : 100) : 0, bottom: shouldShowFooter ? 46 : 0 },
```
mturoci, line 961: This seems to add a lot of unnecessary space between search/groupby and table. Let's go with the previous values (`groupable ? 74 : 48`).

ui/src/table.tsx
```diff
@@ -943,16 +943,22 @@ export const
     }), [filteredItems, groups, expandedRefs, isMultiple, items, m, onFilterChange, selectedFilters, selection, sort, setFiltersInBulk])
 
     return (
-      <div data-test={m.name} style={{ position: 'relative', height: computeHeight() }}>
-        <Fluent.Stack horizontal horizontalAlign='space-between' verticalAlign='end'>
-          {groupable && !m.pagination && <Fluent.Dropdown data-test='groupby' label='Group by' selectedKey={groupByKey} onChange={onGroupByChange} options={groupByOptions} styles={{ root: { width: 300 } }} />}
-          {!!searchableKeys.length && <Fluent.SearchBox data-test='search' placeholder='Search' onChange={onSearchChange} value={searchStr} styles={{ root: { width: '50%', maxWidth: 500 } }} />}
-        </Fluent.Stack >
+      <div data-test={m.name} style={{ position: 'relative', height: computeHeight()  }}>
+        <Fluent.Stack horizontal horizontalAlign='space-between' verticalAlign="end"  >
```
mturoci, line 947: Is the `verticalAlign` necessary here?

ui/src/table.tsx
```diff
@@ -943,16 +943,22 @@ export const
...
+           <Fluent.Stack.Item align='start' styles={{ root: { width: '50%', maxWidth: 500, margin: "28px 0 0 0"} }}>
```
mturoci, line 948: I guess the `align` should be `end` here instead of `start` since we use horizontal Stack and according to the Fluent align docs:

_Defines how to align the StackItem along the x-axis (for vertical Stacks) or the y-axis (for horizontal Stacks)._

Also, what is the extra margin for?

```suggestion
           <Fluent.Stack.Item align='end' styles={{ root: { width: '50%', maxWidth: 500} }}>
```

py/examples/table_search_groupable.py
```diff
@@ -0,0 +1,102 @@
+import random
```
mturoci, line 1: This example made the review buttery smooth, but we have plenty of specialized table examples so will not merge this one. Please remove it.

mturoci (changes requested):
Thanks for the PR @buhaytza2005! It looks good overall, just needs a few final changes.

ui/src/table.tsx
buhaytza2005: I had some issues aligning the two buttons and ended up adding some padding. This in turn pushed the search bar into the table. I tested increasing the whitespace at the top of the table. I think the main issue was either the vertical align or the start align.
image

Hope this is ok now.

py/examples/table_search_groupable.py
buhaytza2005: done

ui/src/table.tsx
buhaytza2005: Not sure, I think it caused some of the issues I had.

mturoci (approved):
Thanks @buhaytza2005! The PR looks good now. I removed the docs changes as these are not needed since this change is visual and doesn't need much more explanation.

PR #1710, opened 2022-11-21
fix: Make disabled ui.button icon and text colours consistent #1664

### LINKED ISSUE #1664: Button icon contrast is inconsistent
#### Wave SDK Version, OS
v0.23.1 on MacOS

#### Actual behaviour
Default theme:
Screenshot 2022-10-08 at 4 10 34 PM

Dark theme:
Screenshot 2022-10-08 at 4 10 50 PM

1st button: Disabled, Not Primary
2nd button: Not Disabled, Not Primary
3rd button: Disabled, Primary
4th button: Not Disabled, Primary

#### Expected behaviour
(Unless there's some trickery that my eye is missing) It seems the 1st, 2nd & 4th buttons have the icon and text using the same colour but the 3rd button has the icon and text in different colour contrasts, irrespective of the theme.

#### Comments
Maybe related to https://github.com/h2oai/wave/issues/791

marek-mihok:
This PR provides an override of icon colour for all disabled FluentUI buttons.

https://user-images.githubusercontent.com/23740173/203029158-a6c771f6-5731-4253-9a94-8a953f300aad.mov

Closes #1664

### DIFF
--- a/ui/src/app.tsx
+++ b/ui/src/app.tsx
@@ -63,7 +63,19 @@ const
       flexDirection: 'column',
       alignItems: 'center',
     },
-  })
+  }),
+  buttonStyles = { styles: { iconDisabled: { color: 'unset' } } },
+  // The global overrides for component styles.
+  componentCustomStyles = {
+    DefaultButton: buttonStyles,
+    PrimaryButton: buttonStyles,
+    IconButton: buttonStyles,
+    ActionButton: buttonStyles,
+    CommandButton: buttonStyles,
+    CommandBarButton: buttonStyles,
+    CompoundButton: buttonStyles,
+    SpinButton: buttonStyles,
+  }
 
 const
   BusyOverlay = bond(() => {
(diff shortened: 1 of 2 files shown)


mturoci (changes requested):
Thanks @marek-mihok, we should ideally fix this globally for every Fluent button component (e.g. we use buttons for pagination in table etc.).

ui/src/app.tsx
```diff
@@ -116,7 +128,7 @@ const
                 const page = e.page
                 const lightbox = lightboxB()
                 return (
-                  <Fluent.ThemeProvider>
+                  <Fluent.ThemeProvider theme={{ components: componentCustomStyles }}>
```
mturoci, line 131: Use a file-scoped const instead to improve perf as suggested in Fluent docs:

image

mturoci (changes requested):

ui/src/app.tsx
marek-mihok: I've first moved the constant `componentCustomStyles`  from inside `bond()` to the upper scope, but now I see I should move also whole `{components: componentCustomStyles}` object. 
Thank you @mturoci , now it should be correct.

mturoci (approved):
Thanks @marek-mihok!

PR #1712, opened 2022-11-21
feat: Wave university Pypi publish

mturoci:
## Demo

https://user-images.githubusercontent.com/64769322/203079349-7ac1e988-aea1-4fb4-9585-6dcd53861f50.mov

There will be 2 commands required to install Wave University:

```sh
pip install h2o_wave_university
wave-university # or maybe h2o-wave-university would be better to be consistent with the package name even for the price of verbosity?
```

@lo5 once you give this thumbs up, I will publish to the Pypi (currently published to test Pypi only). The main point of review is naming, but feel free to point out any other possible improvements.

Tested against MacOS and Windows (thanks @Far0n!)


lo5:
`wave-university` is fine.

The other option is to add the `learn` subcommand to `wave` in the `h2o_wave` package, and try/except `import h2o_wave_university`. If the import fails, exit with a message to `pip install h2o-wave-university` and then run `wave learn`.

PR #1715, opened 2022-11-22
feat: Image popup refactor

marek-mihok:
This PR brings the refactor of image popup/lightbox component and also adds a pointer cursor on clickable areas.

### DIFF
--- a/ui/src/parts/lightbox.tsx
+++ b/ui/src/parts/lightbox.tsx
@@ -26,8 +26,7 @@ const
   NAV_IMAGE_SIZE = 120,
   IMAGE_NAV_HEIGHT = NAV_IMAGE_SIZE + 28, // 28 is the height of the scrollbar.
   ICON_SIZE = 38,
-  ICON_MARGIN = 8,
-  IMAGE_NAV_BOTTOM_MARGIN = 8
+  LIGHTBOX_PAGE_MARGIN = 8
 
 const
   css = stylesheet({
@@ -42,18 +41,18 @@ const
       margin: 'auto',
       left: 0,
       right: 0,
-      top: ICON_SIZE + ICON_MARGIN * 2,
+      top: ICON_SIZE + LIGHTBOX_PAGE_MARGIN * 2,
       maxWidth: '100vw'
     },
     closeButton: {
       position: 'absolute',
-      right: ICON_MARGIN,
-      top: ICON_MARGIN
+      right: LIGHTBOX_PAGE_MARGIN,
+      top: LIGHTBOX_PAGE_MARGIN
     },
     imageNav: {
       position: 'absolute',
       left: '50%',
-      bottom: IMAGE_NAV_BOTTOM_MARGIN,
+      bottom: LIGHTBOX_PAGE_MARGIN,
       height: IMAGE_NAV_HEIGHT,
       whiteSpace: 'nowrap',
       overflowX: 'auto',
@@ -73,7 +72,7 @@ const
     text: {
       position: 'absolute',
       left: '50%',
-      maxWidth: 'calc(100vw - 40px)',
+      maxWidth: `calc(100vw - ${2 * LIGHTBOX_PAGE_MARGIN}px)`,
       boxSizing: 'border-box',
       color: '#fff',
       whiteSpace: 'nowrap',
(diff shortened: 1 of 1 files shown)


ui/src/parts/lightbox.tsx
```diff
@@ -22,93 +22,83 @@ import { getColorFromString, isDark } from '@fluentui/react'
...
-  FOOTER_VERTICAL_PADDING = 10,
-  IMAGE_NAV_HEIGHT = 148,
-  IMAGE_SIZE = 120,
-  ICON_SIZE = 38
+  IMAGE_CAPTIONS_HEIGHT = 46, // Total height of title and decription.
+  NAV_IMAGE_SIZE = 120,
+  IMAGE_NAV_HEIGHT = NAV_IMAGE_SIZE + 28, // 28 is the height of the scrollbar.
+  ICON_SIZE = 38,
+  ICON_MARGIN = 8,
+  IMAGE_NAV_BOTTOM_MARGIN = 8
```
mturoci, line 30: Maybe we could combine this into a single `LIGHTBOX_PAGE_MARGIN`?

ui/src/parts/lightbox.tsx
```diff
@@ -22,93 +22,83 @@ import { getColorFromString, isDark } from '@fluentui/react'
...
       margin: '2px',
       cursor: 'pointer',
       filter: 'brightness(30%)',
       border: '2px solid #000'
     },
     text: {
-      whiteSpace: 'nowrap',
+      position: 'absolute',
+      left: '50%',
+      maxWidth: 'calc(100vw - 40px)',
```
mturoci, line 76: Magic number `40`.

ui/src/parts/lightbox.tsx
```diff
@@ -22,93 +22,83 @@ import { getColorFromString, isDark } from '@fluentui/react'
...
+  ICON_SIZE = 38,
+  ICON_MARGIN = 8,
+  IMAGE_NAV_BOTTOM_MARGIN = 8
 
 const
   css = stylesheet({
     body: {
       position: 'fixed',
       inset: '0px',
       zIndex: 1,
```
mturoci, line 37: Is this `z-index` needed?

ui/src/parts/lightbox.tsx
```diff
@@ -22,93 +22,83 @@ import { getColorFromString, isDark } from '@fluentui/react'
...
+      height: NAV_IMAGE_SIZE,
+      width: NAV_IMAGE_SIZE,
       boxSizing: 'border-box',
       objectFit: 'cover',
-      height: IMAGE_SIZE,
-      width: IMAGE_SIZE,
       margin: '2px',
       cursor: 'pointer',
       filter: 'brightness(30%)',
       border: '2px solid #000'
```
mturoci, line 71: These 2 are overriden in the inline `imageHighlightStyle `. Do we need them in both places? Since they are static, I guess they should be removed from inline style and kept in this class declaration.

ui/src/parts/lightbox.tsx
```diff
@@ -149,7 +139,15 @@ export const Lightbox = ({ images, defaultImageIdx }: LightboxProps) => {
...
-      if ((ev.target as HTMLElement).parentElement === ev.currentTarget) lightboxB(null)
+      if (ev.target === ev.currentTarget) lightboxB(null)
+    },
+    handleCursorOnFreeSpaceHover = (ev: React.MouseEvent<HTMLElement>) => {
+      // Changes the cursor to 'pointer' when hovering over the free space.
+      if (ev.target === ev.currentTarget) ev.currentTarget.style.cursor = 'pointer'
+      else {
+        ev.currentTarget.style.cursor = ''
+        ev.currentTarget.style.cursor = getComputedStyle(ev.currentTarget).cursor
+      }
```
mturoci, line 150: This seems unnecessary. Couldn't it be achieved with a simple CSS rule?

mturoci (changes requested):

ui/src/parts/lightbox.tsx
marek-mihok: It is necessary when you have the image on the same page together with other components with `zIndex: 1` specified, e.g. the `ui.header`:
image

It is maybe good to consider if the `zIndex` with value of `1` is enough and shouldn't be increased just in case.

ui/src/parts/lightbox.tsx
marek-mihok: They are overriden just in case when the image is hovered or currently selected:
```ts
className={clas(css.navImg, 'lazy', style({ $nest: { '&:hover': imageHighlightStyle } }))}
style={activeImageIdx === idx ? imageHighlightStyle : undefined}
```
where 
```ts
imageHighlightStyle = { filter: 'unset', border: '2px solid', borderColor: activeColor },
```
That means brightness is set back to `100%` by providing `unset` value and border changes color to `activeColor` which is computed dynamically based on the current theme.

ui/src/parts/lightbox.tsx
mturoci: I see. I don't see a reason why `ui.header` should have special `z-index`, but it may be a valid option in some other cases.

ui/src/parts/lightbox.tsx
mturoci: Makes sense.

ui/src/parts/lightbox.tsx
mturoci: > Another option would be to specify the pointer cursor for body element and also cursors for every child element separately to override the parent's one.

Yes, that's what I meant. Such things should be done in JS only as last-resort solutions.

ui/src/parts/lightbox.tsx
mturoci: Yes, keep it.

mturoci (approved):

PR #1723, opened 2022-11-28
docs: Show how to use local CSS file as stylesheet.

marek-mihok:
Adds the example of how to apply the css stylesheet stored on the Wave server.

image

### DIFF
--- a/website/docs/custom-css.md
+++ b/website/docs/custom-css.md
@@ -24,15 +24,24 @@ q.page['meta'] = ui.meta_card(
 )
 ```
 
-It is also possible to host a stylesheet on the Wave server itself.
+### Using local CSS stylesheet
+
+Prior to using your own CSS files, they need to be uploaded to Wave server. There are 2 ways of doing this:
+
+* Uploading the CSS files from your Wave app via [q.site.upload](https://wave.h2o.ai/docs/files/#provide-file-downloads).
+* Serving the CSS files directly from Wave server via [public/private dir](https://wave.h2o.ai/docs/files/#serving-files-directly-from-the-wave-server).
 
 ```py
-example_dir = os.path.dirname(os.path.realpath(__file__))
+from h2o_wave import Q, main, app, ui
+import os
 
-@app('/demo')
+current_dir = os.path.dirname(os.path.realpath(__file__))
+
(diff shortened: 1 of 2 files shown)


marek-mihok:
@mturoci now I am reading through your Discord answers regarding this. Maybe it would be fine to also add here a point about css stylesheet served directly from the Wave server. Wdyt?

website/docs/custom-css.md
```diff
@@ -24,6 +24,26 @@ q.page['meta'] = ui.meta_card(
 )
 ```
 
+It is also possible to host a stylesheet on the Wave server itself.
```
mturoci, line 27: Let's add h3 here, something like `### Using local CSS file`.

mturoci, line 27:
```suggestion
Prior to using your own CSS files, they need to be uploaded to Wave server. There are 2 ways of doing this:

* Uploading the CSS files from your Wave app via [q.site.upload](https://wave.h2o.ai/docs/files/#provide-file-downloads).
* Serving the CSS files directly from Wave server via [public/private dir](https://wave.h2o.ai/docs/files/#serving-files-directly-from-the-wave-server).
```

website/docs/custom-css.md
```diff
@@ -24,6 +24,26 @@ q.page['meta'] = ui.meta_card(
...
+
+```py
+example_dir = os.path.dirname(os.path.realpath(__file__))
```
mturoci, line 30:
```suggestion
from h2o_wave import Q, main, app, ui
import os

current_dir = os.path.dirname(os.path.realpath(__file__))
```

website/docs/custom-css.md
```diff
@@ -24,6 +24,26 @@ q.page['meta'] = ui.meta_card(
...
+
+@app('/demo')
```
mturoci, line 32:
```suggestion
@app('/')
```

mturoci (changes requested):
Thanks @marek-mihok. In addition to these comments, please also add a small comment to `meta_stylesheet.py`'s path. Something like - this path can also be the one returned from q.site.upload if you want to use your own CSS files.

mturoci (approved):

PR #1724, opened 2022-11-28
Fix: Table grouped incorrectly after rows change

### LINKED ISSUE #1687: Tables Become Incorrectly Grouped in Meta/Dialog (Wave v0.16 --> v0.22+)
#### Wave SDK Version, OS
H2O Wave v0.22+ on macOS

#### Actual behavior
This is for Tables (ui.table) inside a Meta/Dialog. When upgrading Wave v0.16 --> v0.22+, Tables behave differently inside a Meta/Dialog when using "Group by."
(When the table is moved outside of the dialog and back to the main page/form, however, the table works as expected.)

For H2O Wave v0.22+, the issue happens when using "Group by" in a dialog table, selecting a group, and doing an action on it.
For example, in screenshot 1 below, I've Grouped by the column "Group" and selected the top group "Tap". 

When I do an action (with a bottom white button), the items get jumbled so the grouping becomes incorrect. See how items under "Tap" have changed and were incorrectly placed there in screenshot 2 below:

***It only happens when using "Group by," so when selecting normally without the grouping, there's no such issue.***

#### Expected behavior
In Wave v0.16.0, the table used to be reset automatically and ungrouped, so there'd be no inconsistencies like this as shown below:

Wave v0.16.0 - Using "Group by" in a table, selecting a group, and doing an action on it:

Wave v0.16.0 - The result after the action is that the table would be reset automatically and ungrouped so there's no incorrect grouping: 

#### Steps To Reproduce
Please see "Actual behavior" above.

marek-mihok:
This PR makes sure table is reset after table rows change.

https://user-images.githubusercontent.com/23740173/205644233-32af549a-c4b6-4abf-80a2-83742d644493.mov

Closes #1687


mturoci (changes requested):
Thanks @marek-mihok. Please make sure all the unit tests pass.

ui/src/table.tsx
```diff
@@ -930,7 +931,15 @@ export const
       }
       // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [])
-    React.useEffect(() => setFilteredItems(items), [items])
+
+    React.useEffect(() => {
```
mturoci, line 935: This seems like a common thing we use and will use in the future due to React being what it is. Wdyt about extracting this into a separate custom hook? Something like:

```tsx
import React from 'react'

const useUpdateOnlyEffect = (func: (...args: any) => any, deps: any[]) => {
  const didMount = React.useRef(false)
  React.useEffect(() => { didMount.current ? func() : didMount.current = true }, deps)
}

export default useUpdateOnlyEffect
```

marek-mihok: Yes, you are right. We are using it quite often so why not. :)

mturoci (approved):
Perfect, thanks!

PR #1735, opened 2022-12-05
feat: Allow users to remove paddings on frame cards

### LINKED ISSUE #495: Allow frame cards to be displayed without margin/padding
Would be useful when embedding other Wave or third-party pages. 
Currently the card has its own padding, combined with the padding/margin of the embedded page.

kitty3190:
Added a parameter called `compact` in the `State` interface of `frame.tsx`. Users can specify whether they want the frame cards to have title and paddings or not. `compact` defaults to `False`.

Example:
Screenshot 2022-12-07 at 12 56 00
Frame card with no padding specified:
Screenshot 2022-12-07 at 12 56 13

The changes are reflected above. The first frame card with default padding (`compact` not specified), the second frame card with padding specified, and the third frame card with no padding specified. 

The code used for testing the changes above:
```py
# Frame card with default padding
page['example1'] = ui.frame_card(
    box='1 1 -1 4',
    title='Frame card with default padding',
    path='https://example.com',
)

# Frame card with padding
page['example2'] = ui.frame_card(
    box='1 5 -1 4',
    title='Frame card with padding specified',
    path='https://example.com',
    compact=False
)

# Frame card with no padding
page['example3'] = ui.frame_card(
    box='1 9 -1 4',
    title='Frame card with no padding specified',
    path='https://example.com',
    compact=True
)
```
This pull request closes #495.

### DIFF
--- a/ui/src/frame.tsx
+++ b/ui/src/frame.tsx
@@ -110,7 +110,7 @@ export const XFrame = ({ model: { name, path, content, width = '100%', height = 
 export const
   View = bond(({ name, state, changed }: Model<State>) => {
     const render = () => (
-      <div data-test={name} className={clas(css.card, (state.compact ?? false) ? '' : css.cardPadding)}>
+      <div data-test={name} className={clas(css.card, state.compact ? '' : css.cardPadding)}>
         {!state.compact && <div className='wave-s12 wave-w6'>{state.title}</div>}
         <div className={css.body}>
           <InlineFrame path={state.path} content={state.content} />
--- a/ui/src/frame.tsx
+++ b/ui/src/frame.tsx
@@ -26,6 +26,11 @@ const
       flexDirection: 'column',
       padding: grid.gap,
     },
+    card_no_padding: {
+      display: 'flex',
+      flexDirection: 'column',
+      padding: 0,
+    },
     body: {
       flexGrow: 1,
       position: 'relative',
@@ -83,6 +88,10 @@ interface State {
    * :value ""
    */
   content?: S
+  /**
+   * True if the component should have paddings. Defaults to True.
+   */
+  has_padding?: B
 }
 
 const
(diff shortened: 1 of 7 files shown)


ui/src/frame.tsx
```diff
@@ -83,6 +85,10 @@ interface State {
    * :value ""
    */
   content?: S
+  /**
+   * True if the component should have paddings. Defaults to True.
+   */
+  has_padding?: B
```
mturoci, line 91: Let's rename it to `compact` to be consistent with the existing API.

mturoci (changes requested):
Thanks @kitty3190!

Based on your screenshots, it seems like it might be a good idea to also not display the title (even if specified) if the padding is turned off. This way, the whole frame card should be occupied purely with the `content` (be it an HTML or 3rd party site link).

Make sure to reflect the above in docstrings as well.

kitty3190:
Thanks for the review @mturoci! I addressed the issues in my latest commit.

ui/src/frame.tsx
```diff
@@ -104,8 +110,8 @@ export const XFrame = ({ model: { name, path, content, width = '100%', height =
 export const
   View = bond(({ name, state, changed }: Model<State>) => {
     const render = () => (
-      <div data-test={name} className={css.card}>
-        <div className='wave-s12 wave-w6'>{state.title}</div>
+      <div data-test={name} className={clas(css.card, (state.compact ?? false) ? '' : css.cardPadding)}>
```
mturoci, line 113: nit: It's going to be falsy by default anyway so this should be enough.
```suggestion
      <div data-test={name} className={clas(css.card, state.compact ? '' : css.cardPadding)}>
```

mturoci (changes requested):
Looking good. I forgot the docs last time. Can you please add this to both `frame.md` files? Once done, it's good to be merged.

kitty3190:
Thanks @mturoci! I've added documentation in the `frame.md` file inside the `content` folder which relates to `ui.frame_card` in my latest commit. I believe the other `frame.md` file in the `form` folder relates to `ui.frame` which I did not modify. Let me know if additional changes are required!

lo5 (dismissed):

mturoci (approved):

PR #1737, opened 2022-12-06
feat: Add “align” parameter for ui.inline #1535

### LINKED ISSUE #1535: Allow vertical alignment for ui.inline

Nicolapps:
This PR adds a new `align` parameter to `ui.inline` that allows users to specify the vertical alignment of the items.

The API is updated as follows:

```diff
 interface Inline {
   /** The components laid out inline. */
   items: Component[]
   /** Specifies how to lay out the individual components. Defaults to 'start'. */
   justify?: 'start' | 'end' | 'center' | 'between' | 'around'
+  /** Specifies how the individual components are aligned on the vertical axis. Defaults to 'center'. */
+  align?: 'start' | 'end' | 'center' | 'baseline' | 'stretch'
   /** Whether to display the components inset from the parent form, with a contrasting background. */
   inset?: B
 }
```

The PR also adds a new example showing how values for the `justify` and `align` parameters of `ui.inline` work:
<video src="https://user-images.githubusercontent.com/7029582/205581533-c932cbdb-a2ae-4265-a4d2-8fcc076bbdcb.mov">

I also renamed the `alignment` parameter of the internal function `XComponents` to `justify` (more details here: 08d6813fe66bc1b316f00d93010e40095fb26331).

Closes #1535.


mturoci (changes requested):
Thanks @Nicolapps! The changes look good to me, just a few small nits.

Can you also document the various alignments in `inline.md`? Make a separate section for both horizontal and vertical one. Thanks!

Nicolapps:
Thank you for your review! I expanded the documentation of `inline.md`, please tell me if there is anything else you would like to see.

Inline page of the H2O Wave documentation

website/widgets/form/inline.md
```diff
@@ -22,3 +22,18 @@ q.page['form'] = ui.form_card(
 ```
 
 Check the full API at [ui.inline](/docs/api/ui#inline).
+
+## Customize the alignment
+
```
mturoci, line 27: Not needed.
```suggestion
```

website/widgets/form/inline.md
```diff
@@ -22,3 +22,18 @@ q.page['form'] = ui.form_card(
...
+### Horizontal alignment (`justify`)
```
mturoci, line 28:
```suggestion
## Horizontal alignment (`justify`)
```

website/widgets/form/inline.md
```diff
@@ -22,3 +22,18 @@ q.page['form'] = ui.form_card(
...
+
+You can specify how the elements should be horizontally aligned using the `justify` parameter. The default value is `start`.
+
+![Supported values for the justify parameter](/img/widgets/inline-justify.svg)
```
mturoci, line 32: Instead of adding the images yourself, you can write a minimal code (see L17 of this file) and during the build, the code gets executed and a corresponding screenshot for it is generated and appended. This way people can directly see what gets rendered after copypasting the code.

It's enough if you just include a few code examples (one per align/justify).

If you are curious though, you can try the build yourself:

* From project root run `make build-ui` to create a UI build including your most recent changes.
* `make pydocs` which will generate all the docs assets.
* `cd website && npm start` to see the changed docs.

Make sure to remove your custom `svg` files from the PR once you are done.

website/widgets/form/inline.md
```diff
@@ -22,3 +22,18 @@ q.page['form'] = ui.form_card(
...
+
+### Vertical alignment (`align`)
```
mturoci, line 34:
```suggestion
## Vertical alignment (`align`)
```

mturoci (changes requested):

Nicolapps:
Thanks again, I updated the examples:

Screenshot 2022-12-08 at 17 32 03
Screenshot 2022-12-08 at 17 32 07

I realized that the `stretch` alignment doesn’t work with buttons because only the wrapper div of the button gets expanded. Should I remove this option entirely from `ui.inline`?

mturoci:
The docs look perfect now! @Nicolapps 

> I realized that the stretch alignment doesn’t work with buttons because only the wrapper div of the button gets expanded. Should I remove this option entirely from ui.inline?

I think we can remove it since I cannot think of a usecase where you would want to have a button (or any other component, e.g. dropdown) to stretch to the whole size of the container - would look odd. Thus you can remove it for now (don't forget to run `make generate` to update the generated API as well).

Nicolapps:
Thanks, I removed the `stretch` option in my latest commit.

mturoci (approved):
Great PR. Good job @Nicolapps!

Issue #348, opened 2020-12-02
Handle Enter key press in textboxes


mtanco:
I have found on apps with a text box and button, as a user, I keep hitting `enter` when I'm done typing. I know text boxes have the option for a time trigger, but this might be more intuitive. Thoughts?

lo5:
> Isn't the main purpose of trigger prop getting rid of submit button and submit the value automatically? 

The usage patterns are different:
The purpose of the `trigger` property is "sync values whenever a component's value changes" (equivalent to textbox/dropdown `onChange`).
What this particular issue is asking for is: "sync values only when the enter key is pressed".

For example, if you enabled the former behavior on a login dialog, it would send username/password to the app as you were typing in either username or password.

> How is this prop different from any other textbox prop? 

Same explanation as above. For example, submitting a username/password combo is not an attribute of either the username textbox or the password textbox.

geomodular:
I think I can highlight some information here to help with conversation.

The rule:

>  if the enter key is pressed when the focus is on a child of a form_card and there is at least one button with primary=True, simulate a click on that button.

Is correct, because the button carries `name` information which will be passed to client along with other changes. Without the `name`, it would be just `qd.sync()`.

This `name` is then used in Python `if` to drive the program execution in desired way. Therefore, it makes sense to tie `enter` with the primary button for given `form_card`.

> What about the usecase where you have just a single textbox (no buttons). You should be able to submit it with enter as well I think.
> Not every input element needs accompanying submit button.

Consider having `textbox` with no buttons on form card but trigger, the `enter` action will be the same as `trigger` action on Python side. User would not be able to distinguish between them.

> Yes, that's my point. Imagine you type something within app, do not save / submit it just yet. In the meantime, some dialog / notification etc. pops up, you submit an action and accidentally submit all the data you previously filled.

As stated above, this is by design.

Consider two `form_cards` simultaneously on page. Both have input boxes and own `primary button`. User fills the first form but clicks on the primary button of second form. **All** of the changes will be passed to server. On the python side, however, `if` will handle the `primary button` for second form with no values for the second form.

I hope I didn't mess something up. I think, the fact that `button` has `name` attribute was overlooked.

mturoci:
Thanks for joining the discussion @geomodular!
> Is correct, because the button carries name information which will be passed to client along with other changes. Without the name, it would be just qd.sync().
[...]

PR #1744, opened 2022-12-09
feat: Add trigger_on_enter parameter to ui.inline #372

### LINKED ISSUE #372: Add enter handler for textbox

sherrychen2:
This PR adds a new `trigger_on_enter` parameter to `ui.inline` that allows users to press the Enter key to submit the form in `textbox.tsx`.

The feature is showcased in `textbox.py` by adding a `ui.textbox` that submits when pressing the Enter key:

```py
ui.textbox(name='textbox_enter', label='Submit the form when pressing the Enter key', trigger_on_enter=True),
```

https://user-images.githubusercontent.com/7029582/206616450-46c8594d-6be7-4664-8d54-f6684e6bfb7a.mov

Unit tests were added in `textbox.test.tsx` as follows:
```js
it('Doesn\'t submit when the Enter key is pressed by default', () => {
   const { getByTestId } = render(<XTextbox model={{ ...textboxProps }} />)
   userEvent.type(getByTestId(name), '123{Enter}')
   expect(pushMock).not.toBeCalled()
 })

it('Submits when the Enter key is pressed using the trigger_on_enter option', () => {
   const { getByTestId } = render(<XTextbox model={{ ...textboxProps, trigger_on_enter: true }} />)
   userEvent.type(getByTestId(name), '123{Enter}')
   expect(pushMock).toBeCalled()
 })
```

Closes #372.

### DIFF
--- a/py/examples/textbox.py
+++ b/py/examples/textbox.py
@@ -40,6 +40,7 @@ async def serve(q: Q):
                        placeholder='I am disabled'),
             ui.textbox(name='textbox_multiline', label='Multiline textarea', multiline=True),
             ui.textbox(name='textbox_spellcheck_disabled', label='Spellcheck disabled', spellcheck=False),
+            ui.textbox(name='textbox_enter', label='Submits the textbox value on Enter key', icon='Search'),
             ui.button(name='show_inputs', label='Submit', primary=True),
         ])
     await q.page.save()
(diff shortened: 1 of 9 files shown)


mturoci (commented):
Thanks @sherrychen2. We need to discuss this with @lo5 and get back to you, but the implementation will most likely need to be adhered to https://github.com/h2oai/wave/issues/348#issuecomment-737788217 instead of introducing a new `trigger_on_enter` attr.

Issue #1693, opened 2022-11-11
Implement an image zoom inside the image popup.


marek-mihok:
Based on the [feedback](https://github.com/h2oai/label-genie/issues/360#issuecomment-1305464397) from label genie.

It would be great to have an image zoom when clicking on the image inside the image popup similar to this:

https://user-images.githubusercontent.com/23740173/201366543-bf0d7cea-d8c5-4570-ada6-b83837d2ca79.mov

jding712:
Hi, my team member would like to work on this issue if it is still available.

nhadlaw:
I am working with @jding712 on this issue

jding712:
@marek-mihok Hi, would you like this feature to be implemented on image popups or on all images? It would be great if you could direct us to the files that would be related to this feature.

PR #1755, opened 2022-12-11
Implemented an image zoom inside image popup

nhadlaw:
This pull request resolves issue #1693 by adding an image zoom to the image pop-up screen when a user clicks on the image. When a user clicks again, the image zooms out going back to normal.

### DIFF
--- a/ui/src/parts/lightbox.tsx
+++ b/ui/src/parts/lightbox.tsx
@@ -18,6 +18,9 @@ import { stylesheet, style } from 'typestyle'
 import * as Fluent from '@fluentui/react'
 import { clas, cssVar } from '../theme'
 import { getColorFromString, isDark } from '@fluentui/react'
+import styled, { keyframes } from "styled-components"
+
+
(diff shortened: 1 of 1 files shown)


mturoci (commented):
Thanks @nhadlaw! I need to think about what the proper UX should be, whether we want a click or mouse scroll.

One thing is for sure though: When scrolling, all the elements except the image should be hidden to prevent odd positioning (see screenshots below).
Screen Shot 2022-12-12 at 3 38 53 PM

mturoci:
Let's go with zoom on scroll and reset on double-click. Moreover, to make the UX good, we also need a way to move the zoomed area - can be the same as 


https://user-images.githubusercontent.com/64769322/207278012-e3a4bb96-5e0c-4809-8b0c-ba1b4c05d885.mov

mturoci:
@nhadlaw are you going to continue to work on this or may I close this PR?

mturoci:
Closing due to inactivity.

Issue #1689, opened 2022-11-08
Display active table filter count


mturoci:
via @shihan007

The table currently supports filtering, but to see how many filters are active per column, one needs to click it and see which filter options are checked.

Instead, we could display the filter count in the top right corner relative to the chevron icon so that it's visible without clicking the chevron.

<img width="398" alt="image" src="https://user-images.githubusercontent.com/64769322/217746938-31c72191-c0aa-429c-8f90-eea43bf7067a.png">

Implementation note: If there are > 9 filters active, display 9+ instead of the actual number to prevent too big count numbers and related spacing problems.

sherif-ffs:
Hi @mturoci I'm new to open source. Could I take up this issue?

jding712:
Hi me and other friends would like to take on this issue.

mturoci:
Go ahead @jding712!

mturoci:
The issue is up for grabs again for anyone interested.

Stijn-Kuijper:
Hi,

I was looking into fixing this, but as far as I can tell Fluent UI does not allow you to add custom elements to an IColumn. A different solution that is possible would be to use the Tooltip to show the amount of filters when hovering over the column header label.

I've made a start at this approach, and it would look something like this: 
![image](https://user-images.githubusercontent.com/25306980/228533921-b3af6c95-8a5a-4342-b712-b60ab1f0a935.png)

Would this be a satisfactory solution? If so, I can continue to fix this issue. If not I'm afraid it will be too difficult for me to fix.

mturoci:
Thanks for your interest @Stijn-Kuijper!

> Fluent UI does not allow you to add custom elements to an IColumn

Looking at the Fluent API, it seems like `onRenderHeader` might be what is needed.
![image](https://user-images.githubusercontent.com/64769322/228535949-792218a8-1cb9-4ae9-9670-959d4ecde408.png)

> use the Tooltip to show the amount of filters when hovering

Although this is an option, it may not be the optimal UX we are striving for as it would require an additional hover to show the counts.

Stijn-Kuijper:
Hi, thanks for the quick reply!

`onRenderHeader` is actually what I am using right now to use the tooltip. However, this seems to only replace the content/label of the column header, rather than the entire header element. I tried adding a `div` element there that would have two children: one with the label and one with the filter amount, but this did not seem like an appropriate way of going about doing this, which is why I switched over to the tooltip approach. 

[...]

PR #1760, opened 2022-12-13
Resolves Issue #1689: Adding filter count to column headers

SiddharthParatkar:
This PR resolves issue #1689. When a filter is selected for a column that is instantiated with `filterable=True`, this feature will automatically add a number for how many filters are applied to that column. For 0 filters, no number will be shown; for more than 9 filters, '9+' will be shown.

### DIFF
--- a/ui/src/table.tsx
+++ b/ui/src/table.tsx
@@ -305,7 +305,7 @@ const
       <div style={{ padding: 10 }}>
         <Fluent.Text variant='mediumPlus' styles={{ root: { paddingTop: 10, paddingBottom: 10, fontWeight: 'bold' } }} block>Show only</Fluent.Text>
         <Fluent.Text variant='small'>
-          <Fluent.Link onClick={selectAll}>Select All</Fluent.Link> | <Fluent.Link onClick={deselectAll}>Deselect All</Fluent.Link>
+          <Fluent.Link onClick={selectAll}>Select All</Fluent.Link> | <Fluent.Link onClick={deselectAll}>Deselectttttt All</Fluent.Link>
         </Fluent.Text>
         {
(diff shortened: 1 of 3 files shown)


ui/src/table.tsx
```diff
@@ -376,11 +379,26 @@ const
...
+        onFilterChange(filterKey, filterVal, checked)
+        setColumns(columns.map(c => {
+          if (columnFilterCount[c.key] || columnFilterCount[c.key] == 0) {
+            let filterCount = ""
+            filterCount = columnFilterCount[c.key] 
+                            ? columnFilterCount[c.key] > 9 
+                              ? " (9+) " 
+                              : " (" + columnFilterCount[c.key] + ") "
+                            : ""
+            c.name = c.name.substring(0, c.key.length) + filterCount
```
mturoci, line 392: Instead of adding the filter count to the col name, we would like to have it as a separate square in the top right corner relative to the filter chevron so that it's clear that the number belongs to the filters.

ui/src/table.tsx
```diff
@@ -965,7 +984,7 @@ export const
       sort,
       isMultiple,
       setFiltersInBulk
-    }), [filteredItems, groups, expandedRefs, isMultiple, items, m, onFilterChange, selectedFilters, selection, sort, setFiltersInBulk])
```
mturoci, line 968: Why was this removed?

mturoci (changes requested):
Thanks @SiddharthParatkar! Can you also please update the PR description with the screenshots of your changes in action once you are done?

mturoci:
@SiddharthParatkar are you going to finish this or may I close this PR?

SiddharthParatkar:
I don't think I'll be able to finish, sorry.

PR #1765, opened 2022-12-15
perf: Copy files instead of sending over HTTP if Wave app and Wave server are running on the same machine. #982

### LINKED ISSUE #982: `q.site.upload` is slow
#### Wave SDK Version, OS

0.17.0, Linux

#### Actual behavior

We are dealing with some large files (predictions, model weights etc.) with a couple hundred MB, which we want to make available for download.

I'm running Wave locally, and with the following code:

```python
import time

@app("/demo")
async def serve(q: Q):
    start = time.time()

    (url,) = await q.site.upload(["data.dump"])

    upload_time = time.time() - start
    print(f"upload time: {upload_time}")

    q.page["meta"].redirect = f"http://localhost:10101/{url}"

    await q.page.save()
```

I get

```
$ head -c 100MB < /dev/urandom > data.dump
$ wave run app
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     Started reloader process [28044] using statreload
INFO:     Started server process [28046]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     127.0.0.1:37576 - "POST / HTTP/1.1" 200 OK
upload time: 33.335158348083496
```

i.e. it takes ~ 33 seconds for a 100MB file to start downloading. 

#### Expected behavior

Since the file only has to be copied to a location on the same machine, I'd expect the download to start almost immediately.

#### Steps To Reproduce

0. Start `waved` locally.

1. `head -c 100MB < /dev/urandom > data.dump`

2. Write `app.py`:
```python
import time

@app("/demo")
async def serve(q: Q):
    start = time.time()

    (url,) = await q.site.upload(["data.dump"])

    upload_time = time.time() - start
    print(f"upload time: {upload_time}")

    q.page["meta"].redirect = f"http://localhost:10101/{url}"

    await q.page.save()
```
3. `wave run app`

mturoci:
This is a rough proposal for further discussion. If approved, will finish the rest of the sync/async upload methods.

Rationale: When uploading files, we send them over HTTP with serialization/deserialization overhead that slows the whole method. However, most of the Wave apps run both Wave server (waved) and Wave app on the same machine which means a simple file copy should be enough.

## Perf testing

HTTPX 0.16.1 is the original version from the time of creating issue #982.

| File size |  HTTPX 0.16.1 | HTTPX 0.23.1 (latest - Mac) | cp (Mac) | robocopy (Windows) | xcopy (Windows) | HTTPX 0.23.1 (latest - Windows) |
|:---------:|:-------------:|:---------------------------:|:--------:|:------------------:|:---------------:|:-------------------------------:|
|    1GB    |     ~8.5s     |            ~4.3s            |   ~14ms  |        ~50ms       |      ~800ms     |              ~5.8s              |
|   10 GB   | 10min timeout |             ~40s            |   ~10ms  |       ~150ms       |       ~75s      |          10min timeout          |

Although we have public/private dirs, the concept of serving files and figuring out what is the correct URL can be troublesome for non-software engineering users so I would suggest keeping the current q.site.upload as is.

This change would impact most (if not all) Wave users who use `q.site.upload`.

Wdyt @lo5?

Closes #982


lo5 (commented):
The implementation looks good. 

However, because the implementation unconditionally resorts to copy-if-local, it's possible we might hit some unforeseen OS permissions/config where the copying might fail. 

Possible safeguards are:
A. Fallback to HTTP if local copy fails (try-except).
B. Enable copy-if-local behavior only if an env var is set.
C. Enable copy-if-local by default, but allow disabling behavior if an env var is set (provides an escape hatch if existing installations get hosed).

Thoughts?

mturoci:
Thanks for the suggestions @lo5! A combination of A and C seems like the way to go for me. Will do.

mturoci:
The PR is done.

* Tested against both Windows and Unix.
* Made async (to not block the main thread).

@lo5 can you please let me know what you think about naming? (`H2O_WAVE_NO_COPY_UPLOAD`, `H2O_WAVE_WAVED_DIR`). Once resolved, can be merged.

website/docs/configuration.md
```diff
@@ -239,6 +239,14 @@ Disable/enable Wave server boot during `wave run`. Defaults to `false`. Availabl
 
 The base URL (path prefix) to be used for resolving relative URLs (e.g. /foo/ or /foo/bar/, without the host (default "/"). If you run your Wave server (waved) and Wave app separately, it's necessary to set this env variable for both.
 
+### H2O_WAVE_NO_COPY_UPLOAD
+
+If your Wave server and Wave app run on the same machine, `q.site.upload` simply copies the files instead of pushing them through an HTTP request due to performance reasons (extra serialization). However, if you would like to force HTTP every time, you can se `H2O_WAVE_NO_COPY_UPLOAD` to `1` or `t` or `true`.
+
+### H2O_WAVE_WAVED_DIR
+
+Tell your Wave app the location of your Wave server if running on the same machine. Useful for performance optimizations during file uploads.
```
lo5, line 248:
```suggestion
Provides the location of the Wave server's root directory to the Wave app, if both the server and the app are running on the same machine. Useful for performance optimizations during file uploads. Makes `q.site.upload()` copy files instead of using HTTP requests.
```

lo5 (changes requested):
Names look fine. Minor changes to the docs.

PR #1770, opened 2022-12-16
feat: Forward HTTP headers to the Wave app. #1704

### LINKED ISSUE #1704: Forward all inbound HTTP headers to apps for additional processing
See https://github.com/h2oai/wave/discussions/1383#discussioncomment-4163733

mturoci:
Let's use this PR as a starting point for discussion. Once we agree on the implementation, the docs will be added as well.

wdyt about the proposal in this PR @lo5 ? It's quite straightforward I would say.

Closes #1704


socket.go
```diff
@@ -51,7 +51,7 @@ func (s *SocketServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	client := newClient(getRemoteAddr(r), s.auth, session, s.broker, conn, s.editable, s.baseURL)
+	client := newClient(getRemoteAddr(r), s.auth, session, s.broker, conn, s.editable, s.baseURL, &r.Header)
```
lo5, line 54: I'm hazy about how this works. How are the HTTP headers from the original GET (index.html) request preserved on this request? Does the browser do that automatically?

py/h2o_wave/server.py
```diff
@@ -135,6 +137,8 @@ def __init__(
         """The route served by the server."""
         self.auth = auth
         """The authentication / authorization details of the user who initiated this query."""
+        self.forwarded_headers = forwarded_headers
+        """Forwarded HTTP headers from the initial browser connection."""
```
lo5, line 141:
```suggestion
        """Original HTTP connection headers forwarded from the Wave server to this application."""
```

socket.go
mturoci: > How are the HTTP headers from the original GET (index.html) request preserved on this request?

They are not. If original GET request headers are needed, then we would need to store them temporarily on Wave server and send them with the initial post Request to the Wave app. This should not be a problem.

The possible problem can arise during WS connection initialization since it's not possible to specify additional headers when initializing WS, only cookies are appended automatically. My guess is the proxy would reject such request as unauthorized. We can discuss on tomorrow's sync.

MaksymSofer:
ubuntu 20.04.5 LTS has 1.15 which not able to build waved
manually replaced go with 1.19.5 and  everything works like a magic
waiting for this to be merged into main

mturoci:
Added support for the headers from the initial `index.html GET` request as well. The general architecture:

* Initial `index.html GET` generates a clientID and sets it as a cookie.
* The headers from the GET request are stored in a Map<clientID, headers> within `Site` (might change the location).
* During websocket connection, clientID cookie is retrieved and used as previous `c.id`.
* When forwarding new changes to the Wave app, headers stored in Map and headers from the WS connection are merged and sent to the Wave app.
* Wave app manages the forwarded headers in a similar fashion as `q.client` is handled - Dict<clientID, headers>.
image

Needs discussion
* General thoughts about this approach?
* Is exposing `clientID` in the cookie fine security-wise?
* Should the forwarded headers be sent as headers or as a JSON payload instead? Probably the latter.
* Are there any headers we might want to blacklist? E.g. `Authorization`?

cc @lo5

lo5:
The above approach seems like the only way to address the issue. Security-wise, hard to say - since auth needs to be turned off for this to even work.

> Should the forwarded headers be sent as headers or as a JSON payload instead? 

If I'm not mistaken, there's already a mechanism in place to add headers to the app-request. Best to continue using that mechanism.

That said, I'm not sure it's even worth addressing this issue in the first place. How many users are using the Wave server with auth turned off?

mturoci:
> If I'm not mistaken, there's already a mechanism in place to add headers to the app-request. Best to continue using that mechanism.

Yep, the problem is the existing mechanism doesn't support adding headers dynamically during runtime, only during startup as a config option.

> How many users are using the Wave server with auth turned off?

I would say most of the apps. The reason this PR exists is so that Wave could also be used in (usual enterprisy) environments where auth is not handled by OIDC directly, but via proxy (proxy-based auth). I would love to solve this so that Wave could level up from single-person / OIDC-only envs.

On the other hand, I agree that the implementation is somewhat clunky.

client.go
```diff
@@ -132,7 +134,19 @@ func (c *Client) listen() {
 				echo(Log{"t": "query", "client": c.addr, "route": m.addr, "error": "service unavailable"})
 				continue
 			}
-			app.forward(c.id, c.session, m.data)
+
+			// TODO: Ugly. Find a better way to wrap m.data into {"args": m.data}.
```
zoido, line 138: you can define a type

```go
type body {
	Args map[string]string `json:"keys"`
}
```

client.go
```diff
@@ -151,7 +165,26 @@ func (c *Client) listen() {
...
-				app.forward(c.id, c.session, boot)
+				// TODO: Ugly. Find a better way to wrap m.data into {"args": m.data}.
+				body := map[string]map[string]string{}
+				argsMap := map[string]string{}
+				json.Unmarshal(boot, &argsMap)
+				body["args"] = argsMap
+
+				if getRequestHeaders, ok := c.broker.site.clientIDToHeaders[c.id]; ok {
+					headersMap := map[string]string{}
+					mergeHeaders(headersMap, *getRequestHeaders, *c.header)
```
zoido, line 176: QQ: Is there something more behind creating the destination result here instead of allocating it in the merge function?

Just asking because because I find it unusual.

client.go
```diff
@@ -151,7 +165,26 @@ func (c *Client) listen() {
...
+					body["headers"] = headersMap
+					// Send headers just once as they are stored in Wave app anyway.
+					delete(c.broker.site.clientIDToHeaders, c.id)
```
zoido, line 179: marshaling below can fail. Shouldn't we remove the headers only after we are sure that the operation succeeded?

py/h2o_wave/server.py
```diff
@@ -311,10 +315,16 @@ async def _receive(self, req: Request):
         access_token = req.headers.get('Wave-Access-Token')
         refresh_token = req.headers.get('Wave-Refresh-Token')
         session_id = req.headers.get('Wave-Session-ID')
+
+        body = await req.json()
+        forwarded_headers = body.get('headers', None)
+        if forwarded_headers:
+            # TODO: Think about how to clean this up.
+            self._forwarded_headers[client_id] = forwarded_headers
```
zoido, line 323: We send headers as CSV. What if split them here to have them in more comfortable form.

client.go
```diff
@@ -132,7 +134,19 @@ func (c *Client) listen() {
 				echo(Log{"t": "query", "client": c.addr, "route": m.addr, "error": "service unavailable"})
 				continue
 			}
-			app.forward(c.id, c.session, m.data)
+
+			// TODO: Ugly. Find a better way to wrap m.data into {"args": m.data}.
+			body := map[string]map[string]string{}
+			argsMap := map[string]string{}
+			json.Unmarshal(m.data, &argsMap)
```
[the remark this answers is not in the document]
mturoci: Done.

client.go
mturoci: Thanks! Exactly what I was looking for.

client.go
mturoci: This is outdated now, but agree that its unusual. No specific reason.

client.go
mturoci: Outdated now, headers are now stored on WS client instance which means they are automatically garbage collected when the connection terminates.

py/h2o_wave/server.py
mturoci: Hm.. not sure I follow, maybe outdated as well since headers are no longer sent as headers, but rather as a JSON payload which is then unmarshalled into python dict, which is the most comfortable IMO. But happy to discuss further if I am missing something.

zoido: outdated. I'll take another look at the whole PR.

mturoci:
* Reverted to the original approach - only forward the headers from initial WS connection for these reasons:
   * Less overhead - no need for site-wide storage on server.
   * `index.html GET` headers can be outdated during WS reconnection - e.g. unreliable internet connection. Forwarding only websocket ones makes sure the headers are always fresh.
   * No need to expose client-id in a public cookie.

@lo5 can you please go over API naming?

py/h2o_wave/server.py
```diff
@@ -112,6 +112,7 @@ def __init__(
             client_state: Expando,
             args: Expando,
             events: Expando,
+            headers: Dict[str, str]
```
zoido, line 115: `http.Header` is 
```go
type Header map[string][]string
```
https://pkg.go.dev/net/http#Header

this is equivalent to `map[string]string`

`self._headers` below is equivalent to `map[string]map[string]string`

I think that the annotation is mismatched.

WDYT?

Am I missing something?

app.go
```diff
@@ -62,6 +62,29 @@ func newApp(broker *Broker, mode, route, addr, keyID, keySecret string) *App {
...
+
+	req.SetBasicAuth(app.keyID, app.keySecret)
+	req.Header.Set("Wave-Client-ID", clientID)
+
+	resp, err := app.client.Do(req)
+	if err != nil {
+		return fmt.Errorf("request failed: %v", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != 200 {
```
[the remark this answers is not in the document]
mturoci: I would say it's not a big deal since we are in charge of sending the response from python which is only capable of sending 200, 401 or 400 meaning even if we allow full 2xx range, it would never execute beyond 200.

py/h2o_wave/server.py
mturoci: Good catch! I forgot that HTTP spec allows multiple values for a single key hence the proper type shall be `Dict[str, List[str]]`.

> self._headers below is equivalent to map[string]map[string]string

Actually it's `map[string][]string`. Since we retrieve the headers from JSON payload via `body.get('headers', None)` so we go from `map[string]map[string][]string` to `map[string][]string`

website/docs/configuration.md
```diff
@@ -23,6 +23,7 @@ You can either configure your Wave server via environment variables (works for b
...
 | H2O_WAVE_DEBUG [^1]                     | -debug                                | enable debug mode (profiling, inspection, etc.)                                                                                                                                                                                                                                                                      |
 | H2O_WAVE_EDITABLE [^1]                  | -editable                             | allow users to edit web pages                                                                                                                                                                                                                                                                                        |
+| H2O_WAVE_FORWARDED_HTTP_HEADERS        | -forwarded-http-headers string         | comma-separated list of case insesitive HTTP header keys to forward to the Wave app from the browser WS connection. If not specified, defaults to '*' - all headers are allowed. If set to an empty string, no headers are forwarded.                                         |
```
lo5, line 26:
```suggestion
| H2O_WAVE_FORWARDED_HTTP_HEADERS        | -forwarded-http-headers string         | comma-separated list of case-insensitive HTTP header keys to forward to the Wave app from the browser WS connection. If not specified, defaults to '*' - all headers are allowed. If set to an empty string, no headers are forwarded.                                         |
```

lo5 (changes requested):
Minor typo in docs.

Apart from that - fingers crossed. Hopefully no existing functionality is broken.

zoido (approved):

mturoci:
Thanks for the review @zoido @lo5!

PR #1795, opened 2023-01-17
feat: Keyboard shortcuts for ui.image_annotator

### LINKED ISSUE #1461: Image annotator box dragging anomalies
#### Wave SDK Version, OS

wave 0.22 macOS 12.3

#### Actual behavior
Using `ui.image_annotator` unexpected sudden dragging boxes around the edges could resize them or create new ones.

#### Expected behavior

* Boxes should keep their original size during dragging
* Dragging should not create new boxes

#### Steps To Reproduce
https://user-images.githubusercontent.com/5063584/171141105-da3135a6-05cf-48c7-b2b8-7d0345c67726.mov

### LINKED ISSUE #1596: Keyboard shortcuts
Keyboard shortcuts for the image annotation component.

Keyboard shortcuts will speed up image annotation tasks significantly. The shortcuts can be a part of the corresponding components (built-in by default) without any extra API. The exception would be actions, that need to be handled by the app dev, e.g. next/prev image. These could be registered via 'events' to keep API consistent. For discoverability a question mark with a tooltip including the shortcut table would work.

A few examples
image

marek-mihok:
This contribution brings the image annotating experience to a whole new level by introducing keyboard shortcuts!

https://user-images.githubusercontent.com/23740173/212863826-6560a3f2-ee09-49dd-bc61-628c4e8ad3f7.mov

|         **Key**          |                         **Description**                       |
|:------------------------:|:-----------------------:|
|          **a**           |                        Select all shapes                      |
|          **c**           |                      Copy selected shapes                     |
|          **v**           |                      Paste selected shapes                    |
|          **d**           |                     Delete selected shapes                    |
|    **Shift + Click**     |       Select multiple shapes when in the selection mode       |
| **Arrow keys (↑ ↓ → ←)** | Move selected shapes by 1px (or 10px while holding Shift key) |
|  **Ctrl + Mouse wheel**  |                          Zoom in/out                          |
|        **Enter**         |                    Finish drawing polyshape                   |
|      **Backspace**       |                  Delete last polyshape vertex                 |
|         **Esc**          |                      Cancel ongoing task                      |
|          **l**           |                          Toggle label                         |
|          **b**           |                    Toggle drawing function                    |
|          **r**           |                     Select rectangle tool                     |
|          **p**           |                       Select polygon tool                     |
|          **s**           |                     Activate selection tool                   |

And this is not all. It also brings the following:
-  adds the image zoom feature
- makes polygon to respect canvas boundaries
- rounds the `wave.args` coordinates
-  fixes #1461.

Closes #1461 
Closes #1596


ui/src/image_annotator.test.tsx
```diff
@@ -669,6 +669,339 @@ describe('ImageAnnotator.tsx', () => {
...
+        await waitForLoad()
+        const canvasEl = container.querySelectorAll('canvas')[1]
+        expect(wave.args[name]).toMatchObject(items)
+        fireEvent.click(getByText('Remove all'))
+        fireEvent.wheel(canvasEl, { deltaY: 1, ctrlKey: true })
+        expect(wave.args[name]).toMatchObject([])
+      })
+    })
+
+    describe('Active shape switching', () => {
```
mturoci, line 919: Would be good to also add tests if cursors are changed with shape as well.

mturoci (changes requested):
Thanks @marek-mihok. This is the first round of review since this PR is trying to solve multiple issues at once. As discussed, it's always better and safer to create more smaller PRs instead of gigantic one.

The colors for shortcuts popup seem odd, I would expect the popup to be light for light theme and dark for dark theme.

image

In addition:
* Why is ctrl required during zooming? Could we do without it?
* Same point ^^ for moving when zoomed.

ui/src/image_annotator.tsx
```diff
@@ -109,10 +114,52 @@ const
...
+    { key: 'Enter', description: 'Finish drawing polyshape' },
+    { key: 'Backspace', description: 'Delete last polyshape vertex' },
+    { key: 'Esc', description: 'Cancel ongoing task' },
+    { key: 'l', description: 'Toggle label' },
+    { key: 'b', description: 'Toggle drawing function' },
+    { key: 'r', description: 'Select rectangle tool' },
+    { key: 'p', description: 'Select polygon tool' },
+    { key: 's', description: 'Activate selection tool' }
+  ],
+  eventToCursor = (e: React.MouseEvent, rect: DOMRect, scale: F, position: ImageAnnotatorPoint) =>
```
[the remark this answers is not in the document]
marek-mihok: Resolving as we discussed it on the call.

ui/src/image_annotator.test.tsx
```diff
@@ -669,6 +669,339 @@ describe('ImageAnnotator.tsx', () => {
...
+      it('Use "b" to switch the active shape', async () => {
+        const { container } = render(<XImageAnnotator model={model} />)
+        await waitForLoad()
+        const canvasEl = container.querySelectorAll('canvas')[1]
+        fireEvent.keyDown(canvasEl, { key: 'b' })
+
+        await waitFor(() => expect(document.querySelector('[class*="is-checked"]')).toHaveAttribute('title', 'Rectangle'))
```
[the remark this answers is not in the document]
marek-mihok: I have this kind of test already `Change tag of the selected shape when changing the active tag`.
But this one you are referring to is regarding changing the active boundary (select, polygon, rect), not about labels.

ui/src/image_annotator.test.tsx
```diff
@@ -669,6 +669,339 @@ describe('ImageAnnotator.tsx', () => {
...
+        await waitForLoad()
+        const canvasEl = container.querySelectorAll('canvas')[1]
+        fireEvent.click(canvasEl, { clientX: 50, clientY: 50 })
+        fireEvent.mouseDown(canvasEl, { clientX: 50, clientY: 50, buttons: 1 })
+        fireEvent.mouseMove(canvasEl, { clientX: 55, clientY: 55, buttons: 1 })
+        fireEvent.keyDown(canvasEl, { key: 'b' })
+        fireEvent.mouseMove(canvasEl, { clientX: 60, clientY: 60, buttons: 1 })
+        fireEvent.click(canvasEl, { clientX: 60, clientY: 60 })
+
+        expect(wave.args[name]).toMatchObject([{ tag: 'person', shape: { rect: { x1: 15, x2: 105, y1: 15, y2: 105 } } }, polygon])
```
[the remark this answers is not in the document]
marek-mihok: The reason why I am expecting at the end is that the `wave.args` are updated in the `onClick` handler. I think expecting at the end is fine - we know that moving has stopped when pressing `b` key. If it was not the case, the rectangle would move by 10 on both axes, not by 5. The only case I can think this test could bring a false positive result is that rectangle is not moving before pressing `b` but only after. But we can also handle it by changing `clientX` and `clientY` to let's say 70 instead of 60.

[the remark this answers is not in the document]
marek-mihok: > Let's add expect after the first mouseMove
Ok, my previous comment is irrelevant. I am setting the `wave.args` when pressing `b` as well. So I've added the expect after `keyDown` and it works.

ui/src/image_annotator.test.tsx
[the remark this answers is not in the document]
mturoci: I see, makes sense although the internal implementation point still stands.

ui/src/image_annotator_polygon.ts
```diff
@@ -1,41 +1,113 @@
...
   }
 
   resetDragging() {
+    // Update the boundaries of the polygon when point dragging ends.
+    const draggedPolygon = this.draggedShape?.shape.polygon
+    if (this.draggedPoint && draggedPolygon) {
+      this.boundaryRect = draggedPolygon.boundaryRect
+      this.updateBoundaryRect(this.draggedPoint)
+      draggedPolygon.boundaryRect = this.boundaryRect
+      this.boundaryRect = undefined
```
mturoci, line 24: Make pass the `boundaryRect` you want to mutate as a param to `updateBoundary` to avoid these gymnastics.

Side question: Why do we need to update the boundaries when resetting the dragging? Isn't the latest boundary value from onMove enough?

ui/src/image_annotator_polygon.ts
```diff
@@ -1,41 +1,113 @@
...
+      newPolygon = {
+        shape: {
+          polygon: {
+            vertices: [...this.currPolygonPoints],
+            boundaryRect: this.boundaryRect
+          }
+        },
+        tag
+      },
+      isPolygon = this.currPolygonPoints.length > 2
```
mturoci, line 68: Shouldn't we check this as the first thing and do an early return if we do have < 2 points?

ui/src/image_annotator_polygon.ts
```diff
@@ -1,41 +1,113 @@
...
+      boundaryRect.x2 = x2 + moveX
+    }
+    if (moveY) {
+      boundaryRect.y1 = y1 + moveY
+      boundaryRect.y2 = y2 + moveY
     }
-    if (this.currPolygonPoints.length) this.drawLine(cursor_x, cursor_y)
 
-    this.currPolygonPoints.push({ x: cursor_x, y: cursor_y })
+    movedPolygon.vertices.forEach(p => {
```
mturoci, line 107:
```suggestion
    if (moveX && moveY) movedPolygon.vertices.forEach(p => {
```

ui/src/image_annotator_polygon.ts
```diff
@@ -45,18 +117,20 @@ export class PolygonAnnotator {
       return
     }
 
+    // TODO: Find a way to do this without iterating through all points.
```
mturoci, line 120:
```suggestion
    // TODO: Calculate this in mouseDown handler and store the current result for further use here.
```

ui/src/image_annotator_rect.ts
```diff
@@ -72,6 +72,31 @@ export class RectAnnotator {
...
 
+  move = (dx: U, dy: U, shape?: DrawnShape) => {
+    // Prevent moving behind image boundaries.
+    const movedRect = (shape || this.movedRect)?.shape.rect
+    if (!movedRect) return
+
+    const
+      { width, height } = this.canvas,
+      { x1, x2, y1, y2 } = movedRect,
+      xMin = x1 < x2 ? x1 : x2,
```
mturoci, line 83:
```suggestion
      xMin = Math.min(x1, x2),
```

ui/src/image_annotator.tsx
```diff
@@ -20,6 +20,8 @@ export interface ImageAnnotatorPoint {
 export interface ImageAnnotatorPolygon {
   /** List of polygon points. */
   vertices: ImageAnnotatorPoint[]
+  /** The polygon's boundary rectangle. */
+  boundaryRect?: ImageAnnotatorRect
```
mturoci, line 24: This is an implementation detail and shouldn't be exposed in the public API.

ui/src/image_annotator.tsx
```diff
@@ -109,27 +115,104 @@ const
...
+  getPolygonBoundaries = (vertices: ImageAnnotatorPoint[]) => {
+    if (!vertices.length) return
+    let minX: U, minY: U, maxX: U, maxY: U
+    vertices.forEach(({ x, y }) => {
+      if (minX === undefined || x < minX) minX = x
+      if (minY === undefined || y < minY) minY = y
+      if (maxX === undefined || x > maxX) maxX = x
+      if (maxY === undefined || y > maxY) maxY = y
+    })
+    return { x1: minX!, y1: minY!, x2: maxX!, y2: maxY! }
```
mturoci, line 214: Should be the same

```suggestion
    const [firstPoint] = vertices
    return vertices.reduce((acc, { x, y }) => {
      if (x < acc.x1) acc.x1 = x
      if (y < acc.y1) acc.y1 = y
      if (x > acc.x2) acc.x2 = x
      if (y > acc.y2) acc.y2 = y
      return acc
    }, { x1: firstPoint.x, y1: firstPoint.y, x2: firstPoint.x, y2: firstPoint.y })
```

ui/src/image_annotator.tsx
```diff
@@ -248,56 +388,75 @@ export const XImageAnnotator = ({ model }: { model: ImageAnnotator }) => {
...
+            // If left mouse btn is not held during moving, ignore.
+            if (e.buttons !== 1) break
+            const shape = focused?.shape.rect ? rectRef.current : polygonRef.current
+            // TODO: Move all selected shapes at once.
+            if (shape) {
+              shape.onMouseMove(cursor_x, cursor_y, focused, intersected, clickStartPosition)
+              // Deselect all other shapes when one starts moving.
+              if (drawnShapes.some(ds => ds.isFocused && ds !== focused)) {
+                setDrawnShapes(drawnShapes => drawnShapes.map(ds => ({ ...ds, isFocused: intersected === ds })))
+              }
```
mturoci, line 438: No need to go through the drawnShapes twice, we can update anyway.

```suggestion
                setDrawnShapes(drawnShapes => drawnShapes.map(ds => ({ ...ds, isFocused: intersected === ds })))```

File: ui/src/image_annotator.tsx [hunk 13]
```diff
@@ -339,23 +501,130 @@ export const XImageAnnotator = ({ model }: { model: ImageAnnotator }) => {
...
+      else if (e.key === 'Control' && canvasRef.current && zoom > 1) {
+        canvasRef.current.style.cursor = 'grab'
+      }
+      else if (e.key === 'a') {
+        setDrawnShapes(shapes => shapes.map(s => ({ ...s, isFocused: true })))
+        redrawExistingShapes()
+      }
+      else if (e.key === 'c') {
+        const selectedShapes = drawnShapes.filter(s => s.isFocused)
+        if (selectedShapes.length) clipboardRef.current = JSON.stringify(selectedShapes)
```
mturoci, line 585: Why is serialization needed here?

ui/src/image_annotator.tsx
```diff
@@ -339,23 +501,130 @@ export const XImageAnnotator = ({ model }: { model: ImageAnnotator }) => {
...
+      }
+      else if (e.key === 'Delete') {
+        const newShapes = drawnShapes.filter(s => !s.isFocused)
+        if (newShapes.length !== drawnShapes.length) {
+          setDrawnShapes(newShapes)
+          setWaveArgs(newShapes)
+          redrawExistingShapes()
+        }
+      }
+      // Remove last polygon vertice.
```
mturoci, line 606:
```suggestion
      // Remove last polygon vertex.
```

mturoci (changes requested):
Is this the final solution though? Seems like there are a couple of critical TODOs still left.

image

ui/src/image_annotator_polygon.ts
marek-mihok: This won't work. Did you mean `if(moveX || moveY)` ?

ui/src/image_annotator_polygon.ts
marek-mihok: >  Isn't the latest boundary value from onMove enough?

We update latest boundaries only when you are moving the whole shape. This is for the case when you are dragging a single point (to re-shape the existing polygon), so you don't have the latest boundary values up to date and we are updating it here.

ui/src/image_annotator_polygon.ts
```diff
@@ -1,41 +1,113 @@
 import { F, S, U } from "h2o-wave"
-import { DrawnPoint, DrawnShape, ImageAnnotatorPoint, Position } from "./image_annotator"
+import { DrawnPoint, DrawnShape, ImageAnnotatorPoint, ImageAnnotatorRect, Position } from "./image_annotator"
 import { ARC_RADIUS } from "./image_annotator_rect"
 
 export class PolygonAnnotator {
   private currPolygonPoints: ImageAnnotatorPoint[] = []
+  private boundaryRect: ImageAnnotatorRect | undefined
```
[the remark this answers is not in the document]
marek-mihok: To match the type of 
```ts
export interface ImageAnnotatorPolygon {
  /** List of polygon points. */
  vertices: ImageAnnotatorPoint[]
  /** The polygon's boundary rectangle. */
  boundaryRect?: ImageAnnotatorRect
}
```
but it is not relevant anymore since the prop should not be public facing and needs to be removed.

ui/src/image_annotator_polygon.ts
mturoci: Yep, you are right.

ui/src/image_annotator_polygon.ts
mturoci: I see, makes sense. Thanks!

mturoci (commented):
Thanks @marek-mihok! Looks better. Will finish this up myself.

mturoci:
@marek-mihok I did some small refactors here and there. Could you please do a bit of QA to make sure I didn't break anything? Once approved, we are good to merge.

marek-mihok:
@mturoci, it is good to go!

PR #1808, opened 2023-01-23
ENV setup required when deploying with Uvicorn

### LINKED ISSUE #1807: Update to the Deploying Wave Apps Documentation Needed
### Discussed in https://github.com/h2oai/wave/discussions/1803

<sup>Originally posted by **tolmanam** January 20, 2023</sup>
I don't understand the interactions between the `waved` server and the `wave apps`.  I'd like to be able to spin up the `waved` server and then run my applications with `uvicorn` as described in the [Deploying Wave apps docs](https://wave.h2o.ai/docs/deployment#deploying-wave-apps) but the two commands provided **do not actually seem** to be equivalent.

```
(venv) $ wave run --no-reload --no-autostart foo
```

```
(venv) $ uvicorn foo:main
```

Both seem to spin up the application API, but when I start the app with the `wave` command it also connects up to the running `waved` service and registers the route.  The `uvicorn` command does not.

Am I missing a step?

Thanks.

_versions:_
 * _waved - 0.24.2 20221209122225_
 * _h2o-wave - 0.24.2_
 * _uvicorn - 0.20.0_</div>

tolmanam:
I was just running through the docs and was unable to get the Wave app to connect to the `waved` server when I ran the `uvicorn` command specified in the documentation.

I reached out via the **Discussions** threads and quickly received some help from @mturoci.

The documentation didn't mention the need to set the H2O_WAVE_APP_ADDRESS environment variable (at least not directly on the deployment page).

closes #1807

### DIFF
--- a/website/docs/deployment.md
+++ b/website/docs/deployment.md
@@ -21,7 +21,7 @@ These commands are equivalent:
 ```
 
 ```shell
-(venv) $ uvicorn foo:main
+(venv) $ H2O_WAVE_APP_ADDRESS="http://127.0.0.1:8000" uvicorn foo:main
 ```
 
 For more information, see [uvicorn.org/deployment](https://www.uvicorn.org/deployment/) and [starlette.io/#performance](https://www.starlette.io/#performance).
--- a/website/docs/deployment.md
+++ b/website/docs/deployment.md
@@ -21,7 +21,11 @@ These commands are equivalent:
 ```
 
 ```shell
-(venv) $ H2O_WAVE_APP_ADDRESS="http://127.0.0.1:8000" uvicorn foo:main
+(venv) $ uvicorn foo:main
 ```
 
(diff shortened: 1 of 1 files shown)


website/docs/deployment.md
```diff
@@ -21,7 +21,7 @@ These commands are equivalent:
 ```
 
 ```shell
-(venv) $ uvicorn foo:main
+(venv) $ H2O_WAVE_APP_ADDRESS="http://127.0.0.1:8000" uvicorn foo:main
```
mturoci, line 24: It would be better to explain why this might be needed. See https://github.com/h2oai/wave/discussions/1803#discussioncomment-4757071

mturoci (changes requested):
Thanks @tolmanam! It would be better to provide a more in-depth explanation though.

PR #1810, opened 2023-01-26
feat: Emit event on table selection

### LINKED ISSUE #1654: Trigger option for ui.Table row selection changes

We would like to change the state of some buttons based on table row selection, 
e.g. "Delete selected" => disabled = True if nothing is selected

- `trigger=True` option for table row selection changes

Doable via custom js and wave.emit, but would rather have it natively.

marek-mihok:
Let's welcome the new member to our family of already supported table events, the `select` event!

The `select` event is emitted every time when there is a change in current table rows selection and contains the list of all  selected item names. Here is an example:

https://user-images.githubusercontent.com/23740173/214791759-cdf015db-5c32-4fa4-bf16-0ce8eec8a427.mov

This change also comes with the updated API description:
```ts
export interface Table {
  name: Id
  columns: TableColumn[]
  rows?: TableRow[]
  /** True to allow multiple rows to be selected. */
  multiple?: B
  /** True to allow group by feature. Not applicable when `pagination` is set. */
  groupable?: B
  /** Indicates whether the table rows can be downloaded as a CSV file. Defaults to False. */
  downloadable?: B
  resettable?: B
  height?: S
  width?: S
  values?: S[]
  visible?: B
  tooltip?: S
  groups?: TableGroup[]
  /** Display a pagination control at the bottom of the table. Set this value using `ui.table_pagination()`. */
  pagination?: TablePagination
  /** The events to capture on this table. One of 'search' | 'sort' | 'filter' | 'download' | 'page_change' | 'reset' | 'select'. */
  events?: S[]
}
```

Closes #1654


py/examples/table_events_select.py
```diff
@@ -0,0 +1,36 @@
...
+    q.page['meta'] = ui.meta_card(
+    box='',
+    layouts=[
+        ui.layout(breakpoint='xs', zones=[
+            ui.zone('main', direction=ui.ZoneDirection.ROW, zones=[
+                ui.zone('table', size='50%'),
+                ui.zone('selected', size='50%')
+            ])
+        ])
+    ])
```
mturoci, line 18: Let's keep the examples as simple as possible by using the grid layout as the rest of the examples do.

py/examples/table_events_select.py
```diff
@@ -0,0 +1,36 @@
...
+    if q.args.table:
+        q.page['description'].content = f'{q.args.table}'
```
mturoci, line 20: This should be q.events instead of q.args since we want to showcase clicking the selection, not row submission.

ui/src/table.tsx
```diff
@@ -117,7 +117,7 @@ interface TableGroup {
  * set to True, the form is not submitted automatically, and one or more buttons are required in the form to trigger
  * submission.
  * 
- * If `pagination` is set, you have to handle search/filter/sort/download/page_change/reset events yourself since
+ * If `pagination` is set, you have to handle search/filter/sort/download/page_change/reset/select events yourself since
```
mturoci, line 120: Is the imperative selection supported though?

website/widgets/form/table.md
```diff
@@ -102,6 +102,29 @@ q.page['example'] = ui.form_card(box='1 1 3 4', items=[
 ])
 ```
 
+### Select event
+
+Sometimes it can be useful to update other components immediately based on the current row selection (e.g. enable/disable the submit button).
+This is where the `select` event comes into the game. When the `select` event is registered, it is emitted every time when there is a change
+in the current table row selection. It can be accessed through `q.args.table_name` and contains the list of all selected item names, where `table_name`
```
mturoci, line 109: We should encourage people to access it using `q.events` like all the other Wave events to keep the consistency.

mturoci (changes requested):
Thanks @marek-mihok! A few changes needed.

ui/src/table.tsx
marek-mihok: Good point, it is not.

py/examples/table_events_select.py
```diff
@@ -0,0 +1,27 @@
+# Table / Events / Select
+# Register the `select` #event to emit Wave event on each #table row selection.
+# #table #events #select
+# ---
+from h2o_wave import main, app, Q, ui
+
+@app('/demo')
+async def serve(q: Q):
+    if q.events.table:
+        if q.events.table.select:
```
mturoci, line 10: nit

```suggestion
    if q.events.table and q.events.table.select:
```

website/widgets/form/table.md
```diff
@@ -102,6 +102,29 @@ q.page['example'] = ui.form_card(box='1 1 3 4', items=[
 ])
 ```
 
+### Select event
+
+Sometimes it can be useful to update other components immediately based on the current row selection (e.g. enable/disable the submit button).
+This is where the `select` event comes into the game. When the `select` event is registered, it is emitted every time when there is a change
+in the current table row selection. It can be accessed through `q.events.table_name.select` and contains the list of all selected item names, where `table_name`
+is the `name` of the table.
```
mturoci, line 110:
```suggestion
is the `name` attribute of the table.
```

website/widgets/form/table.md
```diff
@@ -102,6 +102,29 @@ q.page['example'] = ui.form_card(box='1 1 3 4', items=[
 ])
 ```
 
+### Select event
+
+Sometimes it can be useful to update other components immediately based on the current row selection (e.g. enable/disable the submit button).
+This is where the `select` event comes into the game. When the `select` event is registered, it is emitted every time when there is a change
+in the current table row selection. It can be accessed through `q.events.table_name.select` and contains the list of all selected item names, where `table_name`
```
mturoci, line 109:
```suggestion
in the current table row selection. It can be accessed through `q.events.<table_name>.select` and contains the list of all selected item names, where `table_name`
```

mturoci (approved):
LGTM.

PR #1811, opened 2023-01-26
fix: Do not serve JS files with plaintext MIME type on Windows by upgrading go to 1.19.

mturoci:
Needs discussion: Shall we use the same go version for WaveDB as well? cc @lo5

Update needed due to: https://github.com/golang/go/issues/32350


lo5:
> Shall we use the same go version for WaveDB as well?

Yes - thanks!

PR #1814, opened 2023-01-27
feat: Allow single selection for ui.table

### LINKED ISSUE #216: Allow single selection for table
via @pramitchoudhary 

Currently, we only support multiple or none selection:
image

@lo5 was there a specific reason for this?

We could change the current `multiple` prop to `selection` and allow three values: `none`(default), `single` and `multiple`. This would be a breaking change though. Another option would be to simply add `single: B` in addition to `multiple`. The first option seems to be cleaner to me even for the price of a breaking change.

marek-mihok:
With the new `single` and `value` attributes it is now possible to use the _**ui.table**_ as an **advanced single-select** component.

https://user-images.githubusercontent.com/23740173/215066096-2563a3c2-b414-49c6-a906-13b5d9af942e.mov

The updated API is as follows:

```ts
export interface Table {
  name: Id
  columns: TableColumn[]
  rows?: TableRow[]
  /** True to allow multiple rows to be selected. Mutually exclusive with `single` attr. */
  multiple?: B
  /** True to allow group by feature. Not applicable when `pagination` is set. */
  groupable?: B
  /** Indicates whether the table rows can be downloaded as a CSV file. Defaults to False. */
  downloadable?: B
  resettable?: B
  height?: S
  width?: S
  values?: S[]
  visible?: B
  tooltip?: S
  groups?: TableGroup[]
  /** Display a pagination control at the bottom of the table. Set this value using `ui.table_pagination()`. */
  pagination?: TablePagination
  /** The events to capture on this table. One of 'search' | 'sort' | 'filter' | 'download' | 'page_change' | 'reset' | 'select'. */
  events?: S[]
  /** True to allow only on row to be selected at time. Mutually exclusive with `multiple` attr. */
  single?: B
  /** The name of the selected row. If this parameter is set, single selection will be allowed (`isSingle` is assumed to be `True`). */
  value?: S
}
```

Closes #216


ui/src/table.tsx
```diff
@@ -153,6 +153,10 @@ export interface Table {
   pagination?: TablePagination
   /** The events to capture on this table. One of 'search' | 'sort' | 'filter' | 'download' | 'page_change' | 'reset' | 'select'. */
   events?: S[]
+  /** True to allow only on row to be selected at time. Mutually exclusive with `multiple` attr. */
+  isSingle?: B
```
mturoci, line 157: The public API should be snake-cased. However, let's use `single` instead of `is_single` for this case to be consistent with `multiple`.

website/widgets/form/table.md
```diff
@@ -58,19 +58,18 @@ q.page['example'] = ui.form_card(box='1 1 5 3', items=[
 
 ## With selection
 
-If `multiple` is set to False (default), each row in the table is clickable. When a cell in the column with `link=True`
+By default, each row in the table is clickable. When a cell in the column with `link=True`
 (defaults to first column) is clicked or the row is doubleclicked, the form is
 submitted automatically and `q.args.table_name` is set to `[row_name]`, where
 `table_name` is the `name` of the table, and `row_name` is the `name` of the row that was clicked.
 Note that doubleclicking the row if single selection is active results in submission as well.
```
mturoci, line 65: This probably needs to be changed as well if I read the code correctly. Double-clicking works only if the selection mode is none, right?

ui/src/table.test.tsx
```diff
@@ -199,6 +214,19 @@ describe('Table.tsx', () => {
...
+      const { getAllByRole } = render(<XTable model={{ ...tableProps, isSingle: true, events: ['select'] }} />)
+      const radioButtons = getAllByRole('radio')
+
+      fireEvent.click(radioButtons[0])
+      expect(emitMock).toHaveBeenCalledWith(tableProps.name, 'select', ['rowname1'])
+      expect(emitMock).toHaveBeenCalledTimes(1)
+
+      fireEvent.click(radioButtons[1])
+      expect(emitMock).toHaveBeenCalledWith(tableProps.name, 'select', ['rowname2'])
+      expect(emitMock).toHaveBeenCalledTimes(2)
```
mturoci, line 227: Let's reset mock before firing it again and check if has been only called once.

ui/src/table.test.tsx
```diff
@@ -178,7 +183,17 @@ describe('Table.tsx', () => {
...
 
-    it('Fires event - single selection', () => {
+    it('Sets args - single selection', () => {
+      const { getAllByRole } = render(<XTable model={{ ...tableProps, isSingle: true }} />)
+      const radioButtons = getAllByRole('radio')
+
+      fireEvent.click(radioButtons[0])
+      fireEvent.click(radioButtons[1])
+
+      expect(wave.args[name]).toMatchObject(['rowname2'])
```
mturoci, line 193: let's make this more clear
```suggestion
      fireEvent.click(radioButtons[0])
      expect(wave.args[name]).toMatchObject(['rowname1'])
      fireEvent.click(radioButtons[1])

      expect(wave.args[name]).toMatchObject(['rowname2'])
```

website/widgets/form/table.md
```diff
@@ -58,19 +58,18 @@ q.page['example'] = ui.form_card(box='1 1 5 3', items=[
...
 
-If `multiple` is set to `True`, each row in the table is selectable. A row can be selected by
-clicking on it.
-Multiple rows can be selected either by shift+clicking or using marquee selection. When the form
-is submitted, `q.args.table_name` is set to `[row1_name, row2_name, ...]` where `table_name` is the
-`name` of the table, and `row1_name`, `row2_name` are the `name` of the rows that were selected. Note
-that if `multiple` is set to `True`, the form is not submitted automatically and one or more buttons in the form are
-required to trigger submission.
+However, this behavior changes when the multiple or single selection is activated (`multiple` or `isSingle` is set to `True`). In this case each row in the table becomes selectable.
+Also the form is not submitted automatically and one or more buttons in the form are required to trigger submission.
```
mturoci, line 68: Is this tested?

mturoci (changes requested):
Thanks @marek-mihok. Please also document this feature (emitting events) in `table.md`. It seems like it's currently missing if I looked correctly.

website/widgets/form/table.md
marek-mihok: Right. Especially this sentence is confusing:

> Note that doubleclicking the row if single selection is active results in submission as well.

What did you mean by this when there was not single selection supported before?

website/widgets/form/table.md
marek-mihok: When `multiple` or `single` is set to `True`, double click is not submitting the form, but when clicking the cells in column with `link=True` (first column by default), the submit is triggered, which should not happen according to [our current docs](https://wave.h2o.ai/docs/widgets/form/table/#with-selection):

> Note that if multiple is set to True, the form is not submitted automatically and one or more buttons in the form are required to trigger submission.

...but it happens also in the master. That means we should handle this case and [fire onClick logic](https://github.com/h2oai/wave/blob/cc2d18dee0d6784279311f7f6d437145425b056e/ui/src/table.tsx#L512) only when `!(isMultiple && !isSingle)`. We should either edit the implementation to match what is described in our current docs or to change our docs. I think changing the implementation is better solution here. It makes no sense that you can submit multiple selected items by clicking on any of the table cells in column with `link=True`.

website/widgets/form/table.md
mturoci: > What did you mean by this when there was not single selection supported before?

Previously, single selection meant no checkmark, just direct submission.

website/widgets/form/table.md
mturoci: > It makes no sense that you can submit multiple selected items by clicking on any of the table cells in column with link=True

This would indeed make no sense, but it's not what would happen anyway. If you click on a link True col, only the single clicked rowname shall be submitted, which is correct. The docs are technically correct as well but might be confusing. It would be better if they were clear that checkmark selection (single/multi) can only be submitted via external button or something (useful for making a delete action for example) and clicking the row will always submit only the corresponding row - e.g. go to detail.

marek-mihok: Now it should be clear:
image

mturoci (approved):
Thanks @marek-mihok!

Issue #1780, opened 2023-01-09
Allow specifying height and direction for ui.inline


mturoci:
Some users might want to display form items in the middle of the card (e.g. login form). Wave currently provides a `ui.inline` which has `align` and `justify` attributes that could be of help.

However, the current `ui.inline` container only stretches as far as its content. We would need to make it stretch to either custom height (`px`, `vh`, `rem` CSS units) or to fill all the remaining space of the card (by specifying `1` for example).

And since the `ui.inline` is currently a horizontal container only, we might also want to introduce either a differently named component with the same abilities but vertical direction or add a `direction` attr to the existing `ui.inline`.

cc @lo5 for discussion

n-srinidhi:
Hi!Is there any update on this feature?

mturoci:
https://github.com/h2oai/wave/pull/1820. Will be part of the next release.

PR #1820, opened 2023-02-01
feat: Allow specifying height for ui.inline #1780

marek-mihok:
This change adds the `height` prop for ui.inline component.
```ts
interface Inline {
  items: Component[]
  /** Specifies how the individual components are aligned on the vertical axis. Defaults to 'center'. */
  align?: 'start' | 'end' | 'center' | 'baseline'
  inset?: B
  /** Custom height  in `px`, `vh` or `rem`. Use '1' to fill the height of the remaining space. E.g. '100vh', '300px' or '1'. */
  height?: S
}
```
As API description says, the accepted values are the following CSS units: `px`, `vh` or `rem`. It is also possible to fill the all remaining height of the parent component by specifying `1`.

Resolves _ui.inline height_ part of the #1780.

Also regression tests are adjusted since this change defines `display: flex` for cards and could lead to empty space in screenshots. There are 2 changes for regression tests:

- the height is defined for the generated images
- the screenshot of the widget element is taken instead of the form element if there is only one widget in the form

### DIFF
--- a/ui/src/form.tsx
+++ b/ui/src/form.tsx
@@ -175,7 +175,7 @@ interface Inline {
   align?: 'start' | 'end' | 'center' | 'baseline'
   /** Whether to display the components inset from the parent form, with a contrasting background. */
   inset?: B
-  /** Height of the component. Use `px`, `vh` or `rem` CSS units or use '1' to fill all the remaining space. E.g. '100vh', '300px' or '1'. */
+  /** Custom height  in `px`, `vh` or `rem`. Use '1' to fill the height of the remaining space. E.g. '100vh', '300px' or '1'. */
   height?: S
 }
 
@@ -191,8 +191,8 @@ const
   defaults: Partial<State> = { items: [] },
   css = stylesheet({
     card: {
-      padding: 15,
       display: 'flex',
+      padding: 15,
     },
     vertical: {
       display: 'flex',
--- a/ui/src/form.tsx
+++ b/ui/src/form.tsx
@@ -175,6 +175,8 @@ interface Inline {
   align?: 'start' | 'end' | 'center' | 'baseline'
   /** Whether to display the components inset from the parent form, with a contrasting background. */
   inset?: B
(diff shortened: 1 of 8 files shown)


ui/src/form.tsx
```diff
@@ -175,6 +175,8 @@ interface Inline {
   align?: 'start' | 'end' | 'center' | 'baseline'
   /** Whether to display the components inset from the parent form, with a contrasting background. */
   inset?: B
+  /** Custom height  in `px`, `vh` or `rem`. Use '1' to fill the height of the remaining space. E.g. '100vh', '300px' or '1'. */
```
mturoci, line 178:
```suggestion
  /** Height of the inline container. Accepts any valid CSS unit e.g. '100vh', '300px'. Use '1' to fill the remaining card space. */
```

ui/src/form.tsx
```diff
@@ -252,6 +266,8 @@ export const
       justify={m.justify || 'start'}
       align={m.align || 'center'}
       inset={m.inset}
+      // Checks for '1' or positive float value ending with 'px', 'rem' or 'vh'.
+      height={/^\d+(\.\d+)?(px|rem|vh)$|^1$/.test(m.height || '') ? m.height : undefined}
```
mturoci, line 270: This is unnecessary, the docs should be enough.
```suggestion
      height={m.height}
```

mturoci (changes requested):
Thanks @marek-mihok. Can you also add some useful examples showcasing when can this be helpful to the docs?

marek-mihok:
@mturoci sure. Also I will add a better example (login page) once the direction is implemented for ui.inline.

mturoci:
> Also I will add a better example (login page) once the direction is implemented for ui.inline.

It would be better to have examples for both, height-only and height + direction.

mturoci (approved):
LGTM. @lo5 can you do a quick API review, please? Thanks!

lo5 (approved):

Issue #1837, opened 2023-02-15
Add precision attribute to spinbox


mturoci:
via @ledell

### Discussed in https://github.com/h2oai/wave/discussions/1827

<div type='discussions-op-text'>

<sup>Originally posted by **ledell** February 10, 2023</sup>
I can't find a form input field that looks like it's good for floats.  The [spinbox](https://wave.h2o.ai/docs/widgets/form/spinbox) seems useful for integers, but the only thing I can find for float input would be the [textbox](https://wave.h2o.ai/docs/widgets/form/textbox), and then convert to a float on the backend.  Just wanted to check if there's a better built-in solution available.

If there's not something available, maybe we could make a new field type called floatbox which is a wrapper for a textbox that verifies that the input is a float.</div>

### Description

`ui.spinbox` currently allows controlling the precision via the `step` attribute which is not very intuitive. Let's add a new optional `precision` attribute to solve this. If not specified, it should default to the value of `step` to not break the existing apps.

PR #1839, opened 2023-02-15
feat: Make the popup of ui.time_picker visible over the side panel

### LINKED ISSUE #1833: The clock of the time picker isn't visible over the side panel
#### Wave SDK Version, OS

Wave SDK: 0.24.2, OS: MacOs

#### Actual behavior

The clock of the time_picker appears under the side_panel when using that component in the side_panel.

https://user-images.githubusercontent.com/89178049/218627609-961301c5-3620-4c41-b6e6-cda4fd726555.mov

#### Expected behavior

The clock of the time_picker is visible over the side_panel when using that component in the side_panel.

#### Steps To Reproduce

```py


@app('/demo')
async def serve(q: Q):
    if not q.args.show_side_panel:
        q.page['example'] = ui.form_card(box='1 1 2 6', items=[
            ui.button(name='show_side_panel', label='Open side panel', primary=True),
            ui.time_picker(name='timepicker', label="Standard time picker"),
            ui.date_picker(name='datepicker', label="Standard date picker"),
        ])
    else:
        q.page['meta'].side_panel = ui.side_panel(title='Side panel', items=[
            ui.time_picker(name='timepicker', label="Standard time picker"),
            ui.date_picker(name='datepicker', label="Standard date picker"),
        ], width="200px")

    await q.page.save()
```

### LINKED ISSUE #1847: "Dismissed" event is triggered for the side panel when closing ui.notification bar.
#### Wave SDK Version, OS

Wave SDK: Nightly ([76a76fb942757119d31b3e5b11a5945b55206e6b](https://github.com/h2oai/wave/commit/76a76fb942757119d31b3e5b11a5945b55206e6b)) , OS: MacOs

(Testing Fix of https://github.com/h2oai/wave/pull/1813)

#### Actual behavior

`Dismissed` event is triggered for the side panel when closing ui.notification bar.  

https://user-images.githubusercontent.com/12801761/220541091-735be2e9-e5a2-4361-8837-6b95b03dee12.mov

This is visible in the evidence video of the fix. https://github.com/h2oai/wave/pull/1813#issue-1558216706

#### Expected behavior
Its behaviour should be to close the ui.notification bar only and trigger `Dismissed` event for the notification bar instead of the side panel. 

#### Steps To Reproduce

You can run the following code to create the issue:

```py
@app('/')
async def serve(q: Q):
    if not q.client.initialized:
        # Create an empty meta_card to hold the notification bar
        # Display a button to show the notification bar
        q.page['form'] = ui.form_card(box='1 1 2 4', items=[
            ui.button(name="show_side_panel", label="Show side panel")
        ])
        q.client.initialized = True

    # Was the show_notification_bar button clicked?
    if q.args.show_notification_bar:
        # Create a notification bar
        q.page['meta'].notification_bar=ui.notification_bar(
            text='You can close me!',
            name="my_bar",
            # Get notified when the notification bar is dismissed.
            events=['dismissed'],
        )
    if q.args.show_side_panel:
        q.page["meta"].side_panel = ui.side_panel(
        title="",
        items=[ui.text("Side panel text"), ui.button(name='show_notification_bar', label='Show notification bar')],
        name="side_panel",
        events=["dismissed"],
        closable=True,
    )

    # Did we get events from the notification bar?
    if q.events.my_bar:
        # Was the notification bar dismissed?
        print("events triggered")
        if q.events.my_bar.dismissed:
            # Delete the notification bar
            print("dismissed")
            q.page['meta'].notification_bar = None

    if q.events.side_panel:
        if q.events.side_panel.dismissed:
            print("side panel dismissed")
            q.page["meta"].side_panel = None

    await q.page.save()
```

marek-mihok:
https://user-images.githubusercontent.com/23740173/219204873-44217148-b199-474a-baaa-30ef4fa9aa96.mov

https://user-images.githubusercontent.com/23740173/219204859-060bebcb-dd90-4e51-85f3-a13c2dafadfc.mov

Closes #1833
Closes #1847

### DIFF
--- a/ui/src/time_picker.tsx
+++ b/ui/src/time_picker.tsx
@@ -210,47 +210,49 @@ export const
         {theme && AdapterDateFns && format
           ? <ThemeProvider theme={theme}>
             <LocalizationProvider dateAdapter={AdapterDateFns}>
-              <TimePicker
-                value={value}
-                label={label}
-                open={isDialogOpen}
-                onChange={value => setValue(value as Date)}
-                onAccept={value => onSelectTime(value as Date)}
-                onClose={() => setIsDialogOpen(false)}
-                ampm={hour_format === '12'}
-                showToolbar
-                ToolbarComponent={({ parsedValue, setOpenView, ampm }) =>
-                  <Toolbar
-                    setOpenView={setOpenView}
-                    time={parsedValue ? formatDateToTimeString(parsedValue as D, ampm ? '12' : '24') : parsedValue as null}
-                    label={label}
-                    switchAmPm={switchAmPm}
-                  />
-                }
-                PopperProps={{ ref: popperRef, anchorEl: () => textInputRef.current as VirtualElement, onBlur, ...popoverProps }}
-                minTime={min ? parseTimeStringToDate(min) : undefined}
-                maxTime={max ? parseTimeStringToDate(max) : undefined}
-                minutesStep={allowedMinutesSteps[minutes_step]}
-                disabled={disabled}
-                onOpen={onOpen}
-                renderInput={({ inputProps, error }: TextFieldProps) =>
-                  <div ref={textInputRef} data-test={m.name}>
-                    <Fluent.TextField
-                      iconProps={{ iconName: 'Clock' }}
-                      onClick={() => setIsDialogOpen(true)}
-                      onChange={inputProps?.onChange}
-                      placeholder={placeholder}
-                      disabled={disabled}
-                      readOnly
-                      value={value ? formatDateToTimeString(value, hour_format) : ''}
+              <Fluent.FocusTrapZone isClickableOutsideFocusTrap>
+                <TimePicker
+                  value={value}
+                  label={label}
+                  open={isDialogOpen}
+                  onChange={value => setValue(value as Date)}
+                  onAccept={value => onSelectTime(value as Date)}
+                  onClose={() => setIsDialogOpen(false)}
+                  ampm={hour_format === '12'}
+                  showToolbar
+                  ToolbarComponent={({ parsedValue, setOpenView, ampm }) =>
+                    <Toolbar
+                      setOpenView={setOpenView}
+                      time={parsedValue ? formatDateToTimeString(parsedValue as D, ampm ? '12' : '24') : parsedValue as null}
                       label={label}
-                      required={required}
-                      styles={{ field: { cursor: 'pointer', height: 32 }, icon: { bottom: 7 } }}
-                      errorMessage={error ? getErrMsg(hour_format, min, max) : undefined}
+                      switchAmPm={switchAmPm}
                     />
-                  </div>
-                }
-              />
+                  }
+                  PopperProps={{ ref: popperRef, anchorEl: () => textInputRef.current as VirtualElement, onBlur, ...popoverProps }}
+                  minTime={min ? parseTimeStringToDate(min) : undefined}
+                  maxTime={max ? parseTimeStringToDate(max) : undefined}
+                  minutesStep={allowedMinutesSteps[minutes_step]}
+                  disabled={disabled}
+                  onOpen={onOpen}
+                  renderInput={({ inputProps, error }: TextFieldProps) =>
+                    <div ref={textInputRef} data-test={m.name}>
+                      <Fluent.TextField
+                        iconProps={{ iconName: 'Clock' }}
+                        onClick={() => setIsDialogOpen(true)}
+                        onChange={inputProps?.onChange}
+                        placeholder={placeholder}
+                        disabled={disabled}
+                        readOnly
+                        value={value ? formatDateToTimeString(value, hour_format) : ''}
+                        label={label}
+                        required={required}
+                        styles={{ field: { cursor: 'pointer', height: 32 }, icon: { bottom: 7 } }}
+                        errorMessage={error ? getErrMsg(hour_format, min, max) : undefined}
+                      />
+                    </div>
+                  }
+                />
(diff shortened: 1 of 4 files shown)


ui/src/time_picker.tsx
```diff
@@ -79,8 +79,11 @@ const
     toolbarLabel: { maxWidth: '70%' }
   }),
   popoverProps: Partial<PopperProps> | undefined = {
+    container: () => document.querySelector(".ms-Panel") || document.body,
```
mturoci, line 82: This solution is fine, but I am wondering whether we couldn't use HTML parent of `XTimePicker` component for all cases.

ui/src/time_picker.tsx
```diff
@@ -79,8 +79,11 @@ const
...
     placement: 'bottom-start',
+    style: { pointerEvents: 'auto' },
```
mturoci, line 84: What is this for?

ui/src/time_picker.tsx
```diff
@@ -152,6 +156,11 @@ export const
       },
       // HACK: https://stackoverflow.com/questions/70106353/material-ui-date-time-picker-safari-browser-issue
       onOpen = () => setTimeout(() => (document.activeElement as HTMLElement)?.blur()),
+      onBlur = (ev: React.FocusEvent) => {
+        if (!ev.relatedTarget || !Fluent.elementContains(popperRef.current, ev.relatedTarget as HTMLElement)) {
```
mturoci, line 160: Why don't we want to always close the timepicker popper on blur?

mturoci (changes requested):
Thanks, @marek-mihok. The PR looks fine overall, but I wonder why doesn't stopping the propagation of `onMouseDown` event work. Maybe I am missing something?

ui/src/time_picker.tsx
```diff
@@ -79,8 +79,11 @@ const
     toolbarLabel: { maxWidth: '70%' }
   }),
   popoverProps: Partial<PopperProps> | undefined = {
+    container: () => document.querySelector(".ms-Panel") || document.body,
     placement: 'bottom-start',
+    style: { pointerEvents: 'auto' },
     sx: {
+      zIndex: 9999998, // Needs to be higher than Fluent UI Panel, but lower than ui.notification_bar.
```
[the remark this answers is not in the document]
marek-mihok: It's because Fluent uses `z-index: 1000000`. In this case we would need to override all Fluent z-indexes where applicable.

ui/src/time_picker.tsx
marek-mihok: We could try this.

ui/src/time_picker.tsx
marek-mihok: If the container element is Panel element which uses `pointer-events: 'none'` we need to override this to make the time picker popup responsive to clicks. Let's see if we could use `XTimePicker` as a container.

ui/src/time_picker.tsx
marek-mihok: Because `onBlur` is also called when clicking on the popup buttons or when clicking on the dev console etc.

ui/src/time_picker.tsx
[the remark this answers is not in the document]
mturoci: What does "all" mean in this context? I thought it was just 2 places - side panel and dialog.

ui/src/time_picker.tsx
mturoci: Go ahead and let me know, but first try to understand why `onMouseDown` prevention is not enough since it's suspicious.

ui/src/time_picker.tsx
mturoci: Got it, makes sense.

ui/src/time_picker.tsx
mturoci: Just a thought: maybe the pointerEvents from the next comment prevented the `onMouseDown` from firing.

ui/src/time_picker.tsx
mturoci: > onBlur is also called when clicking on the popup buttons

Good to know. Thanks!

mturoci:
Update:

While I was not able to figure out how to stop the event propagation properly yet, I might have found a better way of resolving this.

As #1837 implies, this problem appears on every "overlay-like" component, not just timepicker. So instead of fixing every single component and risk forgetting this when building a new one, we should probably just check, whether the `event.target` element from side panel's `onDismiss` contains `.ms-overlay` class, granted the elements that are supposed to be rendered "above" sidepanel have the proper `z-index` set.

Something like

```tsx
    onDismiss = (e?: React.SyntheticEvent<HTMLElement, Event> | KeyboardEvent) => {
      if (e && e.target && !(e.target as HTMLElement).classList.contains('ms-Overlay')) return
```

wdyt @marek-mihok ?

The stop propagation question is still a mystery though. Tested both [React 18](https://codesandbox.io/s/musing-grothendieck-48qjsy) and [React 17](https://codesandbox.io/s/react-17-forked-9p8o0p) and both stop the propagation properly, since React doesn't attach event listeners to `document.body` since version 17.

marek-mihok:
@mturoci seems like a solution to me! Let's count on the `ms-Overlay` class together with proper `z-index`es.

Btw, did you really meant to tag #1837 ?

mturoci:
> Btw, did you really meant to tag https://github.com/h2oai/wave/issues/1837 ?

Oops, should be 1847 :)

mturoci (approved):
Thanks @marek-mihok! Let's merge this as is and investigate whether `onBlur` is really needed in the next PR.

PR #1849, opened 2023-02-22
Fixed card->box assignments so that lhs_hello renders to the lhs box …

random-developer:
…as intended.


mturoci (approved):
Thanks @random-developer!

PR #1850, opened 2023-02-23
chore: Img annotator feedback.

mturoci:
This PR incorporates the image annotator feedback.

Features implemented:

* Shift-click now allows for deselection as well.
* If multiple shapes selected, all are dragged together.
* Escape cancels rect drawing as well.
* Initial active tool is picked based on whether items are provided or not.

Bugs fixed:

* When multiple shapes are selected, the corner resize resized random one.
* Multiple copy paste copy paste resulted in more copies than expected (2**).
* Tooltip listed `d` for deletion - changed it to backspace/delete.
* When pasting multiple times, only 1 shape could be moved out, the rest would be stuck with it.
* When copy&pasting polygons and then moving the pasted polygon, the polygon count increased by factor of 2.

Changed:
* Sorted shortcuts info tooltip by importance.
* Backspace or delete can be used for shape deletion.
* Zoom no longer requires ctrl to be pressed.

Part of this PR dealt with an extensive refactor that should also improve rendering performance, especially for complex shapes.

@marek-mihok can you please do a QA and see if I didn't break anything?


marek-mihok:
@mturoci, here is the list of what I've been able to find out:

- **Shapes can be moved by mouse when cursor is outside of the shape**

https://user-images.githubusercontent.com/23740173/221585565-4fcf050f-c379-4eff-8a70-fc5d7ef62b2f.mov

Steps to reproduce:

1. Select any shape
2. Move the cursor out of the shape
3. Mouse down and drag

- **Image is shaking when being dragged while zoomed in**

https://user-images.githubusercontent.com/23740173/221590500-d997c451-d8dc-4d0f-8885-81d0052363d6.mov

Steps to reproduce:

1. Zoom in by mouse wheel
2. Mouse down and drag

- **Error occurs when trying to delete last polygon vertex (`backspace`). Also the preview line does not disappear until mouse is moved**

https://user-images.githubusercontent.com/23740173/221593675-878062bd-3060-417d-aed1-e9679ac29bba.mov

Steps to reproduce:

1. Click on the image (to focus the image annotator)
2. Press `p` to select polygon tool
3. Press `backspace` to try to delete not existing polygon vertex
(error occurs)
4. Draw one polygon line
5. Press `backspace` twice without moving the mouse
(error occurs again and the preview line stays)

- **It is hard to select the shape sometimes** 

If you want to select the shape you need to make sure the cursor is not moving at all. It would be better to select the shape on click even if the cursor is being moved.

https://user-images.githubusercontent.com/23740173/221597780-fbfec1ed-0e66-474c-a87d-a1b029da6c63.mov

marek-mihok:
@mturoci, round two:

- **Resizing is not performed live**

https://user-images.githubusercontent.com/23740173/221600856-37d84b9f-d80a-4d99-91f0-55ddaa69031a.mov


- **Moving is not canceled when cursor goes behind the canvas boundary**

https://user-images.githubusercontent.com/23740173/221601369-b7680d0b-8b3b-4955-8ce9-dbd2dfec4590.mov

Steps to reproduce:

1. Select the shape
2. Mouse down and drag
3. Drag the mouse behind the image/canvas boundary
4. Release the mouse button (mouse up)
5. Move cursor back to the image/canvas
(the shape is being moved)

Please note, all of the issues above were found on Chrome `Version 110.0.5481.100 (Official Build) (arm64)`.

Also it would be good to consider if we shouldn't use a `image_annotator.py` example image from different source.

Screenshot 2023-02-27 at 16 16 57

I am getting a 403 forbidden error when trying to load the image for both Firefox and Safari.

It's probably due to pexels.com human identity check, because when I open the image url in the browser I am getting this:

Screenshot 2023-02-27 at 16 17 21

marek-mihok:
@mturoci, I confirm the issues are the same across all browsers. Also I was not able to find any new issues.

mturoci:
@marek-mihok all the issues fixed except of "It is hard to select the shape sometimes" since IIUC, you drag over the shape instead of clicking it properly. Correct me if I am wrong though. Let me know if the rest of the fixes work properly.

> Also it would be good to consider if we shouldn't use the example image for image_annotator.py from different source.

Sure. Feel free to find one we could use. Just make sure the licensing checks out.

marek-mihok:
@mturoci, that was fast! 

Everything from the issues above is looking good, just a small thing regarding **Moving is not canceled when cursor goes behind the canvas boundary** - the cursor is not updating util clicked when returning to canvas.

https://user-images.githubusercontent.com/23740173/221798537-471d2809-289f-42ae-9d54-f2dfb015a4e9.mov

      
<br><br>

While checking for old ones I was able to find couple of new (smaller) ones:      


- **Wrong cursor when resizing rectangle flipped on vertical/horizontal axis**

https://user-images.githubusercontent.com/23740173/221802250-d2cff8ee-b8db-44eb-8d93-13e04049b448.mov

- **Double-click selects the image (Chrome only)**

E.g. when you click on the shape to select it and then you click on it the second time to move it.

https://user-images.githubusercontent.com/23740173/221801676-828763fc-2638-4b07-9de1-5d70fbecb633.mov

- **Not able to drag the image when needs to drag over the large shape**

This is the main reason why we originally implemented `CTRL/Command + Drag` shortcut. The only "workaround" now is to zoom out and to zoom back in.

https://user-images.githubusercontent.com/23740173/221800116-3739dd11-7800-4b5b-8203-7268948cf5ad.mov


Regarding the **It is hard to select the shape sometimes** point, you are correct that I drag over the shape. On my screen recording I am performing an exaggerated drag moves, but usually almost every time when you click on something you are moving the cursor at least a little bit. That's why I find it hard to select the shape.

> Sure. Feel free to find one we could use. Just make sure the licensing checks out.

I was not able to reproduce the issue since then, but I'll check for some other license-free source to make sure the example will work every time for everyone.

mturoci:
Cursor bugs fixed, img selection prevented as well.

> Not able to drag the image when needs to drag over the large shape

The workaround is fine. Let's not diminish UX (ctr) for the whole zoom just to avoid a single edge case.

> usually almost every time when you click on something you are moving the cursor at least a little bit

Agree, but is this a regression though? IIRC this behavior exists in 0.24.2 as well in which case you can just file a GH issue and we can deal with it later.

marek-mihok:
@mturoci, it seems we are coming to the end. I am not sure if it was there before, but I am **not able to draw a rectangle when zoomed in**. I think we can fix this by allowing dragging only in the 'select' mode. By adding the simple `activeShape === 'select'` condition [here to line 406](https://github.com/h2oai/wave/blob/f989fe1923729edcacab85913c22e1bb214df23c/ui/src/image_annotator.tsx#L406): 
```ts
if (zoom > 1 && clickStartPosition?.dragging && activeShape === 'select' && !(rectRef.current?.isMovedOrResized() || polygonRef.current?.isMovedOrResized())) {
```

https://user-images.githubusercontent.com/23740173/221872949-6c8e2460-0897-45f5-931d-04e29cc7df09.mov

<br>

> Let's not diminish UX (ctr) for the whole zoom just to avoid a single edge case.

Isn't it possible to avoid the edge case without using CTRL? Because the issue here is only for polygons so maybe we could use here the same behaviour as for rectangles:

https://user-images.githubusercontent.com/23740173/221877219-62b33e15-7ce2-4420-aa8c-9be87e5c2eee.mov


<br>

FYI, I've pushed a [commit](https://github.com/h2oai/wave/pull/1850/commits/f989fe1923729edcacab85913c22e1bb214df23c) with an updated image to prevent the 403 forbidden error I was getting from pexels.

marek-mihok:
LGTM @mturoci !

PR #1856, opened 2023-02-27
chore: Account for absolute data_dir paths during file uploads.

mturoci:
Make fast file uploads work in cloud env as well.


mturoci:
Tested against cloud, works as expected.

PR #1861, opened 2023-03-01
fix: Prevent table reset on items change for paginated table

marek-mihok:
Fixes the regression caused by #1724 when paginated table became frozen after the page change.

Also fixes API description typo for `single` prop and updates the API description for `groupable` to align with the [current implementation](https://github.com/h2oai/wave/commit/294e642b0a0d8071091dc4b4914a05a89b0ed50b).


mturoci (dismissed):
Thanks @marek-mihok!

PR #1864, opened 2023-03-02
fix: Unable to open time picker

marek-mihok:
Fixes the regression of time picker where user was unable to properly open the time picker in case multiple time pickers are present on the page, e.g. in `time_picker.py` example.


mturoci (approved):
Thanks @marek-mihok. Works as intended!

Issue #1493, opened 2022-06-09
Semantic validation for ui.components


pascal-pfeiffer:
#### Is your feature request related to a problem? Please describe

For some ui components such as `ui.notification_bar` you can specify `events`. This is only working if also a `name` property is given, which is optional otherwise. If a name is not provided, the events are not catched and the root cause is unclear (fails silently).

#### Describe the solution you'd like

Warning the user about a missing parameter or mutually exclusive parameters would speed up debugging and development. This could be done by some kind of semantic validation, either at component input or in the IDE tools.

#### Describe alternatives you've considered

Check in the ui.component 
Check through the IDE plugins

PR #1870, opened 2023-03-06
feat: Allow by-name component access. #148

### LINKED ISSUE #148: Allow by-name access to page elements
API ergonomics improvement:

Currently supported: `card.items[0].caption = 'foo'`
Desirable: `card.items.upload_progress.caption = 'foo'` or `card.items['upload_progress'].caption = 'foo'`

Requires modification to server-side interpreter.

mturoci:
Allows for:

```py
from h2o_wave import Q, ui, main, app


@app('/')
async def serve(q: Q):
    q.page['wizard'] = ui.form_card(box='1 1 2 4', items=[
        ui.text_xl(name='text_name', content='Wizard'),
        ui.textbox(name='textbox_name', label='Textbox'),
        ui.inline(items=[
            ui.button(name='back', label='Back'),
        ]),
    ])
    # Instead of q.page['wizard'].items[0].text_xl.content = 'foo'
    q.page['wizard'].text_name.content = 'foo'
    q.page['wizard'].textbox_name.label = 'foo'
    q.page['wizard'].back.label = 'foo'

    q.page['header'] = ui.header_card(box='4 6 4 1', title='Header', subtitle='Subtitle', secondary_items=[
        ui.button(name='button_name', label='Button'),
    ])
    q.page['header'].button_name.label = 'foo'

    q.page['example'] = ui.form_card(box='5 1 4 5', items=[
        ui.buttons([
            ui.button(name='primary_button', label='Primary', primary=True),
        ]),
    ])
    q.page['example'].primary_button.label = 'foo'

    await q.page.save()

```

## Needs discussion

> issue description: Requires modification to server-side interpreter.

@lo5 can you elaborate what you meant here? AFAIU, the whole UI state is stored client-side meaning, there is not much to be done at the server-side interpreter unless I am missing something. Apart from [guarding the accessors](https://github.com/h2oai/wave/blob/82868c1157df4e679d6052cc189c6dc643ec03f8/py/h2o_wave/h2o_wave/core.py#L289).

> page[, 'upload_progress'].caption = 'foo' # Scans all cards on page for upload_progress
page[, 'stats_table', 'search_bar'].caption = 'foo' # Scans all cards on page for search_bar inside stat_table

Wondering whether these would be useful - cc @Far0n @pascal-pfeiffer @vopani @srini-x for feedback.

Closes #148

### DIFF
--- a/card.go
+++ b/card.go
@@ -66,7 +66,7 @@ func loadCard(ns *Namespace, c CardD) *Card {
 		}
 		if k == "items" || k == "secondary_items" || k == "buttons" {
 			card.nameComponentMap = make(map[string]interface{})
-			fillComponentNameMap(card.nameComponentMap, v)
+			fillNameComponentMap(card.nameComponentMap, v)
 		}
 		ks[0] = k
 		card.set(ks, v)
@@ -98,7 +98,7 @@ func (c *Card) set(ks []string, v interface{}) {
 		if v, ok := c.nameComponentMap[ks[0]]; ok && len(ks) == 2 {
 			x = v
 		} else {
-			// DEPRECATED: access page.items[idx].wrapper.prop.
+			// DEPRECATED: Access via page.items[idx].wrapper.prop.
 			for _, k := range ks[:len(ks)-1] {
 				x = get(x, k)
 			}
@@ -184,18 +184,22 @@ func deepClone(ix interface{}) interface{} {
 	return ix
 }
 
-func fillComponentNameMap(m map[string]interface{}, items interface{}) {
+func fillNameComponentMap(m map[string]interface{}, items interface{}) {
 	for _, v := range items.([]interface{}) {
-		// This map always has a single key - wrapper.
+		// This map always has a single key - wrapper so this is O(1) not O(n).
 		for _, v := range v.(map[string]interface{}) {
-			for attributeKey, attributeValue := range v.(map[string]interface{}) {
-				if attributeKey == "name" {
-					m[attributeValue.(string)] = v
-					continue
-				}
-				if attributeKey == "items" || attributeKey == "secondary_items" || attributeKey == "buttons" {
-					fillComponentNameMap(m, attributeValue)
-				}
+			component := v.(map[string]interface{})
+			if name, ok := component["name"]; ok {
+				m[name.(string)] = v
+			}
+			if items, ok := component["items"]; ok {
+				fillNameComponentMap(m, items)
+			}
+			if secondaryItems, ok := component["secondary_items"]; ok {
+				fillNameComponentMap(m, secondaryItems)
+			}
+			if buttons, ok := component["buttons"]; ok {
+				fillNameComponentMap(m, buttons)
 			}
 		}
 	}
--- a/card.go
+++ b/card.go
@@ -79,9 +79,9 @@ func loadCard(ns *Namespace, c CardD) *Card {
 					}
 				}
 			}
-			ks[0] = k
-			card.set(ks, v)
 		}
+		ks[0] = k
+		card.set(ks, v)
 	}
 	return card
 }
--- a/card.go
+++ b/card.go
@@ -46,13 +46,14 @@ func loadBuf(ns *Namespace, b BufD) Buf {
 
 // Card represents an item on a Page, and holds attributes and data for rendering views.
 type Card struct {
-	data map[string]interface{}
+	data             map[string]interface{}
+	nameComponentMap map[string]interface{} // Cache for cards with items, secondary_items or buttons.
 }
 
 const dataPrefix = "~"
 
 func loadCard(ns *Namespace, c CardD) *Card {
-	card := &Card{make(map[string]interface{})}
+	card := &Card{make(map[string]interface{}), nil}
 	ks := make([]string, 1) // to avoid allocation during card.set() below
 	for k, v := range c.D {
 		if len(k) > 0 && strings.HasPrefix(k, dataPrefix) {
@@ -63,8 +64,24 @@ func loadCard(ns *Namespace, c CardD) *Card {
 				}
 			}
 		}
-		ks[0] = k
-		card.set(ks, v)
+		// TODO: Add support for secondary_items and buttons.
+		if k == "items" {
+			card.nameComponentMap = make(map[string]interface{})
+			for _, v := range v.([]interface{}) {
+				// This map always has a single key - wrapper.
+				for _, v := range v.(map[string]interface{}) {
+					// TODO: Add support for nested items, e.g. ui.inline.
+					for attributeKey, attributeValue := range v.(map[string]interface{}) {
+						if attributeKey == "name" {
+							card.nameComponentMap[attributeValue.(string)] = v
+							break
+						}
+					}
+				}
+			}
+			ks[0] = k
+			card.set(ks, v)
+		}
 	}
 	return card
 }
@@ -88,10 +105,17 @@ func (c *Card) set(ks []string, v interface{}) {
 		}
 	default: // .foo.bar.baz = qux
 		var x interface{} = c.data
-		p := ks[len(ks)-1]
-		for _, k := range ks[:len(ks)-1] {
-			x = get(x, k)
+
+		// By-name access.
+		if v, ok := c.nameComponentMap[ks[0]]; ok && len(ks) == 2 {
+			x = v
+		} else {
+			for _, k := range ks[:len(ks)-1] {
+				x = get(x, k)
+			}
 		}
+
+		p := ks[len(ks)-1]
 		set(x, p, v)
 	}
 }
--- a/card.go
+++ b/card.go
@@ -24,9 +24,9 @@ type Buf interface {
 	// get cursor at key k
 	get(k string) (Cur, bool)
 	// overwrite all records
-	put(v interface{})
+	put(v any)
 	// set record at key k
-	set(k string, v interface{})
+	set(k string, v any)
 	// dump contents
 	dump() BufD
 }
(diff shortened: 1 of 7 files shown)


Far0n:
> Wondering whether these would be useful

definitely yes imho

mturoci:
>definitely yes imho

Can you expand with some examples?

Far0n:
> Can you expand with some examples?

I have no concrete example at hand, but I found these `.items[x]` accesses alsways a bit wonky and "hard to read".

mturoci:
>I found these .items[x] accesses always a bit wonky and "hard to read" and error prone (due to only getting Ref type information)

Oh, I should have made the question more clear then. Let me rephrase. The problem you just described is already implemented in this PR: `q.page['card_name'].<component_name>.<prop_to_mutate>`. I am more interested in the proposals:

* page[, 'upload_progress'].caption = 'foo' # Scans all cards on page for upload_progress - When would I want to update all the components with the same name?
* page[, 'stats_table', 'search_bar'].caption = 'foo' # Scans all cards on page for search_bar inside stat_table - do I need a hierarchy, when I can access the components directly by name?

Far0n:
> he problem you just described is already implemented in this PR

ah lol sry my bad.

```
    page[, 'upload_progress'].caption = 'foo' # Scans all cards on page for upload_progress - When would I want to update all the components with the same name?
    page[, 'stats_table', 'search_bar'].caption = 'foo' # Scans all cards on page for search_bar inside stat_table - do I need a hierarchy, when I can access the components directly by name?
```

That's nothing I find that useful as of now (no use case comes to mind).

Far0n:
I mean there are certainly use-cases for some sort of batch processing of components like enabling/disabling or resetting, but at least personally I didn't miss that feature yet.

lo5:
> can you elaborate what you meant here? 

Given two scripts S1 and S2, say:
- S1 creates the card and sets `card.items[0].caption = 'foo'` 
- S2 sets `card.items[0].caption = 'bar'`
 
The final caption will be `bar`.

Instead, say there was a way to name the item `upload_progress`:
- S1 creates the card and sets `card.upload_progress.caption = 'foo'`
- S2 sets `card.upload_progress.caption = 'bar'`

For this to work, the server needs to be aware that `upload_progress` is the name of a card, can iterate through the cards, and patch the correct card. Currently the patch algo can handle only slice-access by integer index: https://github.com/h2oai/wave/blob/master/card.go#L111-L116

lo5:
More thoughts: this is tricky to do primarily because `card.go` has no awareness of `items` or even what attributes a card has. One way to do this brute-force is to scan nested attributes for a `name` when the `map` case fails, but that can be expensive in the worst case, and I don't know if `.name` is guaranteed to be a component name universally.

The other option is to merge your PR, but with the caveat that by-name access is not supported for scripts, or for the uni-/multicast apps.

pascal-pfeiffer:
Thanks for implementing this @mturoci . This will be super handy when updating single items on a page. 

I can't say anything about the technical implementation, but I am wondering if it works on any hierarchy level of the component? 

Does any process enforce name uniqueness? If not, how would that be resolved? Would that go into the direction of your second question? 
>  page[, 'upload_progress'].caption = 'foo' # Scans all cards on page for upload_progress - When would I want to update all the components with the same name?

mturoci:
> The other option is to merge your PR, but with the caveat that by-name access is not supported for scripts, or for the uni-/multicast apps.

I would like to avoid this if possible. Thanks for the pointers, will see what I can do.

> wondering if it works on any hierarchy level of the component?

It should. If you check my example above, I target a button what is within items > inline > button.

> Does any process enforce name uniqueness? If not, how would that be resolved? Would that go into the direction of your second question?

There is no enforcement from Wave side currently (https://github.com/h2oai/wave/issues/1493). The name uniqueness needs to be preserved only within the "card" scope. This means that `q.page['card1'].items` can have the same named items as `q.page['card2']`. However, if 2 or more components with the same name are present within the same card, only the first one in order gets updated currently.

Generally speaking, I think we should enforce unique names per card as it makes sense for getting the outputs via `q.args` and also React depends on this. Currently, it's more of an "unwritten rule".

mturoci:
Seems like I managed to find the optimal solution to support scripts, broadcast, multicast as well without any serious perf degradation. In fact, it should run a bit faster now.

@geomodular could you please review the Go code to make sure it's idiomatic?

The parts I am not so sure about:

* Unsafe type casts.
* Is `fillNameComponentMap` function okayish since it mutates passed hashmap, thus is not a pure function.
* I believe `map[string]interface{}` only stores pointers to the actual objects rather than values. Is this correct?

card.go
```diff
@@ -46,13 +46,14 @@ func loadBuf(ns *Namespace, b BufD) Buf {
...
-	data map[string]interface{}
+	data             map[string]interface{}
+	nameComponentMap map[string]interface{} // Cache for cards with items, secondary_items or buttons.
 }
 
 const dataPrefix = "~"
 
 func loadCard(ns *Namespace, c CardD) *Card {
-	card := &Card{make(map[string]interface{})}
+	card := &Card{make(map[string]interface{}), nil}
```
geomodular, line 56: I'd make this initialization more verbose because it's extended now. But it's more important to follow the code style of the project:
```go
card := &Card{
  data: make(map[string]interface{}),
  nameComponentMap: nil,
}
```

card.go
```diff
@@ -170,3 +185,24 @@ func deepClone(ix interface{}) interface{} {
 	}
 	return ix
 }
+
+func fillNameComponentMap(m map[string]interface{}, items interface{}) {
+	for _, v := range items.([]interface{}) {
+		// This map always has a single key - wrapper so this is O(1) not O(n).
+		for _, v := range v.(map[string]interface{}) {
```
geomodular, line 192: I'd fix the variable clashing here. I think it's simpler to reason about the code without clashing. You can do e.g.:

```go
for _, item := range items.([]interface{}) {
    // This map always has a single key - wrapper so this is O(1) not O(n).
    for _, v := range item.(map[string]interface{}) {
```

You can use `i` as well but it's more idiomatic to use `i` for index. Or you can rename `items interface{}` to `entries interface{}` and use `e` as a pointer.

card.go
```diff
@@ -170,3 +185,24 @@ func deepClone(ix interface{}) interface{} {
...
+			component := v.(map[string]interface{})
+			if name, ok := component["name"]; ok {
+				m[name.(string)] = v
```
geomodular, line 195: To be safe here, I'd do a `type switch` or just check the correctness of type with `if`:
```go
if n, ok := name.(string); ok {
    m[n] = v
}
```

card.go
```diff
@@ -170,3 +185,24 @@ func deepClone(ix interface{}) interface{} {
 	}
 	return ix
 }
+
+func fillNameComponentMap(m map[string]interface{}, items interface{}) {
+	for _, v := range items.([]interface{}) {
```
geomodular, line 190: Is it always expected to get an array? This will panic if not. It would be safer to expect `items []interface{}` but I assume the variable passed inside (`c.D`) is of type `interface{}`.

geomodular (approved):
Good job, Martin. Overall I think it's fine. Take my comments as suggestions.

> Unsafe type casts.

This sometimes emerges. You have `interface{}`; you need to cast. If by `unsafe` you mean without the type checking, then I'd say it's better to check. But it depends on the context.

> Is fillNameComponentMap function okayish since it mutates passed hashmap, thus is not a pure function.

That's ok. We are not even trying to do functional programming in `go`. I'd say the function is readable, and that's the goal.

> I believe map[string]interface{} only stores pointers to the actual objects rather than values. Is this correct?

Yes, that's correct. `map[string]interface{}` is essentially a dictionary of pointers.

card.go
mturoci: Seems like there is no strict style regarding the Struct initialization. I was able to find examples for both: named and anonymous params. Will go with your suggestion.

card.go
mturoci: Fair point. I am hesitant to go with `entries`, since we iterate over `items` on the Python side so it's easier to reason about when the naming matches. Instead, will go with:

```go
	for _, wrappedItem := range wrappedItems.([]any) {
		// This map always has a single key - wrapper so this is O(1) not O(n).
		for _, item := range wrappedItem.(map[string]any) {
			component := item.(map[string]any)
```

card.go
mturoci: Will do.

card.go
mturoci: >Is it always expected to get an array? This will panic if not

Yes, we are only interested in Wave component arrays - `items | secondary_items | buttons` which are all supposed to be of type `Component[]`. If any of them is not, the panic is desired as it hints that something is off with the typing.

I could do an if check for type casting and throw a nicer err than the panic one, but not sure if worth it.

>It would be safer to expect items []interface{} but I assume the variable passed inside (c.D) is of type interface{}

Correct. The casting would need to happen either in outer function or here.

mturoci:
Thanks for the review @geomodular!

> If by unsafe you mean without the type checking, then I'd say it's better to check

Yes, that's exactly what I meant, especially the `fillNameComponentMap` part which has a lot of casts. The thing is though that all the types are expected and an err should be thrown if any of them doesn't match. Custom err is more readable than panic ofc, just not sure if worth it.

All the comments were addressed. Updated the go version via `go mod edit -go 1.19 ` and `go mod tidy`.

card.go
geomodular: > I could do an if check for type casting and throw a nicer err than the panic one, but not sure if worth it.

I'd probably skip it. But maybe a comment would be nice.

geomodular (dismissed):
Marvelous!

vopani:
This is a killer feature, can't wait to try it out.
Thanks for working on it! 🙏

mturoci:
Available in nightly, feedback / bug reports encouraged since this spans all the components throughout Wave.

> This is a killer feature, can't wait to try it out.

Glad to hear that!

PR #1890, opened 2023-03-14
fix: Update wave args of ui.time_picker on every time change #1889

### LINKED ISSUE #1889: Visuals of time_picker selection don't always match passed argument
### Discussed in https://github.com/h2oai/wave/discussions/1888

<sup>Originally posted by **aranvir** March 14, 2023</sup>
Hi! First of all, I really enjoy working with this framework and exploring this.

I found an issue that can be best reproduced with the time_picker https://wave.h2o.ai/docs/examples/time-picker

When selecting the required time, one needs to explicitly click on the the hour and the minutes to set a proper time. If, for example, you just click on 12 (which auto-completes to 12:00) and then click on the white space around, the value will be displayed BUT when you click Submit, the time passed is actually `None`.

So this is of course a bit counter-intuitive because you see the time you wanted to enter, but it is then not actually set. I assume the same will happen for other "multi-stage" pickers where implicit values are not considered a confirmation of the form input, but I haven't tested it.

Not sure, if this can be solved programmatically or if this is just "how forms work". If it's the latter then I think this is something that should be mentioned in the documentation so frontend noobs like me are not caught by surprise ;)</div>

via @aranvir

marek-mihok:
Updating of wave args if now moved from time picker's `onAccept` to `onChange` to make sure it is updated every time user changes the time, not only on popup close.

Closes #1889

### DIFF
--- a/ui/src/time_picker.tsx
+++ b/ui/src/time_picker.tsx
@@ -148,10 +148,11 @@ export const
           return date
         })
       },
-      onSelectTime = (time: D | null) => {
+      onChangeTime = (time: D | null) => {
         wave.args[m.name] = time ? formatDateToTimeString(time, '24') : null
-        if (m.trigger) wave.push()
+        setValue(time)
       },
+      onSelectTime = () => { if (m.trigger) wave.push() },
       // HACK: https://stackoverflow.com/questions/70106353/material-ui-date-time-picker-safari-browser-issue
       onOpen = () => setTimeout(() => (document.activeElement as HTMLElement)?.blur()),
       onBlur = (ev: React.FocusEvent) => {
@@ -212,8 +213,8 @@ export const
                 value={value}
                 label={label}
                 open={isDialogOpen}
-                onChange={value => setValue(value as Date)}
-                onAccept={value => onSelectTime(value as Date)}
+                onChange={value => onChangeTime(value as Date)}
+                onAccept={onSelectTime}
                 onClose={() => setIsDialogOpen(false)}
                 ampm={hour_format === '12'}
                 showToolbar
--- a/ui/src/time_picker.tsx
+++ b/ui/src/time_picker.tsx
@@ -145,12 +145,15 @@ export const
         setValue((prevValue) => {
           const date = new Date(prevValue!)
           date.setTime(date.getTime() + 12 * 60 * 60 * 1000)
+          wave.args[m.name] = formatDateToTimeString(date, '24')
           return date
         })
       },
-      onChangeTime = (time: D | null) => {
-        wave.args[m.name] = time ? formatDateToTimeString(time, '24') : null
-        setValue(time)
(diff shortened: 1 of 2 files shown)


ui/src/time_picker.tsx
```diff
@@ -212,8 +213,8 @@ export const
                 value={value}
                 label={label}
                 open={isDialogOpen}
-                onChange={value => setValue(value as Date)}
-                onAccept={value => onSelectTime(value as Date)}
+                onChange={value => onChangeTime(value as Date)}
```
mturoci, line 216: Let's use proper types here.

```suggestion
                onChange={onChangeTime}
```

mturoci (changes requested):
Thanks @marek-mihok. The fix looks correct. Please add unit tests as well.

ui/src/time_picker.tsx
marek-mihok: What do you mean by proper types? onChange gives us the value of type `unknown`, which is basically a Date object every time. Do you mean I should replace the param type in my `onChangeTime` handler with `unknown`? 

```ts
- onChangeTime = (time: D | null) => {
+ onChangeTime = (time: unknown) => {
```

mturoci: >onChange gives us the value of type unknown, which is basically a Date object every time

If it were true, it wouldn't be typed as unknown, but Date.

>Do you mean I should replace the param type in my onChangeTime handler with unknown?

The current `onChangeTime` function signature doesn't match the one in the lib. In addition to the value, there is also `keyboardInputValue` which I assume allows for selecting time with a keyboard as well so would be good to support that also.

marek-mihok: > I assume allows for selecting time with a keyboard as well so would be good to support that also

IIRC the keyboard input was supported during development of this component but in later phases we agreed on removing the support. I can't remember the specific reason, but I think there were a lot of edge cases to handle and the code was too robust. Also we wanted to make it consistent with a date picker where you cannot use keyboard input as well. You can pick only what the picker offers you to select and also you don't need to care about formats (should I write `04:00`, `4:00`, `4:00 am` or `4:00am`?).

However, If you want to support it, I can create a feature request and/or add there a TODO.

mturoci: Ah, I remember now, fair point. Let's leave it as is then and handle only the `unknown` part of my prev comment.

marek-mihok:
@mturoci regarding unit tests there is no easy way of triggering click events on the clock component. That's why we have only tests which exclude time change using the clock picker. That means having tests where `onChange` is called (where `wave.args` are set) can be problematic. The question is if we want to spend a long time efforts (as the last time) trying to create these tests without a guaranteed success.

mturoci (approved):
The test you added is enough as it fails without this fix. Good job!

PR #1906, opened 2023-03-27
feat: Dropdown button #1887

### LINKED ISSUE #1887: Dropdown button

A new UI component - a dropdown button similar to e.g. bootstrap https://getbootstrap.com/docs/4.0/components/dropdowns/

image

A dropdown button to group several related actions under a single UI element.

marek-mihok:
This change brings the `commands` attribute for _ui.button_ where you can provide other relevant actions for the button that are displayed in the context menu.

```ts
export interface Button {
  /** An identifying name for this component. If the name is prefixed with a '#', the button sets the location hash to the name when clicked. */
  name: Id
  /** The text displayed on the button. */
  label?: S
  /** The caption displayed below the label. */
  caption?: S
  /** A value for this button. If a value is set, it is used for the button's submitted instead of a boolean True. */
  value?: S
  /** True if the button should be rendered as the primary button in the set. */
  primary?: B
  /** True if the button should be disabled. */
  disabled?: B
  /** True if the button should be rendered as link text and not a standard button. */
  link?: B
  /** An optional icon to display next to the button label (not applicable for links). */
  icon?: S
  /** The width of the button, e.g. '100px'. */
  width?: S
  visible?: B
  tooltip?: S
  /** The path or URL to link to. If specified, the `name` is ignored. The URL is opened in a new browser window or tab. */
  path?: S
  /** When specified, a split button is rendered with extra actions tied to it within a context menu. Mutually exclusive with `link` attribute. */
  commands?: Command[]
}
```

Example:

https://user-images.githubusercontent.com/23740173/232728214-44ae100f-9b04-472b-8ca8-b2c3251725c3.mov

Closes #1887

### DIFF
--- a/ui/src/button.tsx
+++ b/ui/src/button.tsx
@@ -114,6 +114,13 @@ const
     center: 'center',
     between: 'space-between',
     around: 'space-around',
+  },
+  // Fixes the vertical scrollbar for menu items with icons.
+  subComponentStyles: Partial<Fluent.IContextualMenuSubComponentStyles> = {
+    menuItem: {
+      icon: { lineHeight: 'initial' },
+      subMenuIcon: { lineHeight: 'initial', height: 'auto' },
+    }
   }
 
 const
@@ -141,6 +148,7 @@ const
             ? { items: getItemProps(items), styles: { subComponentStyles } }
             : undefined
         })),
+      isIconOnly = !label && icon,
       // HACK: Our visibility logic in XComponents doesn't count with nested components, e.g. Butttons > Button.
       styles: Fluent.IButtonStyles = {
         root: {
@@ -152,14 +160,10 @@ const
           fontSize: 20,
           display: 'flex',
           alignItems: 'center'
-        }
-      },
-      // Fixes the vertical scrollbar for menu items with icons.
-      subComponentStyles: Partial<Fluent.IContextualMenuSubComponentStyles> = {
-        menuItem: {
-          icon: { lineHeight: 'initial' },
-          subMenuIcon: { lineHeight: 'initial', height: 'auto' }
-        }
+        },
+        splitButtonMenuButton: isIconOnly
+          ? { backgroundColor: cssVar('$card'), border: 'none' }
+          : undefined
       }
 
 
@@ -184,7 +188,7 @@ const
       } : undefined,
       split: !!commands
     }
-    if (!label && icon) return <Fluent.IconButton {...btnProps} data-test={name} title={caption} />
+    if (isIconOnly) return <Fluent.IconButton {...btnProps} data-test={name} title={caption} />
 
     return caption?.length
       ? primary
--- a/ui/src/button.tsx
+++ b/ui/src/button.tsx
@@ -21,7 +21,7 @@ import { cssVar, formItemWidth, padding } from './theme'
 import { XToolTip } from './tooltip'
 import { wave } from './ui'
 
-interface ButtonCommand {
+interface ButtonChoice {
   /** An identifying name for this component. If the name is prefixed with a '#', the command sets the location hash to the name when executed. */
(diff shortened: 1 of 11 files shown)


mturoci (changes requested):
Thanks @marek-mihok! Instead of introducing a new `ButtonChoice`, let's stick with what we already have for these types of scenarios: `ui.command`.

marek-mihok:
@mturoci originally I've considered it (please, check [this commit](https://github.com/h2oai/wave/pull/1906/commits/f74a169d6b75722000e39446b9150d8bee2e1644) out) but our existing `ui.command` does not contain the `path` property and contains a deprecated `data` prop. Are you sure with that?

And what about the `choices` prop? Should I rename it to `commands` or can I stick with `choices`?

website/widgets/form/button.md
```diff
@@ -126,3 +126,45 @@ q.page['form'] = ui.form_card(box='1 1 1 1', items=[
     ui.button(name='external_path_button', label='External', path='https://h2o.ai/')
 ])
 ```
+
+## With actions
+
+Using the `commands` attribute, you can provide other relevant actions for the button that are displayed in the context menu.
```
mturoci, line 132:
```suggestion
The `commands` attribute can provide other relevant actions for the button that are displayed in the context menu.
```

ui/src/button.tsx
```diff
@@ -125,6 +135,20 @@ const
           wave.push()
         }
       },
+      onClick = handleOnClick(name, value, path),
+      getItemProps: (commands: Command[]) => Fluent.IContextualMenuItem[] = commands =>
```
mturoci, line 139: Wondering whether we cannot reuse the existing [toCommands](https://github.com/h2oai/wave/blob/ced07d4a630d898da5f544e283e4d7890f0b3abe/ui/src/toolbar.tsx#L87) function that solves this already. It should already have styling (instead of your own `subComponentStyles`) and onClick figured out.

ui/src/button.tsx
```diff
@@ -64,6 +65,8 @@ export interface Button {
   tooltip?: S
   /** The path or URL to link to. If specified, the `name` is ignored. The URL is opened in a new browser window or tab. */
   path?: S
+  /** The menu with button actions. Ignored if `link` is True. */
```
mturoci, line 68:
```suggestion
  /** When specified, a split button is rendered with extra actions tied to it within a context menu. Mutually exclusive with `link` attribute. */
```

ui/src/button.tsx
```diff
@@ -125,6 +135,20 @@ const
...
         }
       },
+      onClick = handleOnClick(name, value, path),
+      getItemProps: (commands: Command[]) => Fluent.IContextualMenuItem[] = commands =>
+        commands.map(({ name, label, caption, icon, value, data, items }) => ({
+          key: name,
+          text: label,
+          title: caption,
+          iconProps: { iconName: icon },
+          data: value ?? data, disabled,
```
mturoci, line 145: is this `data` used anywhere?

mturoci (changes requested):
Thanks @marek-mihok. The spacing looks a bit odd compared to the rest of Wave contextual menus.

image
image

vs

image

ui/src/button.tsx
marek-mihok: Nope, we are using `value` inside onClick handler. Removing.

ui/src/button.tsx
marek-mihok: Nice to find that almost identical function as I've implemented already exists. It's reused and it works. Thanks!

mturoci (approved):
Looks good, just need to remove 1 line I missed before and we are good to go.

mturoci (approved):
LGTM.

PR #1911, opened 2023-03-30
Resolves Issue #1689: Display active table filter count

Stijn-Kuijper:
There are still a few things that need to be fixed, but I'm opening the pull request so I can explain what goes wrong better.


Stijn-Kuijper:
Closing this pull request, as the issue is more complex than initially thought and I will not be able to fix it.

PR #1925, opened 2023-04-12
fix: Display tags in wrapped cells correctly #1917

### LINKED ISSUE #1917: Table tags overlap and split across new line when ui.table_column cell_overflow is set to "wrap"
#### Wave SDK Version, OS
H2O Wave 0.25.2 20230317093151
Linux Rocky 8.7
Firefox Extended Support Release 102.8.0esr

#### Actual behavior
When creating a column with tags and setting the cell overflow to wrap, tags within a cell will overlap and split across lines
image

#### Expected behavior
I would expect tags to not overlap, and only split across lines if column width was less than the tag width

#### Steps To Reproduce
Run the following app:
```python
# Table / Tags
# Use tags in order to emphasize a specific value. For multiple tags in a single row use `,` as a delimiter.
# ---


_id = 0


class Issue:
    def __init__(self, text: str, tags: str):
        global _id
        _id += 1
        self.id = f'I{_id}'
        self.text = text
        self.tags = tags


# Create some issues
issues = []
issues.append(Issue("Issue 1", "DONE,SUCCESS,Feature"))
issues.append(Issue("Issue 2", "FAIL,Critical"))
issues.append(Issue("Issue 3", "DONE,SUCCESS,Critical, Feature"))
issues.append(Issue("Issue 4", "DONE,SUCCESS,Backlog"))
issues.append(Issue("Issue 5", "DONE,SUCCESS,Refactor,Backlog"))

columns = [
    ui.table_column(name='text', label='Issue', min_width='100px'),
    ui.table_column(name='tag', label='Badge', max_width='150px', cell_overflow='wrap', cell_type=ui.tag_table_cell_type(name='tags', tags=[
        ui.tag(label='FAIL', color='$red'),
        ui.tag(label='Critical', color= '$red'),
        ui.tag(label='DONE', color='#D2E3F8', label_color='#053975'),
        ui.tag(label='Feature', color= '$green'),
        ui.tag(label='Backlog', color= '$yellow'),
        ui.tag(label='Refactor', color= '$orange'),
        ui.tag(label='SUCCESS', color='$mint'),
    ])),
]


@app('/demo')
async def serve(q: Q):
    q.page['example'] = ui.form_card(box='1 1 -1 -1', items=[
        ui.table(
            name='issues',
            columns=columns,
            height='500px',
            width='300px',
            rows=[ui.table_row(name=issue.id, cells=[issue.text, issue.tags]) for issue in issues],
        )
    ])
    await q.page.save()

```

marek-mihok:
https://user-images.githubusercontent.com/23740173/232830137-41d48075-9e95-4ce1-bef9-d6fda0954244.mov

Closes #1917

### DIFF
--- a/ui/src/tag_table_cell_type.tsx
+++ b/ui/src/tag_table_cell_type.tsx
@@ -52,13 +52,11 @@ const css = stylesheet({
     display: 'inline-block',
     borderRadius: 4,
     padding: padding(4, 16),
-    $nest: {
-      '&:not(:last-child)': {
-        marginRight: 8
-      }
+    '&:not(:last-child)': {
+      marginRight: 8
     }
   },
-  multiline: { marginBottom: 8, }
+  multiline: { marginBottom: 8 }
 })
 
 export const XTagTableCellType = ({ model, serializedTags, isMultiline }: { model: TagTableCellType, serializedTags: S, isMultiline?: B }) => {
--- a/ui/src/tag_table_cell_type.tsx
+++ b/ui/src/tag_table_cell_type.tsx
@@ -52,14 +52,15 @@ const css = stylesheet({
     display: 'inline-block',
     borderRadius: 4,
     padding: padding(4, 16),
+    marginTop: 4,
+    marginBottom: 4,
     '&:not(:last-child)': {
       marginRight: 8
     }
-  },
-  multiline: { marginBottom: 8 }
+  }
 })
 
-export const XTagTableCellType = ({ model, serializedTags, isMultiline }: { model: TagTableCellType, serializedTags: S, isMultiline?: B }) => {
+export const XTagTableCellType = ({ model, serializedTags }: { model: TagTableCellType, serializedTags: S, isMultiline?: B }) => {
   const
     mapTags = ((tagLabel: S, i: U) => {
       const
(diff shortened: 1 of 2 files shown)


ui/src/tag_table_cell_type.tsx
```diff
@@ -66,7 +68,7 @@ export const XTagTableCellType = ({ model, serializedTags }: { model: TagTableCe
         background = cssVar(tagColor),
         color = cssVar(tag?.label_color || getContrast(tagColor))
 
-      return <span key={i} style={{ background, color }} className={clas(css.tag, 'wave-s12 wave-w6')}>{tagLabel}</span>
+      return <div key={i} style={{ background, color }} className={clas(css.tag, 'wave-s12 wave-w6', isMultiline ? css.multiline : '')}>{tagLabel}</div>
```
mturoci, line 71: Why do we need `div` instead of `span` here?

mturoci (dismissed):
LGTM @marek-mihok, thanks! Just a single question.

ui/src/tag_table_cell_type.tsx
marek-mihok: It was helpful in debugging process, but with current solution it's not necessary anymore. Replacing back with `span`.

mturoci (changes requested):
Actually, I noticed this is not a correct solution since it would disalign the tags as shown in your demo:

image

Edit: The tags should also be centered within the row at all times. Adding margin bottom breaks this alignment.

marek-mihok:
Good catch, @mturoci. I've fixed it by adding half-sized top and bottom margins instead of one bottom margin. This way tags are centered and the margins are present only for wrapped tags.

PR #1938, opened 2023-04-20
feat: Update all examples to use by-name access when updating components

### LINKED ISSUE #1882: Update all examples to use the new `by-name` access when updating components
#148 brought us by-name access to components which is a handy way to update components. It would be good to rewrite our existing examples to adhere to this new pattern.

image

Make sure all the examples work as before.

marek-mihok:
Closes #1882

### DIFF
--- a/py/examples/theme_generator.py
+++ b/py/examples/theme_generator.py
@@ -109,7 +109,7 @@ async def serve(q: Q):
             get_contrast('text', 'page', q),
             get_contrast('page', 'primary', q),
             ui.text_xl('Copy code'),
-            ui.frame(content=get_theme_code(q), height='180px'),
+            ui.frame(name='frame', content=get_theme_code(q), height='180px'),
         ])
         q.page['sample'] = ui.form_card(box='preview', items=[
             ui.text_xl(content='Sample App to show colors'),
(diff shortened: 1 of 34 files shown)


py/examples/theme_generator.py
```diff
@@ -42,9 +42,24 @@ def get_contrast(color1: str, color2: str, q: Q, min_contrast=4.5):
     darkest = min(lum1, lum2)
     contrast = (brightest + 0.05) / (darkest + 0.05)
     if contrast < min_contrast:
-        return ui.message_bar(type='error', text=f'Improve contrast between **{color1}** and **{color2}**.')
+        return False
     else:
-        return ui.message_bar(type='success', text=f'Contrast between **{color1}** and **{color2}** is great!')
+        return True
+    
+def update_contrast_check(color1: str, color2: str, q:Q, min_contrast=4.5):
```
mturoci, line 49: I suppose this function deprecates the original `get_contrast`. If so, `get_contrast` should be removed as it's not used anymore.

mturoci (changes requested):
Thanks @marek-mihok seems like the theme generator is broken now - copy code section should reflect the currently picked colors.

Master
image
This PR:
image

py/examples/theme_generator.py
marek-mihok: Yep, a leftover. Thanks.

mturoci (approved):

PR #1939, opened 2023-04-21
fix: Filter out empty strings for tags and filter options #1905

### LINKED ISSUE #1905: Empty space in table nor rendered properly for tags and also causing weird filter behavior
via @g-eoj 

#### Wave SDK Version, OS

0.25.2

#### Actual behavior

image

#### Expected behavior

If tag value is an empty string, nothing should be rendered. The empty option should also not be present in filters if the table column is filterable.

#### Steps To Reproduce

```py
# Table / Tags
# Use tags in order to emphasize a specific value. For multiple tags in a single row use `,` as a delimiter.
# ---
from faker import Faker

fake = Faker()

_id = 0


class Issue:
    def __init__(self, text: str, tag: str):
        global _id
        _id += 1
        self.id = f'I{_id}'
        self.text = text
        self.tag = tag


# Create some issues
issues = [Issue(text=fake.sentence(), tag=('' if i % 2 == 0 else 'DONE,SUCCESS')) for i in range(10)]

columns = [
    ui.table_column(name='text', label='Issue', min_width='400px'),
    ui.table_column(name='tag', label='Badge', filterable=True, cell_type=ui.tag_table_cell_type(name='tags', tags=[
        ui.tag(label='FAIL', color='$red'),
        ui.tag(label='DONE', color='#D2E3F8', label_color='#053975'),
        ui.tag(label='SUCCESS', color='$mint'),
    ])),
]


@app('/demo')
async def serve(q: Q):
    q.page['example'] = ui.form_card(box='1 1 -1 -1', items=[
        ui.table(
            name='issues',
            columns=columns,
            rows=[ui.table_row(name=issue.id, cells=[issue.text, issue.tag]) for issue in issues],
        )
    ])
    await q.page.save()

```

marek-mihok:
After fix:
image

Before fix:
image

Closes #1905

### DIFF
--- a/ui/src/table.test.tsx
+++ b/ui/src/table.test.tsx
@@ -77,6 +77,91 @@ describe('Table.tsx', () => {
     expect(getAllByRole('gridcell')[0].textContent).toBe('6/23/2022, 12:50:28 AM')
   })
 
+  it('Renders tags correctly', () => {
+    tableProps = {
+      ...tableProps,
+      columns: [
+        { name: 'colname1', label: 'col1' },
+        {
+          name: 'colname2',
+          label: 'col2',
+          filterable: true,
+          cell_type: {
+            tag: {
+              name: 'tags',
+              tags: [
+                { label: 'TAG1', color: 'red' },
+                { label: 'TAG2', color: 'green' },
+                { label: 'TAG3', color: 'blue' },
+              ]
+            }
+          }
+        },
+      ],
+      rows: [
+        { name: 'rowname1', cells: [cell11, 'TAG1'] },
+        { name: 'rowname2', cells: [cell21, 'TAG2,TAG3'] },
+        { name: 'rowname3', cells: [cell31, 'TAG2'] }
+      ]
+    }
+
+    const { container, getAllByRole, getAllByTestId } = render(<XTable model={tableProps} />)
+
+    fireEvent.click(container.querySelector('.ms-DetailsHeader-filterChevron')!)
+
+    const checkboxes = getAllByRole('checkbox')
+    expect(checkboxes).toHaveLength(3)
+
+    const [tag1, tag2, tag3] = checkboxes
+    expect(tag1).not.toBeChecked()
+    expect(tag2).not.toBeChecked()
+    expect(tag3).not.toBeChecked()
+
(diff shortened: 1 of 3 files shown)


ui/src/table.tsx
```diff
@@ -393,7 +393,7 @@ const
           /> : null
       }, [onFilterChange, setFiltersInBulk]),
       onColumnContextMenu = React.useCallback((col: WaveColumn, e: React.MouseEvent<HTMLElement>) => {
-        const menuFilters = col.filters || items.map(i => i[col.fieldName || col.key])
+        const menuFilters = (col.filters || items.map(i => i[col.fieldName || col.key])).filter(item => item !== '')
```
mturoci, line 396: Another loop through items seems unnecessary since they will contain a lot of duplicates anyway. The filter should be applied only once the duplicates are taken care of.

mturoci (changes requested):
Thanks @marek-mihok. Please add unit tests as well.

ui/src/table.test.tsx
```diff
@@ -77,6 +77,91 @@ describe('Table.tsx', () => {
     expect(getAllByRole('gridcell')[0].textContent).toBe('6/23/2022, 12:50:28 AM')
   })
 
+  it('Renders tags correctly', () => {
+    tableProps = {
+      ...tableProps,
+      columns: [
+        { name: 'colname1', label: 'col1' },
```
mturoci, line 84: I guess this one is not needed, right?

ui/src/table.test.tsx
```diff
@@ -77,6 +77,91 @@ describe('Table.tsx', () => {
...
+            tag: {
+              name: 'tags',
+              tags: [
+                { label: 'TAG1', color: 'red' },
+                { label: 'TAG2', color: 'green' },
+                { label: 'TAG3', color: 'blue' },
+              ]
+            }
+          }
+        },
```
mturoci, line 99: Can be extracted into a separate variable so that it can be reused for both tests.

mturoci (changes requested):

mturoci (approved):
Thanks @marek-mihok!

PR #1946, opened 2023-04-26
feat: Pin the python runtime version for Appstore Wave apps #1873

### LINKED ISSUE #1873: Wave Apps: Feature Deprecation Python 3.7
This is for updating the app.toml files of the following apps in this repo:
- [ ] Tour of Wave
- [ ] Wave University
- [ ] Dashboards 
- [ ] Themegenerator

https://github.com/h2oai/cloud-platform/issues/31

Python 3.10 will be the default runtime for apps in HAIC 23.04 and in 23.07 Python 3.7 will no longer be available. As a best practice, core apps should pin the runtime rather than using the system default. 

- [ ] Set the `RuntimeVersion` to  deb11_py310_wlatest for Python 3.10 (this will be the new default) in the 23.04.0 release in June
- [ ] Upload your app to cloud-dev to ensure it works as expected
- [ ] Let the #ai-app-store slack channel know if anything breaks so we can prioritize ensuring your app will work for customers when the new python release is set in June
- [ ] Create a new release of your app that explicitly sets the runtime rather than relying on the default

marek-mihok:
Screenshot 2023-04-26 at 14 33 28

Closes #1873

### DIFF
--- a/py/apps/dashboard/app.toml
+++ b/py/apps/dashboard/app.toml
@@ -9,6 +9,7 @@ LongDescription = "about.md"
 
 [Runtime]
 Module = "demo.app"
+RuntimeVersion = "deb11_py310_wlatest"
 
 [[Env]]
 Name = "TRACKER_ID"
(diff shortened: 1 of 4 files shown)


mturoci (approved):
Thanks @marek-mihok!

PR #1948, opened 2023-04-26
feat: Make tour examples to support path-based routing #1875

### LINKED ISSUE #1875: Test Tour of Wave with Path-Based Routing
When deployed in H2O's AI App Store, apps can choose to do Path-Based routing or Domain based routing. The previous default was domain based, but more and more customers are moving to path-based to reduce costs and improve security. Tour of Wave does not have a preference set in the app.toml file and does whatever the system has as a default. Because this app relies so heavily on routing, I want to ensure it does work well in a path-based environment, and that we set this as the default if there are no issues. 

**Requirements:**
- [ ] Add the following line to your `app.toml` `[Runtime]` section: `RoutingMode = "BASE_URL"`
- [ ] Test that your app works as expected when published to an AI App Store

marek-mihok:
A relative URL with `/` in the beginning replaces the entire path name of the base URL.
What we want here is to appended a relative URL to the full base URL and this can be achieved by removing the leading slash.

https://user-images.githubusercontent.com/23740173/234636257-ad5ee447-fff2-46b2-a066-3e60bfe1f4b1.mov

Closes #1875

### DIFF
--- a/py/apps/tour/app.toml
+++ b/py/apps/tour/app.toml
@@ -8,4 +8,5 @@ LongDescription = "about.md"
 
 [Runtime]
 Module = "examples.tour"
+RoutingMode = "BASE_URL"
 RuntimeVersion = "deb11_py310_wlatest"
--- a/py/examples/link.py
+++ b/py/examples/link.py
@@ -9,10 +9,10 @@ page = site['/demo']
 page['example'] = ui.form_card(
     box='1 1 4 7',
     items=[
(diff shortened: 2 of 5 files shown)


mturoci (approved):
LGTM. Thanks @marek-mihok!

marek-mihok:
> please resolve the conflicts and we are good to merge

@mturoci , done  ✅

PR #1953, opened 2023-04-28
feat: Wave share client.

mturoci:
Python client for reverse proxy tunnel, integrated into Wave CLI.

image

Detailed documentation to be done in #1954 once this PR is merged and proxy tunnel server deployed in prod.

### DIFF
--- a/py/h2o_wave/h2o_wave/cli.py
+++ b/py/h2o_wave/h2o_wave/cli.py
@@ -12,24 +12,29 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import time
-import tarfile
-import shutil
-import sys
-import socket
-from contextlib import closing
-import subprocess
-from pathlib import Path
-import platform
-import uvicorn
-import click
-import inquirer
 import os
-from click import Choice, option
+import platform
+import shutil
+import socket
+import subprocess
+import sys
+import tarfile
+import time
+from contextlib import closing
+from pathlib import Path
+from threading import Thread
 from urllib import request
 from urllib.parse import urlparse
+
+import click
+import httpx
+import inquirer
+import uvicorn
+from click import Choice, option
+from h2o_wave.share import listen_on_socket
+
+from .metadata import __arch__, __platform__
 from .version import __version__
-from .metadata import __platform__, __arch__
 
 _localhost = '127.0.0.1'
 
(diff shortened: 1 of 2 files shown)


py/h2o_wave/h2o_wave/cli.py
```diff
@@ -276,4 +284,70 @@ def learn():
...
+        print('Could not connect to the remote sharing server.')
+        exit(1)
+
+    res = res.json()
+    print(f'BETA: Proxying localhost:{port} ==> {res["url"]}')
+    print('\x1b[7;30;43mDO NOT SHARE YOUR APP IF IT CONTAINS SENSITIVE INFO\x1b[0m.')
+    print('Press Ctrl+C to stop sharing.')
+
+    max_conn_count = res['max_conn_count']
+    step = 100 if max_conn_count > 10 else max_conn_count
```
geomodular, line 342: Is this condition alright? Looks fishy.

py/h2o_wave/h2o_wave/share.py
```diff
@@ -0,0 +1,26 @@
...
+    while True:
+        data = await r.read(4096)
+        if not data:
+            break
+        w.write(data)
+        await w.drain()
+
+
+async def listen_on_socket(local_host: str, local_port: int, remote_host: str, remote_port: int) -> None:
+    while True:
```
geomodular, line 14: Should this be an infinite cycle?

geomodular (dismissed):
Thank you Martin; looks good although I'm a bit rusty with `asyncio`.

py/h2o_wave/h2o_wave/cli.py
mturoci: Good question. This piece probably deserves a comment in the code with further explanations. It's there because there are 2 possible server configurations:

1. The "community" one used for regular Wave users - limits to 10 concurrent connections per client.
2. Using this reverse proxy to tunnel a single app only with significantly more concurrent connections (10k+).

The condition is there simply to create the server connections faster (step of 100 instead of 10 => effectively 10x faster).

Does it make sense?

py/h2o_wave/h2o_wave/share.py
mturoci: I believe so. The client should expose the app until the process is stopped via CTRL+C. Note that Wave uses HTTP 1.1 as of today meaning once the request is done (e.g. CSS resource), the connection is closed, signaling the requestor there is no more data to come, thus a new connection needs to be made and added to the pool.

geomodular: Yes, you are right. I somewhat didn't realize this was a main loop.

PR #1958, opened 2023-05-02
feat: Support external links in navigation

### LINKED ISSUE #1952: Support external links for ui.nav_item
Via #1204: ui.nav_item currently supports hash-change navigation only. 

It would be nice to introduce a new `path` property for `ui.nav_item` to support opening external websites from side navigation.

marek-mihok:
This change adds the `path` attribute for `ui.nav_item`.

Updated API:
```ts
/** Create a navigation item. */
export interface NavItem {
  /** The name of this item. Prefix the name with a '#' to trigger hash-change navigation. */
  name: Id
  /** The label to display. */
  label: S
  /** An optional icon to display next to the label. */
  icon?: S
  /** True if this item should be disabled. */
  disabled?: B
  /** An optional tooltip message displayed when a user hovers over this item. */
  tooltip?: S,
  /** The path or URL to link to. If specified, the `name` is ignored. The URL is opened in a new browser window or tab. E.g. `/foo.html` or `http://example.com/foo.html` */
  path?: S
}
```

https://user-images.githubusercontent.com/23740173/235626410-44d577fd-748c-4d18-a2d8-222a87970429.mov

Closes #1952


mturoci (approved):
Thanks @marek-mihok, just one small nitpick comment.

PR #1960, opened 2023-05-02
feat: Make tour responsive

marek-mihok:
This change enhances the Wave **tour app** experience with the support for smaller screens.

https://github.com/h2oai/wave/assets/23740173/7a528c04-a1fd-4009-82ee-cb81337ed736

### DIFF
--- a/py/examples/tour.py
+++ b/py/examples/tour.py
@@ -281,17 +281,17 @@ def make_blurb(q: Q):
     blurb_card.title = example.title
     blurb_card.subtitle = example.description
     # HACK: Recreate dropdown every time (by dynamic name) to control value (needed for next / prev btn functionality).
-    items = [ui.dropdown(name=q.args['#'] or default_example_name, width='300px', value=example.name, trigger=True,
-                         choices=[ui.choice(name=e.name, label=e.title) for e in catalog.values()])]
+    dropdown = ui.dropdown(name=q.args['#'] or default_example_name, width='300px', value=example.name, trigger=True,
+                         choices=[ui.choice(name=e.name, label=e.title) for e in catalog.values()])
     buttons = []
     if example.previous_example:
         buttons.append(ui.button(name=f'#{example.previous_example.name}', label='Previous'))
     if example.next_example:
         buttons.append(ui.button(name=f'#{example.next_example.name}', label='Next', primary=True))
-    blurb_card.items = items + buttons
+    blurb_card.items = [dropdown] + buttons
     q.page['mobile_blurb'].items = [ui.inline(
         direction='column',
-        items= items + [ui.text_s(example.description), ui.inline(direction='row', items=buttons)]
+        items= [dropdown] + [ui.text_s(example.description), ui.inline(direction='row', items=buttons)]
     )]
             
 
--- a/py/examples/tour.py
+++ b/py/examples/tour.py
@@ -258,17 +258,17 @@ def get_wave_completions(line, character, file_content):
         title=app_title,
         subtitle=f'{len(catalog)} Interactive Examples',
         image=f'{q.app.tour_assets}/h2o-logo.svg',
-        items=[ui.links(inline=True,
-                        items=list(map(lambda item: ui.link(label=item[1], path=item[2], target='_blank'), nav_links)))]
-    )
+        items=[
+            ui.links(inline=True, items=[ui.link(label=item[1], path=item[2], target='_blank') for item in nav_links])
+        ])
     q.page['mobile_header'] = ui.header_card(
         box='mobile_header',
         title=app_title,
         subtitle=f'{len(catalog)} Interactive Examples',
         image=f'{q.app.tour_assets}/h2o-logo.svg',
-        nav=[ui.nav_group('Links', items=list(
-            map(lambda item: ui.nav_item(name=item[0], label=item[1], path=item[2]), nav_links)))]
-    )
+        nav=[
+            ui.nav_group('Links', items=[ui.nav_item(name=item[0], label=item[1], path=item[2]) for item in nav_links])
+        ])
     q.page['blurb'] = ui.section_card(box='blurb', title='', subtitle='', items=[])
     q.page['mobile_blurb'] = ui.form_card(box='mobile_blurb', items=[])
     q.page['code'] = ui.markup_card(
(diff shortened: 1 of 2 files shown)


py/examples/tour.py
```diff
@@ -220,31 +222,49 @@ def get_wave_completions(line, character, file_content):
...
-        items=[
-            ui.links(inline=True, items=[
-                ui.link(label='Wave docs', path='https://wave.h2o.ai/docs/getting-started', target='_blank'),
-                ui.link(label='Discussions', path='https://github.com/h2oai/wave/discussions', target='_blank'),
-                ui.link(label='Blog', path='https://wave.h2o.ai/blog', target='_blank'),
-                ui.link(label='H2O AI Cloud', path='https://h2o.ai/platform/ai-cloud/', target='_blank'),
-                ui.link(label='H2O', path='https://www.h2o.ai/', target='_blank'),
-            ])
-        ]
+        items=[ui.links(inline=True, items=list(map(lambda item: ui.link(label=item[1], path=item[2], target='_blank'), nav_links)))]
```
mturoci, line 257: Check Python list comprehension to make this more pythonic and less like JS. Same applies for the other header card.

mturoci (changes requested):
Thanks @marek-mihok. I expected something more like an actual mobile app.

image

* The navigation toolbar with search and buttons should be at the bottom to make for better UX.
* Let's leave out the description for mobile screens to save space.
* Maximize the code area space by cutting out margins where possible

Edit: Please set up a linter or use PyCharm.

py/examples/tour.py
```diff
@@ -220,31 +222,49 @@ def get_wave_completions(line, character, file_content):
...
-            ui.layout(breakpoint='xs', zones=[
+            ui.layout(
+                breakpoint='xs',
+                zones=[
+                    ui.zone('mobile_header'),
+                    ui.zone('mobile_blurb'),
+                    ui.zone('main', size=f'calc(100vh - {header_height + mobile_blurb_height}px)', direction=ui.ZoneDirection.COLUMN, zones=[
+                        ui.zone('code', size='50vh'),
+                        ui.zone('preview', size='50vh'),
+                    ])
```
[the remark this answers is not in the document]
marek-mihok: It's only for avoiding "magic" numbers.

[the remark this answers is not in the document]
mturoci: > It's only for avoiding "magic" numbers.

The question is why the calculation (regardless of with or without magic numbers) is necessary here. It's suspicious since the code calculates the size of the parent (main) zone and then ignores it completely by setting both children to `50vh`, but maybe I am missing something?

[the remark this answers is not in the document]
mturoci: How is it fixed if both calculations (parent and children) are still there?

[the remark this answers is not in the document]
marek-mihok: You are right, the parent (`main`) calculation is redundant. Removed.

py/examples/tour.py
```diff
@@ -220,35 +224,57 @@ def get_wave_completions(line, character, file_content):
         scripts=[ui.script(q.app.tour_assets + '/loader.min.js')],
         script=ui.inline_script(content=template, requires=['require'], targets=['monaco-editor']),
         layouts=[
-            ui.layout(breakpoint='xs', zones=[
+            ui.layout(
+                breakpoint='xs',
+                zones=[
+                    ui.zone('mobile_header'),
+                    ui.zone('main', direction=ui.ZoneDirection.COLUMN,
+                            zones=[
```
mturoci, line 232: No need for being explicit.

```suggestion
                    ui.zone('main', zones=[
```

py/examples/tour.py
```diff
@@ -220,35 +224,57 @@ def get_wave_completions(line, character, file_content):
...
-    )
+            ui.links(inline=True, items=[ui.link(label=item[1], path=item[2], target='_blank') for item in nav_links])
+        ])
+    q.page['mobile_header'] = ui.header_card(
+        box='mobile_header',
+        title=app_title,
+        subtitle=f'{len(catalog)} Interactive Examples',
+        image=f'{q.app.tour_assets}/h2o-logo.svg',
+        nav=[
+            ui.nav_group('Links', items=[ui.nav_item(name=item[0], label=item[1], path=item[2]) for item in nav_links])
```
mturoci, line 270: Naming.
```suggestion
            ui.nav_group('Links', items=[ui.nav_item(name=link[0], label=link[1], path=link[2]) for link in nav_links])
```

py/examples/tour.py
```diff
@@ -220,35 +224,57 @@ def get_wave_completions(line, character, file_content):
...
+        ])
     q.page['blurb'] = ui.section_card(box='blurb', title='', subtitle='', items=[])
+    q.page['mobile_blurb'] = ui.form_card(box='mobile_blurb', items=[])
     q.page['code'] = ui.markup_card(
-        box=ui.box('code', height='calc(100vh - 140px)'),
+        box=ui.box('code', height=f'calc(100vh - {header_height + blurb_height}px)'),
```
mturoci, line 275: Is this needed if we already set the size on the layout level?
```suggestion
        box='code',
```

py/examples/tour.py
```diff
@@ -263,11 +289,16 @@ def make_blurb(q: Q):
...
-        items.append(ui.button(name=f'#{example.previous_example.name}', label='Previous'))
+        buttons.append(ui.button(name=f'#{example.previous_example.name}', label='Previous'))
+        mobile_buttons.append(ui.button(name=f'#{example.previous_example.name}', label='Prev'))
     if example.next_example:
-        items.append(ui.button(name=f'#{example.next_example.name}', label='Next', primary=True))
-    blurb_card.items = items
+        buttons.append(ui.button(name=f'#{example.next_example.name}', label='Next', primary=True))
+        mobile_buttons.append(ui.button(name=f'#{example.next_example.name}', label='Next', primary=True))
+    blurb_card.items = items + buttons
+    q.page['mobile_blurb'].items = [ui.inline(direction='row', justify='center', items=items + mobile_buttons)]
```
mturoci, line 301: Let's simplify this. No need for extra `buttons` and `mobile_buttons` as it seems like there is a difference only in `label` (correct me if I am wrong).

GH won't let me add suggestions on deleted lines so here is the whole code.

```py
    if example.previous_example:
        items.append(ui.button(name=f'#{example.previous_example.name}', label='Prev'))
    if example.next_example:
        items.append(ui.button(name=f'#{example.next_example.name}', label='Next', primary=True))
    q.page['mobile_blurb'].items = [ui.inline(direction='row', justify='center', items=items)]
```

mturoci (changes requested):

py/examples/tour.py
marek-mihok: In master it is like this: `box=ui.box('code', height='calc(100vh - 140px)'),`. I couldn't find a reason why. It works the same when the calculation is removed. Therefore I accept your suggestion.

py/examples/tour.py
marek-mihok: I agree. Btw, is my robust implementation the only way how to achieve different labels for smaller screens in Wave?

mturoci: [Robustness](https://en.wikipedia.org/wiki/Robustness_(computer_science)) means something different. If the labels really need to be different, I would go for a label update where needed, e.g. `q.page['mobile_blurb'].items....label = 'Prev'`. This way all there is needed is a single line and no extra variables.

mturoci (approved):

PR #1961, opened 2023-05-03
fix: Add disabled choices support for popup='always' dropdowns #1959

### LINKED ISSUE #1959: Dropdown choice cannot be disabled when popup set to always
#### Wave SDK Version, OS
0.25.2, Mac OS

#### Actual behavior
The choice is always selectable irrespective of the disabled attribute of the choice being set to True when the value of the popup attribute is set to always.

#### Expected behavior
Choices should be disabled in the popup when the choice is disabled.

#### Steps To Reproduce
Use following sample code to reproduce issue.

```python

choices = [
    ui.choice('A', 'Option A'),
    ui.choice('B', 'Option B'),
    ui.choice('C', 'Option C', disabled=True),
    ui.choice('D', 'Option D'),
]

@app('/')
async def serve(q: Q):
    q.page['example'] = ui.form_card(box='1 1 4 7', items=[
        ui.dropdown(name='dropdown', label='With Popup (Disable work)', required=True, choices=choices,values=['A']),
        ui.dropdown(name='dropdown_popup_always', label='Popup Always (Disable Not Working)', values=['A'],
                    required=True, choices=choices, popup='always'),

    ])
    await q.page.save()
```

marek-mihok:
https://user-images.githubusercontent.com/23740173/235876867-7c9de421-61ee-407b-99e3-3bc8fa8b7db5.mov

Closes #1959


mturoci (approved):
Thanks @marek-mihok, LGTM with a minor comment.

What is submitted if I specify the disabled choice as checked initially and then do my selection for other (non-disabled) choices? Is the initial disabled choice part of the final submission or not?

ui/src/dropdown.test.tsx
```diff
@@ -368,6 +368,33 @@ describe('Dropdown.tsx', () => {
       expect(wave.args[name]).toMatchObject(['1', '2'])
     })
 
+    it('Selects all options on Select all - except disabled', () => {
```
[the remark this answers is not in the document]
marek-mihok: There is a different handling of this functionality for basic and dialog dropdown. Therefore renaming accordingly.

marek-mihok:
> What is submitted if I specify the disabled choice as checked initially and then do my selection for other (non-disabled) choices? Is the initial disabled choice part of the final submission or not?

It is but for dialog dropdown only.

https://user-images.githubusercontent.com/23740173/237037407-0a3202b2-bcd3-481e-b8f1-6266369ecb16.mov

For base dropdown it is only when doing selections manually but it is not when using `Select all/Deselect all`. Since this issue relates to dialog dropdown only, I've created [a separate issue](https://github.com/h2oai/wave/issues/1963) for the base dropdown last week.

mturoci:
> it is but for dialog dropdown only.

Both dialog and regular dropdown must always be in sync so let's fix them both in this PR.

marek-mihok:
@mturoci, both of them are fixed now.

ui/src/dropdown.tsx
```diff
@@ -109,24 +110,19 @@ const
         // HACK: Push clears args so run it after useEffect sets them due to model.value change.
         if (trigger) setTimeout(() => wave.push(), 0)
       },
-      selectAll = () => {
+      selectAll = (checked = true) => () => {
```
mturoci, line 113: Let's name this `select` (imperative) so that it's clear we should either select or deselect within the function. 

```suggestion
      selectAll = (select = true) => () => {
```

ui/src/dropdown.tsx
```diff
@@ -109,24 +110,19 @@ const
...
-        setMultiValues(selectionArr)
-        wave.args[name] = selectionArr
-
-        onChange()
-      },
-      deselectAll = () => {
-        if (!selection) return
+        options.forEach(({ disabled, key }) => {
+          checked
+            ? !disabled || values?.includes(key as S) ? selection.add(key as S) : null
```
mturoci, line 119:
```js
? !disabled || values?.includes(key as S) ? selection.add(key as S) : null
```

Clearing whole selection and then putting things back seems odd. Isn't there a more elegant way?

mturoci (changes requested):

ui/src/dropdown.tsx
marek-mihok: Considering we want to keep the Javascript Set for storing the selection in the `BaseDropdown`, there is an option of adding without clearing, but in this case selected choices are not in the original order so it is inconsistent with a `DialogDropdownMulti` where selected choices are in the original order when using Select all.

https://github.com/h2oai/wave/assets/23740173/4dceadbc-0822-427f-b442-4f2be6a0a732

Another option would be to go without Set.

ui/src/dropdown.tsx
mturoci: Yes.

mturoci: Ideally, the intention should be to make the code better in all places rather than defaulting to suboptimal naming for the sake of keeping consistency. In this case, the change lies in variable naming only so a no-brainer.

ui/src/dropdown.tsx
mturoci: > but in this case selected choices are not in the original order

What is the definition of "original" order?

> Another option would be to go without Set.

Yes, at first glance, Set seems redundant here.

> it is inconsistent with a DialogDropdownMulti

Both dropdown types must behave exactly the same, achieved by the most straightforward code possible.

marek-mihok: > What is the definition of "original" order?

As present in `choices` model value.

mturoci (approved):

PR #1987, opened 2023-05-24
fix: Perform numerical sort in map buffer if possible.

mturoci:
This PR makes sure that map buffer entries are sorted in numerical order (if keys are valid numbers), defaulting to regular string sort.

## Needs discussion

@lo5 this is a possible breaking change although I don't think there are many people relying on string sort order. There are ways around making this non-breaking, but not sure if worth the hassle + IMO this should be the right behavior. Wdyt?

### DIFF
--- a/ts/index.ts
+++ b/ts/index.ts
@@ -626,7 +626,11 @@ const
       },
       list = (): Rec[] => {
         const keys = keysOf(tups)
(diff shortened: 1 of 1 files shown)


mturoci:
Closing in favor of a better proposal.

PR #2001, opened 2023-05-31
feat: Notify about client disconnect.

mturoci:
Required by #1979 as it spawns a new Wave app process per client and needs to perform proper clean up.


lo5 (approved):
Thanks @mturoci!

PR #2004, opened 2023-06-02
feat: Support percentage height for ui.form components #326

### LINKED ISSUE #326: Table content overlaps when we use height and tooltip together
**Q SDK Version, OS**
0.9.1

**Actual behavior**

I’m using a table inside a form_card which is the only component of that particular page. I want the table to be expanded to take a full-page width and height. When I add a tooltip to the page content gets overlapped and the table component won’t be displayed properly. Without the tooltip, it works fine. 

Screenshot 2020-11-25 at 16 07 20

**Expected behavior**
Table content should be extended to the full component.
Screenshot 2020-11-25 at 16 07 54

**Steps To Reproduce**
1. Use the table component inside the form_card with the below properties.

```py
q.page["risk_table"] = ui.form_card(box="1 2 -1 -1", items=[
  ui.table(
    name='risk_table',
    columns=get_column_headers_for_df(df, True),
    rows=get_rows(q, df),
    groupable=True,
    resettable=True,
    multiple=False,
    height='100%',
    tooltip='Double click a row to go to select it.'
  )
])
```

2. Access the app URL in the browser.

marek-mihok:
https://github.com/h2oai/wave/assets/23740173/370a6702-37d3-44d3-8670-2b22c349d1dd

Closes #326

### DIFF
--- a/ui/src/copyable_text.tsx
+++ b/ui/src/copyable_text.tsx
@@ -57,10 +57,7 @@ export interface CopyableText {
 export const XCopyableText = ({ model }: { model: CopyableText }) => {
   const
     { name, multiline, label, value, height } = model,
-    isCustomHeight = height && height !== '1',
-    isFullHeight = multiline && height === '1',
-    heightStyle = isCustomHeight || isFullHeight ? fullHeightStyle : undefined,
-    minHeight = isCustomHeight ? height : undefined,
+    heightStyle = multiline && height === '1' ? fullHeightStyle : undefined,
     ref = React.useRef<Fluent.ITextField>(null),
     timeoutRef = React.useRef<U>(),
     [copied, setCopied] = React.useState(false),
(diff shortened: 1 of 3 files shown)


marek-mihok:
@mturoci just fyi, I went through every tour example, tested it in every browser, ran the regression tests and it is working fine, nothing is broken. The support of setting the height in other CSS units is untouched - it works as before. And this implementation works inside flex layouts too.

mturoci:
Sounds good. Can you elaborate more on what was the problem and why this is the right solution? Did this happen for the table only or were any other components affected as well?

> And this implementation works inside responsive layouts too.

I am wondering how this is possible since percentage height requires the parent container to have constant width specified.

marek-mihok:
I'm closing this PR as setting percentage units for the height of ui.form components is ambiguous - 50% of what? Of the parent ui.inline? Of the ui.form, of the card, of the page or viewport?

We will go with the implementation of setting the `height='1'` option to fill the all available height for relevant components like we already did for [ui.copyable_text](https://github.com/h2oai/wave/pull/1879) or [ui.table](https://github.com/h2oai/wave/pull/2014).

marek-mihok:
I'm closing this PR as setting percentage units for the height of ui.form components is ambiguous - 50% of what? Of the parent ui.inline? Of the ui.form, of the card, of the page or viewport?

We will go with the implementation of setting the `height='1'` option to fill the all available height for relevant components like we already did for [ui.copyable_text](https://github.com/h2oai/wave/pull/1879) or [ui.table](https://github.com/h2oai/wave/pull/2014).

PR #2008, opened 2023-06-02
feat: Make chatbot messages to fill the entire line #1998

### LINKED ISSUE #1998: Make chatbot messages fill entire line to improve reading UX

The current chatbot card is more like a messenger chat. However, chatbots usually output a lot of text so it's better if we can use the whole real estate provided by the screen.

Instead of limited left/right aligned messages, let's center each message and fill the entire row (max out at `65ch` for optimal readability).

marek-mihok:
This PR also fixes the issue where text was not changing color when switching themes dynamically. 

https://github.com/h2oai/wave/assets/23740173/a4f6bac6-3d6c-4d3c-9cea-0b288ee8819c

Closes #1998


marek-mihok:
@mturoci, to simplify your PR review process, here is the example used in my recording:
```py
from h2o_wave import main, app, Q, ui, data

MAX_MESSAGES = 500

choices = [
    ui.choice('default', 'default'),
    ui.choice('h2o-dark', 'h2o-dark'),
    ui.choice('one-dark-pro', 'one-dark-pro'),
    ui.choice('monokai', 'monokai'),
    ui.choice('nord', 'nord'),
    ui.choice('winter-is-coming', 'winter-is-coming'),
    ui.choice('fuchasia', 'fuchasia'),
    ui.choice('nature', 'nature'),
    ui.choice('solarized', 'solarized'),
    ui.choice('oceanic', 'oceanic'),
    ui.choice('ember', 'ember'),
    ui.choice('lighting', 'lighting'),
    ui.choice('kiwi', 'kiwi'),
    ui.choice('benext', 'benext'),
    ui.choice('neon', 'neon')
]

@app('/demo')
async def serve(q: Q):
    if not q.client.initialized:
        q.page['meta'] = ui.meta_card(
            box='',
            theme='h2o-dark',
            layouts=[
                ui.layout(
                    breakpoint='xs',
                    zones=[
                        ui.zone(name='header', size='1'),
                        ui.zone(name='main', size='8')
                    ]
                )
            ]
        )
        # Cyclic buffer drops oldest messages when full. Must have exactly 2 columns - msg and fromUser.
        cyclic_buffer = data(fields='msg fromUser', size=-MAX_MESSAGES)
        q.page['form'] = ui.form_card(box='header', items=[
            ui.dropdown(name='theme', label='Theme', trigger=True, choices=choices),
        ])
        q.page['example'] = ui.chatbot_card(box='main', data=cyclic_buffer, name='chatbot')
        q.client.initialized = True

    # A new message arrived.
    if q.args.chatbot:
        # Append user message.
        q.page['example'].data[-1] = [q.args.chatbot, True]
        # Append bot response.
        q.page['example'].data[-1] = ['I am a fake chatbot. Sorry, I cannot help you.', False]

    if q.args.theme:
        q.page['meta'].theme = q.args.theme

    await q.page.save()

```

ui/src/chatbot.tsx
```diff
@@ -101,7 +96,9 @@ export const XChatbot = ({ model }: { model: Chatbot }) => {
       setUserInput('')
     }
 
-  React.useEffect(() => { if (msgContainerRef.current) msgContainerRef.current.scrollTop = msgContainerRef.current.scrollHeight }, [msgs])
+  React.useEffect(() => {
+    if (msgContainerRef.current) msgContainerRef.current.scrollTo({ top: msgContainerRef.current.scrollHeight, behavior: 'smooth' })
```
mturoci, line 100: ScrollTo is only [supported](https://caniuse.com/?search=scrollTo) on Safari (fully) & Edge from 2020, let's add a fallback to the previous `scrollTop` in case it's not available.

ui/src/chatbot.tsx
```diff
@@ -32,44 +32,39 @@ const
...
+    textInput: {
+      position: 'absolute',
+      padding: 15,
+      bottom: 0,
+      left: 0,
+      right: 0,
+      backgroundImage: `linear-gradient(180deg,rgba(0,0,0,0),${cssVar('$card')} 45%)`
     }
-  }),
-  getCornerStyle = (prev?: B, curr?: B, next?: B): React.CSSProperties | undefined => {
```
mturoci, line 60: Do not forget to remove the accompanying unit tests as well.

ui/src/chatbot.tsx
```diff
@@ -32,44 +32,39 @@ const
...
       right: 0,
-      bottom: 62, // Height of input box + padding.
+      bottom: 0,
       overflowY: 'auto',
-      padding: padding(0, 15),
+      padding: padding(0, 0),
+    },
+    msgWrapper: {
+      '&:first-child': { marginTop: important(px(0)) },
+      '&:last-child': { marginBottom: important(px(100)) },
```
mturoci, line 44: Isn't 100 too much? Maybe an explanation comment is needed?

ui/src/chatbot.tsx
```diff
@@ -32,44 +32,39 @@ const
...
-    },
-    msgWrapper: {
-      '&:first-child': { marginTop: important(px(0)) },
+    textInput: {
+      position: 'absolute',
+      padding: 15,
+      bottom: 0,
+      left: 0,
+      right: 0,
+      backgroundImage: `linear-gradient(180deg,rgba(0,0,0,0),${cssVar('$card')} 45%)`
```
mturoci, line 65: Leftover?

ui/src/chatbot.tsx
```diff
@@ -110,29 +107,29 @@ export const XChatbot = ({ model }: { model: Chatbot }) => {
...
             key={idx}
-            className={css.msgWrapper}
+            className={clas(css.msgWrapper, fromUser ? css.userMsg : '')}
             style={{
-              marginTop: msgs[idx - 1]?.fromUser !== fromUser ? 3 : 10,
-              textAlign: fromUser ? 'right' : 'left',
-              color: getContrast(fromUser ? '$themePrimary' : '$text')
+              paddingTop: msgs[idx - 1]?.fromUser !== fromUser ? rem(0.7) : 0,
+              paddingBottom: msgs?.[idx + 1]?.fromUser !== fromUser ? rem(0.85) : 0,
+              color: cssVar(fromUser ? '$neutralDark' : '$themePrimary')
```
mturoci, line 114: Let's stick with a single text color that is easy to read. I would maybe even consider using `getContrast` function to be sure the text is properly visible in all cases.
```suggestion
              color: cssVar('$neutralDark')
```

ui/src/chatbot.tsx
```diff
@@ -32,44 +32,39 @@ const
...
+      '&:first-child': { marginTop: important(px(0)) },
+      '&:last-child': { marginBottom: important(px(100)) },
+      display: 'flex',
+      justifyContent: 'center',
     },
     msg: {
+      width: '65ch',
+    },
+    userMsg: {
+      backgroundColor: cssVar('$neutralLighter'),
```
mturoci, line 52: Let's swap the current background colorss for bot and user - user should have no background (card color) and bot neutralLighter as his responses are more important (should be kinda highlighted).

mturoci (changes requested):
Thanks @marek-mihok. Looking good, just a few comments.

ui/src/chatbot.tsx
marek-mihok: Nope. It makes a chat feed to fade at the bottom.

Screenshot 2023-06-12 at 11 03 02

Screenshot 2023-06-12 at 11 03 50

ui/src/chatbot.tsx
marek-mihok: A problem with `getContrast` is that it does not work dynamically when switching themes.

But I agree with the single easy to read text color.

ui/src/chatbot.tsx
marek-mihok: On Safari only `smooth` behavior is not working in 2017-2020 versions but otherwise it works the same as setting the `scrollTop` (without the animation). That's why we previously agreed using it in the image popup (lightbox). But I'll add a fallback for older versions of MS Edge.

ui/src/chatbot.tsx
marek-mihok: A text input with paddings has a height of 62px so the bottom margin of the last message needs to be greater. I agree 100px can be too much, so I'll stick with 70 for now and increase it once I'll be adding [a stop button](https://github.com/h2oai/wave/issues/2011#event-9459684698) there.

mturoci: Please don't forget to add a comment to avoid using magic numbers like this one.

ui/src/chatbot.tsx
mturoci: Let's use a proper layout to make sure the scrollbar ends where it should - above the textbox, rather than continuing beyond. Same as before:

image

ui/src/chatbot.tsx
mturoci: > A problem with getContrast is that it does not work dynamically when switching themes.

Wouldn't a rerender of the component on the theme change be enough?

mturoci (approved):
LGTM, thanks @marek-mihok!

PR #2014, opened 2023-06-08
feat: Allow ui.table to fill the remaining card height

### LINKED ISSUE #2013: Allow specifying '1' for height in ui.table - fill remaining space
As reported in #326, there is no option for table to fill all of the available height when specifying a tooltip and since [this commit](https://github.com/h2oai/wave/commit/3c574941f37b0e5294e522e2c013c3256ecf3a34) it is not possible even when tooltip is not specified.

It would be good to introduce a similar mechanism as for `ui.inline` or [`ui.copyable_text`](https://github.com/h2oai/wave/issues/1823) where this behavior can be achieved through specifying the `height='1'` prop.

marek-mihok:
https://github.com/h2oai/wave/assets/23740173/efec4285-74ea-448d-9d42-15e5b819894c

Closes #2013

### DIFF
--- a/ui/src/table.tsx
+++ b/ui/src/table.tsx
@@ -991,7 +991,6 @@ export const
       <div data-test={m.name} style={{
         position: 'relative',
         flexGrow: isFullHeight ? 1 : 0,
-        minHeight: isFullHeight ? 100 : 'auto',
         height: isFullHeight ? 'auto' : computeHeight()
       }}>
         <Fluent.Stack horizontal>
--- a/ui/src/table.tsx
+++ b/ui/src/table.tsx
@@ -136,7 +136,7 @@ export interface Table {
   downloadable?: B
   /** Indicates whether a Reset button should be displayed to reset search / filter / group-by values to their defaults. Defaults to False. */
   resettable?: B
-  /** The height of the table, e.g. '400px', '50%', etc. */
+  /** Custom height of the table in px (e.g. '200px') or use '1' to fill the remaining card space. */
   height?: S
   /** The width of the table, e.g. '100px'. Defaults to '100%'. */
   width?: S
@@ -696,6 +696,7 @@ export const
         , [m.rows, m.groups, getItem]),
       isMultiple = Boolean(m.values?.length || m.multiple),
       isSingle = Boolean(m.value || m.single),
+      isFullHeight = Boolean(m.height === '1'),
       [filteredItems, setFilteredItems] = React.useState(items),
       [currentPage, setCurrentPage] = React.useState(1),
       searchableKeys = React.useMemo(() => m.columns.filter(({ searchable }) => searchable).map(({ name }) => name), [m.columns]),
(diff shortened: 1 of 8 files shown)


ui/src/table.tsx
```diff
@@ -696,6 +696,7 @@ export const
         , [m.rows, m.groups, getItem]),
       isMultiple = Boolean(m.values?.length || m.multiple),
       isSingle = Boolean(m.value || m.single),
+      isFullHeight = Boolean(m.height === '1'),
```
mturoci, line 699: Unnecessary cast.

```suggestion
      isFullHeight = m.height === '1',
```

ui/src/table.tsx
```diff
@@ -136,7 +136,7 @@ export interface Table {
   downloadable?: B
   /** Indicates whether a Reset button should be displayed to reset search / filter / group-by values to their defaults. Defaults to False. */
   resettable?: B
-  /** The height of the table, e.g. '400px', '50%', etc. */
+  /** Custom height of the table in px (e.g. '200px') or use '1' to fill the remaining card space. */
```
mturoci, line 139:
```suggestion
  /** The height of the table in px (e.g. '200px') or '1' to fill the remaining card space. */
```

website/widgets/form/table.md
```diff
@@ -285,11 +285,12 @@ needed. For tables with > 10 rows, the initial height is `500px`.
 In some cases though, it might be desirable to control the dimensions yourself via `width` and `height`
 attributes. Both accept [CSS units](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units),
 however `%` values for height might not work as you think (especially in [flex layout](/docs/layout#flex-layout))
-so we discourage its use in favor of more static units like `px` or `rem`.
+so we discourage its use in favor of more static units like `px` or `rem`. If you wish to use all of the available height,
```
mturoci, line 288: Let's add a new paragraph with a separate `height='1'` example rather than changing the existing one.

ui/src/tooltip.tsx
```diff
@@ -45,21 +47,23 @@ const
...
     content?: S,
     showIcon?: B,
     expand?: B
   }) => {
     if (!content) return <>{children}</>
 
-    const tooltipProps: Fluent.ITooltipProps = { onRenderContent: () => <div><Markdown source={content} /></div> }
+    const
+      isFullHeight = children.props.model?.height === '1',
+      tooltipProps: Fluent.ITooltipProps = { onRenderContent: () => <div><Markdown source={content} /></div> }
```
mturoci, line 59: Is the surrounding `<div>` tag necessary?

mturoci (changes requested):
Thanks @marek-mihok! Is this tested to not break existing components with tooltips (regular or within ui.inline) across all supported browsers?

ui/src/tooltip.tsx
marek-mihok: Apparently not, but additional testing needs to be done to ensure there is no reason why it was originally implemented with a `<div>` tag.

marek-mihok: After exhausting testing it seems not to be causing any issues, therefore I am removing it.

marek-mihok:
> Is this tested to not break existing components with tooltips (regular or within ui.inline) across all supported browsers?

@mturoci, it is tested for every component in every browser and outside or within `ui.inline`. Also the implementation of this feature changes a tooltip wrapper styles only if `height='1'` is specified for the component, otherwise the styles are the same as before so there is nothing to break with this change. 

Also the only form component with a tooltip supporting specifying `height='1'` is `ui.table`.

mturoci (approved):
Great, thanks!

PR #2022, opened 2023-06-15
feat: Make ui.chatbot text input expand on newlines #1999

### LINKED ISSUE #1999: Make chatbot text input expand when newline is encountered instead of single line only

The current chatbot card input textbox accepts a single line text only, which is not enough.

The textbox should grow in height as needed (depending on the amount of newlines) and maxing out on a specified threshold. Same as other popular chatbots work.

marek-mihok:
Screen Recording 2023-06-15 at 10 12 38

Closes #1999

### DIFF
--- a/ui/src/chatbot.tsx
+++ b/ui/src/chatbot.tsx
@@ -21,7 +21,7 @@ import { clas, cssVar, getContrast, important, px, rem } from './theme'
 import { bond, wave } from './ui'
 
 const
-  SUBMIT_BTN_SIZE = 30,
+  INPUT_HEIGHT = 30,
   css = Fluent.mergeStyleSets({
     chatWindow: {
       height: '100%',
@@ -45,6 +45,7 @@ const
     msg: {
       maxWidth: '65ch',
       flexGrow: 1,
+      overflowWrap: 'break-word', // Breaks words longer than lines.
     },
     botMsg: {
       backgroundColor: cssVar('$neutralLighter'),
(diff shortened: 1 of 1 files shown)


ui/src/chatbot.tsx
```diff
@@ -57,12 +59,8 @@ const
...
     },
     stopButton: {
-      position: 'absolute',
-      bottom: 54,
-      left: 0,
-      right: 0,
-      marginLeft: 'auto',
-      marginRight: 'auto',
+      left: '50%',
+      transform: 'translate(-50%, -7px)',
```
mturoci, line 63: Why 7?

ui/src/chatbot.tsx
```diff
@@ -95,7 +93,12 @@ export const XChatbot = ({ model }: { model: Chatbot }) => {
       wave.args[model.name] = newVal
       setUserInput(newVal)
     },
-    onKeyDown = (e: React.KeyboardEvent<HTMLInputElement | HTMLTextAreaElement>) => { if (e.key === 'Enter') submit() },
+    onKeyDown = (e: React.KeyboardEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+      if (!e.shiftKey && e.key === 'Enter') {
+        e.preventDefault() // Prevents a newline from being added to the text box.
```
mturoci, line 98:
```suggestion
        e.preventDefault() // Do not add an extra newline to the input text box after submission.
```

mturoci (changes requested):
Thanks @marek-mihok. Looks good overall. Seems like the spacing got broken:

image

Also, let's add a side padding (left and right) as we do for other cards since the chatbot card has currently none.

marek-mihok:
@mturoci, thanks for the good points!

> Seems like the spacing got broken

Applying `white-space: 'pre-wrap'` on the paragraph component instead of its parent fixed it.

> Also, let's add a side padding (left and right)

I added margin instead since we want the right padding to take effect on scrollbar as well.

mturoci (approved):
LGTM

PR #2026, opened 2023-06-19
feat: Implement 'click' and 'tool_change' events for image annotator #2010

### LINKED ISSUE #2010: Image annotator - catch click events for single point coordinates of the image
#### Requested solution

To catch click events and to have a smooth experience with the ui.image annotator, we would like to have something like a point selector that reacts on click events and acts like a trigger. This way we can prompt models given an image point coordinate to obtain the polygon vertices which we would like to use to draw the new polygon shape with the ui.image_annotator.

It's also possible to prompt models of interest with a box but this would slow down the data labeling process and would not be as smart and efficient as using the point coordinate.

marek-mihok:
This change adds `event` prop for `ui.image_annotator` component with an option to specify the `click` and `tool_change` events.

The `click` event contains the `x` and `y` coordinates of the clicked point on the image canvas. A `tool_change` event contains the name of the selected tool (`rect`, `polygon` or `select`).

 Updated API is as follows:

```ts
export interface ImageAnnotator {
  name: Id
  /** The path or URL of the image to be presented for annotation. */
  image: S
  /** The image annotator's title. */
  title: S
  /** The master list of tags that can be used for annotations. */
  tags: ImageAnnotatorTag[]
  /** Annotations to display on the image, if any. */
  items?: ImageAnnotatorItem[]
  /** True if the form should be submitted as soon as an annotation is drawn. */
  trigger?: B
  /** The card’s image height. The actual image size is used by default. */
  image_height?: S
  /** List of allowed shapes. Available values are 'rect' and 'polygon'. If not set, all shapes are available by default. */
  allowed_shapes?: S[]
  /** The events to capture on this image annotator. One of `click` or `tool_change`. */
  events?: S[]
}
```

Closes #2010


website/widgets/form/image_annotator.md
```diff
@@ -86,3 +86,49 @@ The legend can be brought up by clicking the Info icon in the top right toolbar.
 |          **r**           |                     Select rectangle tool                     |
 |          **p**           |                       Select polygon tool                     |
 |          **s**           |                     Activate selection tool                   |
+
+## Click event
+
+Sometimes you may want to handle the annotating yourself and this is where `click` event comes into place. By setting `events=['click']` prop, the event is fired every time you click on the image canvas while using a `rect` or `polygon` tool and contains the `x` and `y` coordinates of the clicked point, e.g. `{x: 100, y: 120}`. The event can be accessed through `q.events.<image_annotator_name>.click`, where `image_annotator_name` is the `name` attribute of the image annotator component.
```
mturoci, line 92: No need to reexplain `events` here. This section should serve as a visual guide to what is possible and how it looks. Since events are "invisible", you can say handling custom click / toolchange events is possible.

Also let's try to use passive voice when possible instead of "Sometimes you want..."

ui/src/image_annotator.tsx
```diff
@@ -468,6 +474,13 @@ export const XImageAnnotator = ({ model }: { model: ImageAnnotator }) => {
         { cursor_x, cursor_y } = eventToCursor(e, rect, zoom, imgPositionRef.current),
         intersected = getIntersectedShape(drawnShapes, cursor_x, cursor_y)
 
+      if (model.events?.includes('click') && start?.x && start?.y) {
```
mturoci, line 477: We probably want to support picking `[0,0]` as well.

```suggestion
      if (start && model.events?.includes('click')) {
```

py/examples/image_annotator_events_click.py
```diff
@@ -0,0 +1,36 @@
...
+                items=[],
+                events=['click']
+            ),
+        ])
+        q.page['details'] = ui.markdown_card(
+            box='1 8 5 2',
+            title='Clicked point',
+            content='Nothing is clicked.',
+        )
+    else:
```
mturoci, line 31: Is this `else` necessary?

py/examples/image_annotator_events_click.py
```diff
@@ -0,0 +1,36 @@
...
+        if q.events.annotator:
+            if q.events.annotator.click:
```
mturoci, line 33:
```suggestion
        if q.events.annotator and q.events.annotator.click:
```

py/examples/image_annotator_smart.py
```diff
@@ -0,0 +1,80 @@
...
+                    ui.image_annotator_tag(name='v', label='Vehicle', color='$cyan'),
+                    ui.image_annotator_tag(name='a', label='Animal', color='$blue'),
+                ],
+                events=['click', 'tool_change'],
+                items=[],
+            ),
+            ui.button(name='submit', label='Submit', primary=True)
+        ])
+        q.client.initialized = True
+    elif q.args.submit:
```
mturoci, line 63: If should be enough I guess.

```suggestion
    if q.args.submit:
```

py/examples/image_annotator_smart.py
```diff
@@ -0,0 +1,80 @@
...
+        q.page['example'].items = [
+            ui.text(f'annotator={q.args.annotator}'),
+            ui.button(name='back', label='Back', primary=True),
+        ]
+    elif q.events.annotator:
+        if q.events.annotator.click:
```
mturoci, line 69:
```suggestion
    if q.events.annotator and q.events.annotator.click:
```

py/examples/image_annotator_smart.py
```diff
@@ -0,0 +1,80 @@
+# Form / Image Annotator / Events
+# Register the `click` and `tool_change` #events to emit Wave event when the image is clicked or a tool is changed.
+# #form #annotator #image #events
+# ---
+from h2o_wave import main, app, Q, ui
```
mturoci, line 5: Let's put this example into `events_click`. Put a text on top saying - "Try annotating the car and see what happens!".

py/examples/image_annotator_smart.py
```diff
@@ -0,0 +1,80 @@
...
+        if active_tool == 'polygon': return animal_annotation_polygon
+    else: return None
+
+
+@app('/demo')
+async def serve(q: Q):
+    if not q.client.initialized or q.args.back:
+        q.client.active_tool = 'rect'
+        q.page['example'] = ui.form_card(box='1 1 5 8', items=[
+            ui.image_annotator(
```
mturoci, line 48: Let's make the example simple and allow just a single tool - either rect or polygon, but not both.

mturoci (changes requested):
Thanks @marek-mihok! The code looks good, the examples need a bit of work. Feel free to ping me in case something is not clear.

mturoci (dismissed):

mturoci:
Thanks @marek-mihok, rebase the branch please and we are good to merge.

mturoci (approved):

PR #2027, opened 2023-06-19
fix: Show curly braces in pydocs correctly #1880

### LINKED ISSUE #1880: Special characters in docstrings are not translated correctly
For example, `@on('#foo/&lcub;fruit&rcub;')`  should be `@on('#foo/{fruit}')`:

```
A function annotated with @on('#foo/&lcub;fruit&rcub;') is invoked 
whenever q.args['#'] matches 'foo/apple', 'foo/orange', etc.
The parameter 'fruit' is passed to the function (in this case, 'apple', 'orange', etc.)
```

Source: https://wave.h2o.ai/docs/api/routing/#on-

Might be a pdoc3 template issue.

marek-mihok:
Pdoc builds the docs correctly when curly braces `{}` are used.
The problem is with [docusaurus using the MDX compiler](https://github.com/facebook/docusaurus/issues/3018) for `.md` files generated by pdoc while building the website.
In MDX, curly braces are reserved for expressions so in our case the MDX compiler is looking for a `fruit` variable which is, obviously, not defined - `@on('#foo/{fruit}')`. We can fix this by writing an expression inside the curly braces which returns what we want, in our case `{"{fruit}"}`.

Closes #1880


mturoci (approved):
Perfect, thanks @marek-mihok!

PR #2029, opened 2023-06-20
feat: Make theme generator responsive

marek-mihok:
https://github.com/h2oai/wave/assets/23740173/0094afcb-448d-47b7-8dd3-98273e6c3ba2

### DIFF
--- a/py/apps/theme-generator/requirements.txt
+++ b/py/apps/theme-generator/requirements.txt
@@ -1,2 +1,1 @@
 h2o_wave=={{VERSION}}
-pygments==2.11.2
--- a/py/apps/tour/requirements.txt
+++ b/py/apps/tour/requirements.txt
@@ -9,7 +9,6 @@ numpy==1.21.6
 opencv-python==4.5.5.64
 pandas==1.3.5
 plotly==5.7.0
-pygments==2.11.2
 scikit-learn==1.0.2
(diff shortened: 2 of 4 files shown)


mturoci:
Thanks @marek-mihok! Could we put the sidebar into a side_panel which would be opened by a button in the header when it no longer fits the screen? Secondly, the sample app should not have the height constrained, but take as much space on the mobile as possible.

marek-mihok:
> Thanks @marek-mihok! Could we put the sidebar into a side_panel which would be opened by a button in the header when it no longer fits the screen? Secondly, the sample app should not have the height constrained, but take as much space on the mobile as possible.

@mturoci is this what you imagine?

https://github.com/h2oai/wave/assets/23740173/f40381eb-02e2-46c9-b9eb-d2eb93671f1f

mturoci:
Yes, that's exactly what I wanted. One last thing: the sample app looks cramped on mobile so would be better to clean it up, e.g. table and plot definitely do not have enough space so put them under each other and so on.

marek-mihok:
> Yes, that's exactly what I wanted. One last thing: the sample app looks cramped on mobile so would be better to clean it up, e.g. table and plot definitely do not have enough space so put them under each other and so on.

@mturoci it's done, however duplicating of items was necessary.

PR #2031, opened 2023-06-22
feat: Make wave university responsive

marek-mihok:
https://github.com/h2oai/wave/assets/23740173/aebfac8d-d011-44e0-b259-3b27a9986a18


marek-mihok:
@mturoci I would personally keep the links inside the hamburger menu for all breakpoints as the content overlaps even on the bigger screens as can be seen in the screen recording above. WDYT?

mturoci:
My preference is to have them visible at all times, not needing an extra click. You can leave them as they are since I need to update them anyway, which might result in less space needed.

mturoci (approved):
LGTM! Thanks @marek-mihok.

PR #2034, opened 2023-06-22
fix: Set active tool other than 'select' to emit the click event

marek-mihok:
This PR fixes the failing ui.image_annotator test.


mturoci (approved):
Thanks @marek-mihok!

PR #2035, opened 2023-06-23
fix: Introduce request retry mechanism to Waved to retry failed requests

senalw:
**Issue:** Currently Waved disconnects app when a request failure occurs between Waved and app. Therefore, this PR will introduce a request retry mechanism to retry failed requests for a certain amount of time period which can be configured by a user. If this retry count (`H2O_WAVE_MAX_REQUEST_RETRY_COUNT`) is not configured, it wont enable request retry mechanism. 

**Testing Conditions,** 
`H2O_WAVE_MAX_REQUEST_RETRY_COUNT=20` and `H2O_WAVE_REQUEST_RETRY_INTERVAL=1s`

**Testing Records,**

1. **Simulate `read: connection reset by peer` disconnection for 2 seconds and reset back to normal with deleting iptables rule after 2 seconds.**

`sleep 5 && iptables -A INPUT -p tcp --dport 8000 -j REJECT --reject-with tcp-reset && sleep 2 && iptables -D INPUT -p tcp --dport 8000 -j REJECT --reject-with tcp-reset`

Result: 
App was able to successfully recover after 2 seconds, 

```
2023/06/23 13:59:21 # {"error":"request failed: Post \"http://127.0.0.1:8000\": dial tcp 127.0.0.1:8000: connect: connection refused","host":"http://127.0.0.1:8000","route":"/","t":"app"}
2023/06/23 13:59:21 # {"addr":"172.17.0.1:58704","route":"/4f4b7de4-1f62-4326-b37b-855454c4c61a","t":"ui_add"}
2023/06/23 13:59:21 # {"error":"request failed: Post \"http://127.0.0.1:8000\": dial tcp 127.0.0.1:8000: connect: connection refused","host":"http://127.0.0.1:8000","max retry count":"20","retry count":"1","retry interval":"1s","route":"/","t":"app"}
2023/06/23 13:59:21 # {"error":"request failed: Post \"http://127.0.0.1:8000\": dial tcp 127.0.0.1:8000: connect: connection refused","host":"http://127.0.0.1:8000","max retry count":"20","retry count":"2","retry interval":"1s","route":"/","t":"app"}
INFO:     127.0.0.1:49624 - "POST / HTTP/1.1" 200 OK
```

https://github.com/h2oai/wave/assets/12801761/d0e50281-fe8b-4136-99b8-15a68cd7afc8

[...]

PR #2050, opened 2023-06-29
feat: Allow keeping unresponsive apps.

### LINKED ISSUE #2043: Wave app being dropped when there's a request failure between Waved and the app
#### Wave SDK Version, OS

Wave SDK: 0.25.3 , OS: MacOs/ Ubuntu

#### Actual behavior

Wave daemon drops Wave apps when there is a http request failure between Waved and apps without even retrying it for a configurable time.  This request failure could occur when there's a resource starvation for a small amount of time period. Therefore, dropping app is not acceptable here. 

#### Expected behavior

Wave daemon should have a request-retry mechanism with user configurable time period. By default this retry-mechanism is turned off and we can enable it by setting env variable. 

#### Steps To Reproduce

1. Start the Wave app. 
2. Simulate sending RST signal for the connection between Waved and the app. 
`iptables -A INPUT -p tcp --dport 8000 -j REJECT --reject-with tcp-reset`
3. With the current behaviour, wave app will be dropped with `connection reset by peer` or `connection refused` error.

mturoci:
Implementation of https://github.com/h2oai/wave/pull/2035#issuecomment-1611239620.

Closes #2043


lo5 (changes requested):
Maybe can change the env var to something shorter, e.g. `H2O_WAVE_KEEP_APP_ALIVE`, default to false. 

Other than that, looks good - thanks @mturoci!

PR #2070, opened 2023-07-11
feat: Add H2O_WAVE_RELOAD_EXCLUDE to exclude files from being watched during app reload.

mturoci:


pascal-pfeiffer (commented):
Thanks a lot for quickly addressing this @mturoci .

I haven't tested the new feature, but from the feature description it is exactly what I had in mind. Thanks again!

PR #2082, opened 2023-07-19
fix: Endless loop when table value dependent on 'select' event  #2069

### LINKED ISSUE #2069: Passing value to table with 'select' event leads to trigger-loop
#### Wave SDK Version, OS

h2o-wave 0.26.1, Windows 10

#### Actual behavior

If a table has `single=True` and `events=['select']`, passing the selected value to the value argument (e.g., `value=q.args.table[0]`) causes the page to enter a loop of infinite refresh.

- The user selects a row
- The page refreshes and passes the selection to the table
- This triggers the select event and refreshes the page
- Value is passed to table........

#### Expected behavior

Passing the current row value only fills the checkbox instead of triggering the `select` event (e.g., same behaviour as when there is no event assigned). The event should be triggered on mouse-click, not on having a default/past value passed.

My use case is: A table provides an overview of data. Clicking a row adds a new tile showing the complete data in case the user wants to see more details. For extra clarity, I'd like the table row to stay selected, but of course it works just as fine when the row is not selected (ergo, I just skip passing anything to `value`). I still thought I raised this issue in case this is indeed NOT the behaviour you wanted for the table.

#### Steps To Reproduce

1. Run code below
2. Select a row
3. Witness infinite refreshs

```python
from h2o_wave import main, app, Q, ui, on, data


async def table_page(q: Q):
    items = [ui.table(
            name='table',
            columns=[ui.table_column(name='text', label='Table select event')],
            rows=[
                ui.table_row(name='row1', cells=['Row 1']),
                ui.table_row(name='row2', cells=['Row 2']),
                ui.table_row(name='row3', cells=['Row 3'])
            ],
            single=True,
            events=['select'],
            value=q.args.table[0] if q.args.table else None
        )]

    if q.args.table:
        # Do something with the selected row, e.g., show more details 
        items.append(ui.text_xl(f"{q.args.table[0]} selected"))

    q.page['example'] = ui.form_card(box='1 1 3 4', items=items)


@app('/')
async def serve(q: Q):
    await table_page(q)
    await q.page.save()
```

marek-mihok:

___

This PR prevents the endless loop by triggering the `select` event only if the selection was done by mouse/keyboard (or to be exact by not triggering the `select` event if the value is passed to the table programatically).

Closes #2069


mturoci:
Thanks @marek-mihok. Do we have any alternative solutions? Also please add accompanying unit tests.

marek-mihok:
> Do we have any alternative solutions? 

After exhausting attempts I was not able find the reasonable & working solution.

- I was trying to set default selected values while creating new `Selection` object instead of the useEffect but without the result since FluentUI does not support other way of making selection other than using functions `setKeySelected`, `setIndexSelected` or `setAllSelected`.
- The possible solution is also to emit the `select` event only on mouse click, but this includes implementing the custom behavior for both checkbox and row mouse clicks.

Do you have any ideas @mturoci ?

> Also please add accompanying unit tests.

Done ✅

marek-mihok:
Thanks @mturoci! I went through your solution and it is looking good. Also I tested manually multiple scenarios and all is working as expected.

PR #2085, opened 2023-07-20
feat: Add data buffer support to components (form card). #441

### LINKED ISSUE #441: Avoid having to specify pack=True when using data() for components
## Question and Background
It seems to the developer that 1. `ui.form_card` with `ui.visualization` would be the same as 2. `ui.plot_card`. This is not explicitly stated in the docs, but from looking at examples it seems like 1. is the only way to replicate 2. in a card with other items, so the user assumes you could copy-past your plot and data code from one and put it in the other. This is not the case when using a dataframe, and addition option `pack=True` is required in 1. 

## Reproducible Example
```
n = 100
    df = pd.DataFrame({'length': np.random.rand(n),
                       'width': np.random.rand(n),
                       'data_type': np.random.choice(a=['Train', 'Test'], size=n, p=[0.8, 0.2])
                       })

    plot_marks = [ui.mark(type='point', x='=length', x_title='Length (cm)', y='=width', y_title='Width (cm)',
                          color='=data_type', shape='circle')]

    q.page['scatter_plot_card'] = ui.plot_card(
        box='1 1 4 4',
        title='Scatter Plot from Dataframe',
        data=data(
            fields=df.columns.tolist(),
            rows=df.values.tolist()
        ),
        plot=ui.plot(marks=plot_marks)
    )

    q.page['scatter_viz_form_card'] = ui.form_card(
        box='1 5 4 4',
        items=[ui.visualization(
                plot=ui.plot(marks=plot_marks
                             ),
                data=data(
                    fields=df.columns.tolist(),
                    rows=df.values.tolist(),
                    pack=True
                ),
        )]
    )
```
Notice the `pack=True` of `scatter_viz_form_card` which is not required for `scatter_plot_card`. If you do not include it you get this error:
> TypeError: Object of type Data is not JSON serializable

## Actual Question 
You can add `pack=True` to `scatter_plot_card` and nothing changes, but I am wondering why it's not required. Is this user confusion, a possible bug in `ui.visualization`, or missing docs?

mturoci:
This PR introduces support for non-packed data buffers within `ui.visualization` components - form cards.

By-name access is supported as well, buffers behave in the same way for form components as they do in cards.

```py
import time

from h2o_wave import site, ui, data

page = site['/demo']

page.add('example', ui.form_card(
    box='1 1 -1 8',
    items=[
        ui.visualization(
            name='my_plot',
            plot=ui.plot([ui.mark(type='interval', x='=profession', y='=salary', y_min=0)]),
            data=data(fields='profession salary', rows=[
                ('medicine', 23000),
                ('fire fighting', 18000),
                ('pedagogy', 24000),
                ('psychology', 22500),
                ('computer science', 36000),
            ]),
        ),
        ui.visualization(
            plot=ui.plot([ui.mark(type='interval', x='=profession', y='=salary', y_min=0)]),
            data=data(fields='profession salary', rows=[
                ('medicine', 21000),
                ('fire fighting', 17000),
                ('pedagogy', 23500),
                ('psychology', 22300),
                ('computer science', 33000),
            ])
        ),
    ],
))

page.save()

time.sleep(5)

page['example'].my_plot.data[0] = ('medicine', 1000)
page.save()

```

Closes #441

### DIFF
--- a/card.go
+++ b/card.go
@@ -285,4 +285,6 @@ func fillFormCardBufs(data map[string]any, outData map[string]any, bufs *[]BufD,
 			*bufs = append(*bufs, b.dump())
 		}
 	}
+	//lint:ignore SA4006 this function is recursive so mutation is necessary.
+	keys = keys[:len(keys)-1]
 }
--- a/card.go
+++ b/card.go
@@ -97,7 +97,6 @@ func (c *Card) set(ks []string, v any) {
 		}
 
 		keys := strings.Split(p, ".")
-		lastKey := keys[len(keys)-1]
 		var x any = c.data
 		// Get the object even if nested.
 		for i := 0; i < len(keys)-1; i++ {
@@ -105,6 +104,7 @@ func (c *Card) set(ks []string, v any) {
 		}
 
 		if x, ok := x.(map[string]any); ok {
+			lastKey := keys[len(keys)-1]
 			if v == nil {
 				delete(x, lastKey)
 			} else {
@@ -120,8 +120,8 @@ func (c *Card) set(ks []string, v any) {
 			x = c.nameComponentMap[ks[0]]
 			startIdx = 1
 		}
-		for i := startIdx; i < len(ks)-1; i++ {
-			x = get(x, ks[i])
+		for _, k := range ks[startIdx : len(ks)-1] {
+			x = get(x, k)
 		}
 
 		p := ks[len(ks)-1]
--- a/card.go
+++ b/card.go
@@ -54,6 +54,12 @@ type Card struct {
 	nameComponentMap map[string]any // Cache for cards with items, secondary_items or buttons.
 }
 
+// Tmp placeholder for key-buffer pairs.
+type BufEntry struct {
+	k   string
+	buf interface{}
+}
+
 const dataPrefix = "~"
 
 func loadCard(ns *Namespace, c CardD) *Card {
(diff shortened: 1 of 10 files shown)


mturoci:
@geomodular could you please have a look at the Go code once you have a bit of time? I feel like the code is too verbose and a more idiomatic solution exists.

card.go
```diff
@@ -94,22 +95,33 @@ func (c *Card) set(ks []string, v any) {
...
-		if v == nil {
-			delete(c.data, p)
-		} else {
-			c.data[p] = v
+
+		keys := strings.Split(p, ".")
+		lastKey := keys[len(keys)-1]
+		var x any = c.data
+		// Get the object even if nested.
+		for i := 0; i < len(keys)-1; i++ {
```
geomodular, line 103: I'd rewrite this as `for ... range` for better readability:

```go
for _, key := range keys {
  x = get(x, key)
}
```

card.go
```diff
@@ -94,22 +95,33 @@ func (c *Card) set(ks []string, v any) {
...
-			for _, k := range ks[:len(ks)-1] {
-				x = get(x, k)
-			}
+		startIdx := 0
+		if _, ok := c.data[ks[0]]; !ok && c.nameComponentMap[ks[0]] != nil {
+			// By-name access.
+			x = c.nameComponentMap[ks[0]]
+			startIdx = 1
+		}
+		for i := startIdx; i < len(ks)-1; i++ {
```
geomodular, line 123: This can be rewritten as:

```go
for _, k := range ks[startIdx:] {
  x = get(x, k)
}
```

card.go
```diff
@@ -163,6 +175,12 @@ func get(ix any, k string) any {
 func (c *Card) dump() CardD {
 	data := make(map[string]any)
 	var bufs []BufD
+
+	// Look for nested buffers within form_card.
+	if v, ok := c.data["view"]; ok && v.(string) == "form" {
```
geomodular, line 180: `v.(string)` might not be a string and can cause some trouble here. I'd handle it inside the `if` scope.

card.go
```diff
@@ -163,6 +175,12 @@ func get(ix any, k string) any {
...
+		fillFormCardBufs(c.data, data, &bufs, make([]string, 0))
```
geomodular, line 181: At first I was puzzled with `make([]string, 0))` but now I see the function is recursive. Seems ok.

geomodular (dismissed):
Looks good for the `go` part. Just few notes.

card.go
mturoci: Sure, the problem is I need to ignore the last entry. Do you think this would be more readable?

```go
for _, key := range keys[:len(keys)-1] {
  x = get(x, key)
}
```

card.go
mturoci: Perfect, exactly what I needed 👍 

```go
for _, k := range ks[startIdx : len(ks)-1] {
	x = get(x, k)
}
```

card.go
mturoci: I guess I need to write a nested if with type cast check, right? Or is there a way how to put it to the existing if statement?

geomodular: yes, nested if

Issue #1154, opened 2021-12-14
Unify value behavior


mturoci:
via @psinger 

## Repro

```py

from h2o_wave import main, app, Q, ui


@app("/demo")
async def serve(q: Q):
    if not q.client.initialized:
        items = [
            ui.slider(name="test", label="test", value=0.5, min=0, max=1, step=0.1),
            ui.dropdown(name="test2", label="test2", value="1", choices=[ui.choice("1", "1"), ui.choice("2", "2"), ui.choice("3", "3")]),
            ui.button(name='submit', label='Submit', primary=True),
        ]
        q.page['example'] = ui.form_card(box='1 1 4 4', items=items)
        q.client.initialized = True
    else:
        items=[
            ui.slider(name="test", label="test", value=0.3, min=0, max=1, step=0.1),
            ui.dropdown(name="test2", label="test3", value="2", choices=[ui.choice("1", "1"), ui.choice("2", "2"), ui.choice("3", "3")]),
            ui.button(name='show_inputs', label='Submit', primary=True),
        ]
        q.page['example'].items = items
    await q.page.save()

```

## Actual behavior

After changing dropdown value and hitting submit, dropdown value should respect user input, but is currently overwritten by the `value` attribute in the else block (2).

## Desired behavior

`value` should be *initial value* only and dropdown should then be fully controlled by the user.

aalencar:
I couldn't find a solution yet, but I found out why dropdown doesn't hold its value when submit is clicked.

@mturoci gave me some leads. `form` card and `bond` function might have something to do with it. After much examination, I didn't find anything there. I removed `bond` , turned `form` into a class component, and tinkered with `componentDidUpdate`, but the problem persisted.

Instead of checking why the dropdown was changing, I decided to investigate why the `Slider` component was not. Turns out that it was, but Fluent `Slider`'s internal logic handles updates differently, not changing its value back to the default when `props` change. Also, contrarily to what was previously assumed, components are not being recreated when submit is clicked.

Based on that, I had to look into how Fluent's Dropdown logic works and the snippet bellow can help us understand what is going on.

```typescript
  // fluentui/packages/office-ui-fabric-react/src/components/Dropdown/Dropdown.base.tsx

  public UNSAFE_componentWillReceiveProps(newProps: IDropdownProps): void {
    // In controlled component usage where selectedKey is provided, update the selectedIndex
    // state if the key or options change.
[...]

PR #2093, opened 2023-07-24
fix: Allow dynamic value change for spinbox #1154

marek-mihok:

___

This PR partially resolves issue #1154.

NOTE: Do not forget to update status in [this comment](https://github.com/h2oai/wave/issues/1154#issuecomment-1252010473) once merged.

### DIFF
--- a/ui/src/spinbox.tsx
+++ b/ui/src/spinbox.tsx
@@ -65,8 +65,9 @@ const
     return -groups[1]?.length || groups[2]?.length || 0
   }
 export const
-  XSpinbox = ({ model: { name, trigger, label, disabled, min = 0, max = 100, step = 1, value = 0 } }: { model: Spinbox }) => {
+  XSpinbox = ({ model: m }: { model: Spinbox }) => {
     const
+      { name, trigger, label, disabled, min = 0, max = 100, step = 1, value = 0 } = m,
       [val, setVal] = React.useState<S>(String(value)),
       precision = Math.max(calculatePrecision(step), 0),
       parseValue = React.useCallback((v: F) => {
@@ -78,6 +79,7 @@ export const
         wave.args[name] = newValue
         if (trigger) wave.push()
         setVal(String(newValue))
+        m.value = newValue
         return String(newValue)
       },
       onDecrement = () => {
@@ -85,6 +87,7 @@ export const
         wave.args[name] = newValue
         if (trigger) wave.push()
         setVal(String(newValue))
+        m.value = newValue
         return String(newValue)
       },
       handleValue = React.useCallback((val: S) => {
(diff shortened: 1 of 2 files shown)


ui/src/spinbox.tsx
```diff
@@ -108,16 +111,23 @@ export const
...
 
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    React.useEffect(() => { wave.args[name] = (value < min) ? min : ((value > max) ? max : value) }, [])
+    React.useEffect(() => {
+      if (value === wave.args[name]) return
+      const val = (value < min) ? min : ((value > max) ? max : value)
+      wave.args[name] = val
+      setVal(String(val))
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [value])
```
mturoci, line 130: We probably want to include all the necessary dependencies than just `value`, right? This way the `value` would be always in bounds of minmax.

If so, please add a quick unit test validating the value is changed properly on minmax update. Should be pretty fast.

ui/src/spinbox.tsx
```diff
@@ -108,16 +111,23 @@ export const
       }, [max, min, parseValue, precision]),
       handleOnInput = React.useCallback((val: U) => {
         wave.args[name] = val
+        m.value = val
         if (trigger) wave.push()
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
mturoci, line 116: Any reason for not including `m` as a dependency?

mturoci (changes requested):
Thanks @marek-mihok. A few questions. Looks good otherwise.

ui/src/spinbox.tsx
marek-mihok: No specific reason, I just don't find it necessary. I have no problem with including it.

ui/src/spinbox.tsx
marek-mihok: Do you mean if the value is updated through wave app, it can also be outside of the min max boundaries?

ui/src/spinbox.tsx
```diff
@@ -108,16 +111,23 @@ export const
...
       debouncedHandleOnInput = React.useRef(wave.debounce(DEBOUNCE_TIMEOUT, handleOnInput)),
       onInput = (e: React.SyntheticEvent<HTMLElement>) => {
         const numVal = handleValue((e.target as HTMLInputElement).value)
         trigger ? debouncedHandleOnInput.current(numVal) : handleOnInput(numVal)
       }
 
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    React.useEffect(() => { wave.args[name] = (value < min) ? min : ((value > max) ? max : value) }, [])
+    React.useEffect(() => {
+      if (value === wave.args[name]) return
```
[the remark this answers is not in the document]
marek-mihok: To prevent from precision being ignored (see the failing test case when the condition is removed). However I've just found out, this solution is not ideal since it causes other problems when using numbers with decimal point. I'm looking for another solution, will keep you updated.

ui/src/spinbox.tsx
mturoci: If the min/max is updated and the value is suddenly out of bounds, it should be adjusted, right? Also the above suffers from stale closure problem as min/max can be stale within useEffect during value change.

Or is there any reason for including just `value`  as the only dependency?

marek-mihok: Makes sense, thanks for the explanation. No, there is no specific reason.

mturoci (approved):
LGTM.

PR #2118, opened 2023-08-21
fix: Make dynamic value change consistent

### LINKED ISSUE #1154: Unify value behavior

### LINKED ISSUE #1776: Allow for no selected tab in tabs group

My application is starting to have too many tabs. At the same time tabs are nice to switch back and forth. My first approach was to add a menu.
Thanks to v 0.24.0 adding labels to menus this even looks quite smooth.
But now, when I use a link from the menu I get the following situation where the tab marking doesn't match the rest
image

I found old threads about similar discussion (two sets of tabs) and understand that there is no complete solution.
My suggestion would be that if you set for ui.tabs value=None then none of the tabs is highlighted.
This would probably be practical for a few scenarios (e.g. you have a Welcome page that is not part of your tabs).

marek-mihok:
**The PR fulfills these requirements:** (check all that apply)

___

This PR addresses all issues found after final revision of #1154. Included issues are:

- **Checklist** - set values to `None`, select value manually, set values to `None` once again (unable to set to `None`)
- **Tabs** - dynamic value change is fully supported for `ui.tab_card` only, but not for `ui.tabs` - try to set value dynamically, select tab manually, set value dynamically to the same value as before (Tab is not switching)
- **Combobox single/multi** - when `choices` are set to `None` it leads to endless loop
- **Dropdown - dialog single/multi** - when `choices` are set to `None` it leads to endless loop
- **Nav** - user was unable to dynamically unselect all items by specifying `value=None` within the Wave app when component was initially rendered with some value

Also this PR adjusts the behavior of Nav/Tab/Tabs in following manner:
- Revert behavior for Menu - when value is set to `None`, no item is active
- Keep behavior for Tab/Tabs - when value is set to `None`, first item is active by default - we are not bringing support for #1776 
- Args setting is consistent

Closes #1154 
Closes #1776

### DIFF
--- a/ui/src/checklist.test.tsx
+++ b/ui/src/checklist.test.tsx
@@ -70,13 +70,13 @@ describe('Checklist.tsx', () => {
     const { getByText, rerender } = render(<XChecklist model={{ ...checklistProps, values: ['Choice1'] }} />)
     expect(wave.args[name]).toMatchObject(['Choice1'])
 
-    rerender(<XChecklist model={{ ...checklistProps, choices: [...choices] }} />)
+    rerender(<XChecklist model={{ ...checklistProps }} />)
     expect(wave.args[name]).toMatchObject([])
 
     fireEvent.click(getByText('Choice2').parentElement!)
     expect(wave.args[name]).toMatchObject(['Choice2'])
 
-    rerender(<XChecklist model={{ ...checklistProps, choices: [...choices] }} />)
+    rerender(<XChecklist model={{ ...checklistProps }} />)
     expect(wave.args[name]).toMatchObject([])
   })
 
--- a/ui/src/checklist.test.tsx
+++ b/ui/src/checklist.test.tsx
@@ -65,6 +65,20 @@ describe('Checklist.tsx', () => {
     expect(wave.args[name]).toMatchObject(['Choice1'])
   })
 
+  it('Set correct args when value is cleared twice', () => {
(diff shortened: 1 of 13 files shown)


mturoci (approved):
Thanks @marek-mihok!

PR #2120, opened 2023-08-28
feat: Add copy to clipboard button to markdown code block #2025

### LINKED ISSUE #2025: Add copy to clipboard button when hovered over markdown code block.
Extract the clipboard button component from our existing `copyable_text` widget and reuse it within markdown. The markdown copy button should only be visible on hover over the code block.

marek-mihok:

___

 This PR reuses the button from ui.copyable_text also for markdown code block.

https://github.com/h2oai/wave/assets/23740173/5153862d-e10c-4b7f-8cec-07cbd544fbb6

Closes #2025

### DIFF
--- a/ui/src/copyable_text.tsx
+++ b/ui/src/copyable_text.tsx
@@ -4,19 +4,18 @@ import React from 'react'
 import { stylesheet } from 'typestyle'
 import { clas, cssVar, pc } from './theme'
 
-
 const
   css = stylesheet({
     btn: {
       position: 'absolute',
       minWidth: 'initial',
-      width: 34,
+      width: 24,
       height: 24,
       right: 0,
       transform: 'translate(-4px, 4px)',
       outlineWidth: 1,
       outlineStyle: 'solid',
-      outlineColor: cssVar('$white'),
+      outlineColor: cssVar('$text'),
       zIndex: 1,
     },
     copiedBtn: {
@@ -24,14 +23,6 @@ const
       $nest: {
         '&:hover': {
           background: cssVar('$green'),
-        }
-      }
-    },
-    animate: {
-      $nest: {
-        'button': {
-          opacity: 0,
-          transition: 'opacity .5s',
         }
       }
     },
@@ -86,12 +77,17 @@ export const ClipboardCopyButton = ({ value }: { value: S }) => {
 
   React.useEffect(() => () => window.clearTimeout(timeoutRef.current), [])
 
-  return <Fluent.PrimaryButton
-    title='Copy to clipboard'
-    onClick={onClick}
-    iconProps={{ iconName: copied ? 'CheckMark' : 'Copy' }}
-    className={clas(css.btn, copied ? css.copiedBtn : '')}
-  />
+  return (
+    <Fluent.PrimaryButton
+      title='Copy to clipboard'
+      onClick={onClick}
+      iconProps={{
+        iconName: copied ? 'CheckMark' : 'Copy',
+        styles: { root: { marginTop: -1.35 } } // Nudge up the icon a bit to account for incorrect Fluent crop.
+      }}
+      className={clas(css.btn, copied ? css.copiedBtn : '')}
+    />
+  )
 }
 
 export const XCopyableText = ({ model }: { model: CopyableText }) => {
@@ -107,16 +103,14 @@ export const XCopyableText = ({ model }: { model: CopyableText }) => {
       onRenderLabel={() =>
         <div className={css.labelContainer}>
           <Fluent.Label>{label}</Fluent.Label>
-          <span className={multiline ? clas(css.animate) : ''}>
-            <ClipboardCopyButton value={value} />
-          </span>
+          <ClipboardCopyButton value={value} />
         </div>
       }
       styles={{
         root: {
           ...heightStyle,
           textFieldRoot: { position: 'relative', width: pc(100) },
-          textFieldMultiline: multiline ? { '&:hover button': { opacity: 1 } } : undefined
+          textFieldMultiline: multiline ? { '& button': { opacity: 0 }, '&:hover button': { opacity: 1 } } : undefined
         },
         wrapper: heightStyle,
         fieldGroup: heightStyle || { minHeight: height },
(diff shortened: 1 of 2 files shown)


mturoci (changes requested):
Thanks @marek-mihok! 

* Using portals seems like overkill. Can we do without?
* The button on dark md syntax highlight blends into background and is barely visible. Can this be improved?
* The tests that check if button is rendered are unnecessary, we have visual regression tests for that. Hover ones are ok.

marek-mihok:
> Using portals seems like overkill. Can we do without?

We can, please check the updated solution. I've reused the original approach we used for copyable_text and applied the same principle also for markdown code block.

> The button on dark md syntax highlight blends into background and is barely visible. Can this be improved?

Improved by adding the outline.

> The tests that check if button is rendered are unnecessary, we have visual regression tests for that. Hover ones are ok.

Good point. Removed ✅

ui/src/copyable_text.tsx
```diff
@@ -54,19 +64,25 @@ export interface CopyableText {
...
   const
-    { name, multiline, label, value, height } = model,
-    heightStyle = multiline && height === '1' ? fullHeightStyle : undefined,
-    ref = React.useRef<Fluent.ITextField>(null),
     timeoutRef = React.useRef<U>(),
     [copied, setCopied] = React.useState(false),
-    onClick = async () => {
-      const el = ref.current
-      if (!el) return
+    [visible, setVisible] = React.useState(!showOnHoverOnly),
```
mturoci, line 80: Unnecessary visibility state. Use CSS instead.

ui/src/copyable_text.tsx
```diff
@@ -54,19 +64,25 @@ export interface CopyableText {
...
+    onClick = React.useCallback(async () => {
+      if (!anchorElement) return
```
mturoci, line 82: Can the copy to clipboard be done without `anchorElement`?

ui/src/copyable_text.tsx
```diff
@@ -76,32 +92,56 @@ export const XCopyableText = ({ model }: { model: CopyableText }) => {
...
+
+export const XCopyableText = ({ model }: { model: CopyableText }) => {
+  const
+    { name, multiline, label, value, height } = model,
+    heightStyle = multiline && height === '1' ? fullHeightStyle : undefined,
+    [inputEl, setInputEl] = React.useState(),
+    domRef = React.useCallback(node => {
+      const inputEl = node?.children[0]?.children[1]
+      if (inputEl) setInputEl(inputEl)
+    }, [])
```
mturoci, line 126: Seems unnecessary.

ui/src/markdown.tsx
```diff
@@ -105,7 +107,18 @@ export const Markdown = ({ source }: { source: S }) => {
...
+          const codeBlockContainer = document.getElementById(codeBlockId)?.parentElement
+          if (codeBlockContainer) {
+            const buttonContainer = document.createElement('span')
+            ReactDOM.render(<ClipboardCopyButton value={str} anchorElement={codeBlockContainer} showOnHoverOnly />, buttonContainer,
+              () => {
+                codeBlockContainer.style.position = 'relative'
+                codeBlockContainer.insertAdjacentElement('afterbegin', buttonContainer)
+              })
+          }
+        }), 0)
```
mturoci, line 121: * Do not mix async/await with promise callbacks.
* Why is finally needed?
* Why is ReactDOM.render callback needed?
* Can it be part of `highlightSyntax` function instead? That way, the codeBlockContainer would be available.
* Why is `insertAdjacentElement('afterbegin'` needed as opposed to an `appendChild`? I thought button is absolutely positioned so its insertion order does not matter.

ui/src/copyable_text.tsx
```diff
@@ -4,19 +4,29 @@ import React from 'react'
 import { stylesheet } from 'typestyle'
 import { clas, cssVar, pc } from './theme'
 
+const
+  BUTTON_HEIGHT = 24,
+  BUTTON_WIDTH = 34
```
mturoci, line 9: This is unnecessary as its only used once anyway and it's clear what it is.

mturoci (changes requested):

ui/src/copyable_text.tsx
marek-mihok: You're right, it was useful during my previous implementations.

ui/src/copyable_text.tsx
marek-mihok: In order for _copy exec command_ to work we need to call `select()` on input/textarea element.
```tsx
if (document.queryCommandSupported('copy')) {
          (anchorElement as any)?.select()
          document.execCommand('copy')
          window.getSelection()?.removeAllRanges()
        }
```
Therefore we need to have its reference inside ClipboardCopyButton component. 

If we don't want to use `anchorElement` prop, the option would be to obtain its reference internally inside the ClipboardCopyButton component. But how? The general solution won't work (e.g. getting the parent of ClipboardCopyButton) because
- for `copyable_text` the button is attached to label, so textarea/input is the sibling of the button container 
- for markdown code block the copy button is a child of the code block

In this case what comes to my mind is to make the ClipboardCopyButton component a wrapper and the input/textarea/code will always be a child to ClipboardCopyButton component.
```tsx
<ClipboardCopyButtonWrapper>
   [markdown_code/copyable_text_input]
</ClipboardCopyButtonWrapper>
```

The problem with this approach is that FluentUI does not provide the option to override the input element for multiline input of `Fluent.TextField` so we couldn't wrap it inside a ClipboardCopyButtonWrapper.

mturoci: > In order for copy exec command to work we need to call select() on input/textarea element.

Correct, but why jump through hoops when the element can simply be created on the fly, its value set, copied and then removed? Since this is a common problem, a quick prompt to copilot returns:

```js
function copyToClipboard(str) {
  const el = document.createElement('textarea');
  el.value = str;
  document.body.appendChild(el);
  el.select();
  document.execCommand('copy');
  document.body.removeChild(el);
}
```

marek-mihok: What a nice trick! Thanks @mturoci .

mturoci (approved):
Thanks @marek-mihok!

I fixed the icon centering + removed some more unnecessary code.

PR #2137, opened 2023-09-13
fix: Apply primary button styles for icon button #2121

### LINKED ISSUE #2121: Primary button UI without label
#### Wave SDK Version, OS
0.26.0 on MacOS

#### Actual behaviour
Button is not primary if label is not set.

Screenshot 2023-08-28 at 6 42 48 PM

eg: Button1 isn't primary.

#### Expected behaviour
Button1 should be primary.

#### Steps To Reproduce
```python


@app('/demo')
async def serve(q: Q):
        box='1 1 4 10',
        items=[
            ui.button(name='button1', icon='Search', primary=True),
            ui.button(name='button2', label='Search', icon='Search', primary=True)
        ])

    await q.page.save()
```

marek-mihok:

___

image

Closes #2121

### DIFF
--- a/ui/src/button.tsx
+++ b/ui/src/button.tsx
@@ -115,11 +115,6 @@ const
     center: 'center',
     between: 'space-between',
     around: 'space-around',
-  },
-  iconButtonPrimaryStyles = {
-    root: { color: cssVar('$white'), background: cssVar('$themePrimary') },
-    rootHovered: { color: cssVar('$white'), background: cssVar('$themeDarkAlt') },
-    rootPressed: { color: cssVar('$white'), background: cssVar('$themePrimary') }
   }
 
 const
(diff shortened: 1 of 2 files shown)


mturoci (changes requested):
Thanks @marek-mihok.

* What's the reason for not using the existing `Fluent.PrimaryButton` instead of mimicking it with custom styles?
* The colors doesn't look good on other themes.
* Make this part of our visual regression testing by including the primary icon button within `md` docs.

marek-mihok:
> What's the reason for not using the existing Fluent.PrimaryButton instead of mimicking it with custom styles?

@mturoci the primary button with icon only is much wider than icon button (it has horizontal padding)

mturoci:
I don't understand. That's the point of the issue, isn't it? If `primary=True`, render a primary button. The padding is specific to primary button styles which is desired - so that people understand it's primary.

mturoci (approved):

PR #2146, opened 2023-09-29
feat: removed reserved space for empty titles in vega_card #1923

### LINKED ISSUE #1923: Do not render reserved title whitespace in vega_card if the `title` is empty
Even though the `title` is empty, in the vega_card there is still a whitespace rendered, which is just increasing unused space. This may be true for other cards as well.

Do not render reserved title whitespace if the `title` is empty.

MattiasOz:

Closes #1923

### DIFF
--- a/ui/src/vega.tsx
+++ b/ui/src/vega.tsx
@@ -147,10 +147,15 @@ export const
         const { specification, data, title, grammar = 'vega-lite' } = state
         return (
           <div data-test={name} className={css.card}>
-            <div className='wave-s12 wave-w6'>{title}</div>
+            {title && <div className='wave-s12 wave-w6'>{title}</div>}
+            {title && 
             <div className={css.body}>
               <XVegaVisualization model={{ specification, data, grammar }} />
-            </div>
+            </div>}
+            {!title && 
+            <div className={css.body} style={{marginTop: 0}}>
+              <XVegaVisualization model={{ specification, data, grammar }} />
+            </div>}
           </div>
         )
       }
--- a/ui/src/vega.tsx
+++ b/ui/src/vega.tsx
@@ -146,7 +146,7 @@ export const
         const { specification, data, title, grammar = 'vega-lite' } = state
         return (
           <div data-test={name} className={css.card}>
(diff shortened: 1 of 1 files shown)


ui/src/vega.tsx
```diff
@@ -147,10 +147,15 @@ export const
...
+            {title && <div className='wave-s12 wave-w6'>{title}</div>}
+            {title && 
             <div className={css.body}>
               <XVegaVisualization model={{ specification, data, grammar }} />
-            </div>
+            </div>}
+            {!title && 
+            <div className={css.body} style={{marginTop: 0}}>
+              <XVegaVisualization model={{ specification, data, grammar }} />
+            </div>}
```
mturoci, line 158: Let's try to avoid unnecessary duplication. Do you think we could move the `margin` from `body` to `title` instead? That would remove the need for hacky conditional logic.

mturoci (changes requested):
Thanks for the valuable contribution @MattiasOz! Just need to resolve 1 comment and it's good to be merged.

ui/src/vega.tsx
mturoci: Does the `vega_card` with a `title` look the same as before? A quick look at the code suggests the margin is missing.

MattiasOz: The margin is gone. We assumed it was a mistake since the other cards in the dashboard don't seem to have one. We assumed this is what the issuer was complaining about since the title doesn't actually have a reserved height if the string is empty. Our fix removes the margin altogether as well as the title div if it's empty. 

We could add marginBottom to the title div if the extra spacing was necessary.

mturoci: It would be better to keep the original behavior untouched - if the title is specified, it should render as before. If not specified, there should be no extra whitespace.

Please check the rest of the components that may suffer from similar problem as well.

MattiasOz: We added a margin when the title isn't blank. There will be no margin when the title is absent.

mturoci (approved):
LGTM. Thanks @MattiasOz!

PR #2149, opened 2023-10-03
feat: Format dates inside plot tooltips properly.

mturoci:

### DIFF
--- a/ui/src/plot.tsx
+++ b/ui/src/plot.tsx
@@ -1042,18 +1042,22 @@ export interface Visualization {
 const tooltipContainer = document.createElement('div')
 tooltipContainer.className = 'g2-tooltip'
 
-const PlotTooltip = ({ items, originalItems }: { items: TooltipItem[], originalItems: any[] }) =>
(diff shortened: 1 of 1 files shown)


marek-mihok:
LGTM @mturoci! I went through all of the plot examples and also tested different scenarios (e.g. updating the data) and everything is working fine. The code also makes sense. There's only one thing I've noticed: After your change the date is also formatted in examples with only the `year` specified, e.g. `plot_line.py`.

This PR:
image

Current master:
image

However, I think your implementation makes sense (to format tooltip date if `x_scale === 'time'`) so I would only suggest removing `x_scale='time'` prop from these `.py` examples. WDYT?

mturoci:
> I would only suggest removing x_scale='time' prop from these .py examples. WDYT?

I don't think it's necessary since people may stumble into this scenario regardless of its presence in examples. I would say this is expected since `scale='time'`. If anyone wants a plain number in the tooltip, they should go with the default scale instead.

marek-mihok:
> If anyone wants a plain number in the tooltip, they should go with the default scale instead

@mturoci agreed. Then it's good to go.

Issue #2103, opened 2023-07-26
Show table groups without table rows


Nayananga:
#### Wave SDK Version, OS

(e.g. 4.2.0, Windows)

0.26.1, MacOS 

#### Actual behavior

(A clear and concise description of what happened.)
I want to show table groups that may or may not have table rows, But when I provide `rows=[]` to a particular `table_group` it doesn’t visible in the UI.

```
ui.table_group(
    label="Model Name 3",
    rows=[],
    collapsed=False,
)
```

But when I provide a `table_group` with a `table_row` where it has an empty list for `cells` it shows `Invalid Date` text for corresponding empty values where the `table_column`  has `data_type="time"`

```
ui.table_group(
    label="Model Name 2",
    rows=[
        ui.table_row(
            name="model_2_version_3_table_row",
            cells=[],
        )
    ],
    collapsed=False,
),
```

<img width="760" alt="Screenshot 2023-07-26 at 17 50 35" src="https://github.com/h2oai/wave/assets/25351304/e23d5eed-0972-46e2-9f19-261b9d602622">

#### Expected behavior

(A clear and concise description of what you expected to happen.)
Table groups should be able to exist even without table rows

#### Steps To Reproduce
sample code

```
from h2o_wave import main, app, Q, ui


@app('/')
async def serve(q: Q):
    q.page['meta'] = ui.form_card(
        box='1 1 5 5',
        items=[
            ui.table(
                name='models_table',
                height="400px",
                columns=[
                    ui.table_column(
                        name="model_version_name_table_column",
                        label="Model Version Name",
                    ),
                    ui.table_column(
                        name="created_date_table_column",
                        label="Created Date",
                        data_type="time",
                    ),
                    ui.table_column(
                        name="last_updated_table_column",
                        label="Last Updated",
                        data_type="time",
                    ),
                ],
                groups=[
                    ui.table_group(
                        label="Model Name 1",
                        rows=[
                            ui.table_row(
                                name="model_1_version_1_table_row",
                                cells=[
                                    "Model Version 1",
                                    "2023/07/26 5:30 pm",
                                    "2022/06/25 04:29 am"
                                ],
                            ),
                            ui.table_row(
                                name="model_1_version_2_table_row",
                                cells=[
                                    "Model Version 2",
                                    "2023/07/26 5:30 pm",
                                    "2022/06/25 04:29 am"
                                ],
                            )
                        ],
                        collapsed=False,
                    ),
                    ui.table_group(
[...]

PR #2150, opened 2023-10-03
chore: Prevent internal table groups regression. #2103

mturoci:


marek-mihok:
Well done, @mturoci! 

I'm not able to find any more issues with the table.

PR #2151, opened 2023-10-04
fix!: Return originally passed data as select events instead of mutated ones.

mturoci:

# Breaking change alert

* The returned event data is now the same as passed in. Prior to this PR, the data could have been mutated eg when `scale='time'` to timestamps.
* Slight return format change: The events are returned in a flat array instead of a nested (`[[]]`) array for line, area plots etc.

### DIFF
--- a/py/examples/plot_events_disabled.py
+++ b/py/examples/plot_events_disabled.py
@@ -1,4 +1,4 @@
-# Plot / Events/ Disabled
+# Plot / Events / Disabled
 # Customize for which marks on a #plot card you do not wish to handle #events .
 # ---
 from h2o_wave import main, app, Q, ui, data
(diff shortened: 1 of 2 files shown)


marek-mihok (commented):
LGTM @mturoci!

But note that taking into account possible override of `idx` value (that user specifies `idx` as a field name) does not make sense since the `idx` is overiden in `refactorData` on line 591 anyway:
```tsx
ds.forEach((d, idx) => d.idx = idx)
```
You can try replacing `value` with `idx` in the `plot_events_disabled.py` example and see for yourself. Data are plotted with `0,1,2,3,4...` indexes regardless of what indexes are specified.

image

It's worth considering fixing.

mturoci:
Good point. Not related to the events though so will fix in another commit.

PR #1988, opened 2023-05-25
fix: Set default width for components

marek-mihok:
The width of the following components was affected when used with `ui.inline`:

- `ui.dropdown`
- `ui.date_picker`
- `ui.plot`
- `ui.vega`
- `ui.table`
- `ui.slider`
- `ui.range slider`
- `ui.toggle`

https://github.com/h2oai/wave/assets/23740173/ba5974db-7be4-4c36-8639-54033e12b1d6

By default each of these elements either stretches to the width of its parent or the width can be explicitly specified through the `width` prop. The problem is when the parent has no defined width, in our case when placed inside `ui.inline`.

This PR sets the default width for each of these components in case parent width is not defined.

https://github.com/h2oai/wave/assets/23740173/b8b6cdb9-aaba-404b-a7ec-2f491655d7a5

I also include the example used in recordings:
```py
from .synth import FakeCategoricalSeries
from h2o_wave import main, app, Q, ui, pack, data
import random

html = '''
<ol>
    <li>Spam</li>
    <li>Ham</li>
    <li>Eggs</li>
</ol>
'''
spec = '''
{
  "description": "A simple bar plot with embedded data.",
  "mark": "bar",
  "encoding": {
    "x": {"field": "a", "type": "ordinal"},
    "y": {"field": "b", "type": "quantitative"}
  }
}
'''
f = FakeCategoricalSeries()
n = 20


# Generate random datum between 1 and 100
def rnd():
    return random.randint(1, 100)


@app('/demo')
async def serve(q: Q):
    q.page['example'] = ui.form_card(box='1 1 5 -1', items=[
        # ROW
        ui.inline(items=[
            ui.date_picker(name='date_picker', label='DP'),
            ui.dropdown(name='dropdown1', label='DPDWN', choices=[
                ui.choice(name=x, label=x) for x in ['Egg', 'Bacon', 'Spam']
            ]),
            ui.dropdown(name='dropdown2', label='DPDWN', values=[], choices=[
                ui.choice(name=x, label=x) for x in ['Egg', 'Bacon', 'Spam']
            ]),
        ],
        ),
        ui.inline(items=[
            ui.dropdown(name='dropdown3', label='DPDWN2', popup='always', values=[], choices=[
                ui.choice(name=x, label=x) for x in ['Egg', 'Bacon', 'Spam']
            ]),
            ui.dropdown(name='dropdown4', label='DPDWN3', popup='always', values=[], choices=[
[...]

Issue #1974, opened 2023-05-12
Default slider width in ui.inline is wrong


mturoci:
### Discussed in https://github.com/h2oai/wave/discussions/1973

<div type='discussions-op-text'>

<sup>Originally posted by **LaurenedePeyrelongue** May 11, 2023</sup>
Hi,

I am trying to adjust the size of a slider when it is in an ui.inline, however, I can't find any size/width option.
The slider is very small (see image below)
![image](https://github.com/h2oai/wave/assets/88673878/3e233328-6fb5-405a-9561-7de71d89a9d2)

So far the only way I managed to change the slider size is by adding characters to the slider title 
![image](https://github.com/h2oai/wave/assets/88673878/a050e3fe-815a-40a0-919a-6503a17c7a6f)

Is there any way to modify the size?
</div>

Check the rest of the components for the same problem as part of this issue.

PR #2152, opened 2023-10-05
chore: Use maximum available space in ui.inline instead of hardcoded constant. #1974

mturoci:

This is a rewrite of #1988 for #1974.

The reason some components shrunk drastically was wrong use of default `align='center` for case when `direction='column'`. This PR sets the following defaults:

* All components should try to occupy all the available space unless explicitly specified otherwise. This rule is in sync with regular `form_card.items`.
* The above default does not apply for cases when explicit alignment for main axis (`justify`) is chosen. In such case, the default width should correspond to the component's content, be it a label or anything else.

### DIFF
--- a/ui/src/form.tsx
+++ b/ui/src/form.tsx
@@ -238,26 +238,22 @@ type XComponentsProps = {
   inset?: B
   height?: S
   direction?: 'row' | 'column'
+  isInline?: B
 }
 
-const
-  needsDefaultWidth = new Set(['date_picker', 'dropdown', 'slider', 'range_slider', 'table', 'visualization', 'vega_visualization']),
-  getComponentWidth = (componentKey: S, isInline: B) => isInline && needsDefaultWidth.has(componentKey) ? '400px' : 'auto'
-
 export const
-  XComponents = ({ items, justify, align, inset, height, direction = 'column' }: XComponentsProps) => {
+  XComponents = ({ items, justify, align, inset, height, direction = 'column', isInline = false }: XComponentsProps) => {
     const
       components = items.map((m: any, i) => {
         const
           // All form items are wrapped by their component name (first and only prop of "m").
           [componentKey] = Object.keys(m),
-          isInline = !!(justify || align),
-          { name, visible = true, width = getComponentWidth(componentKey, isInline), height } = m[componentKey],
+          { name, visible = true, width = 'auto', height } = m[componentKey],
           visibleStyles: React.CSSProperties = visible ? {} : { display: 'none' },
           // TODO: Ugly, maybe use ui.inline's 'align' prop instead?
-          alignSelf = componentKey === 'links' ? 'flex-start' : undefined
-
-        if (m[componentKey].width !== width) m[componentKey].width = width
+          alignSelf = componentKey === 'links' ? 'flex-start' : undefined,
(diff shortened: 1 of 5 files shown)


ui/src/form.tsx
```diff
@@ -286,11 +282,12 @@ export const
   XInline = ({ model: m }: { model: Inline }) => (
     <XComponents
       items={m.items}
-      justify={m.justify || 'start'}
-      align={m.align || 'center'}
+      justify={m.justify}
+      align={m.align || m.direction === 'row' ? 'center' : undefined}
```
marek-mihok, line 286:
```suggestion
      align={m.align || (m.direction === 'row' ? 'center' : undefined)}
```
Without the parentheses the condition results in wrong value a therefore alignment can result always being `center` in some cases - please, see the **align** demo in broken `inline.py` example.

marek-mihok (changes requested):
Only one fix is needed and it's good to go, @mturoci !

ui/src/form.tsx
mturoci: Good catch, thanks!

marek-mihok:
@mturoci, my QA efforts has continued also after the review and brought the new finding that two of our examples using `ui.inline` need adjustment to new changes:

- background_progress.py

image

image



- form_visibility.py


image

image

Our Tour and Wave University apps are intact.
I'll further continue with my QA and let you know if something pops up.

mturoci:
Perfect. Those are expected and simple to fix. Good job nonetheless. Will fix in a sec.

marek-mihok:
@mturoci, I hope this is the last one - if the user sets a fixed width for the element, it should not shrink to fit the available space:

https://github.com/h2oai/wave/assets/23740173/72523d84-4bff-437c-9617-d5ac7d99df71

marek-mihok:
@mturoci, the similar problem as previous one, but with fixed `height` of the element when inline direction is `column`. When `height` of the `ui.inline` is lower than its content, the content does not shrink and there is no scrollbar either, it simply visibly overflows:

image

The example from the recording is [this one](https://github.com/h2oai/wave/pull/1988#issue-1725725890)  with visualization part modified:
```py
        ui.inline(direction='column', height='400px', items=[
            ui.visualization(
                height='500px',
                plot=ui.plot([ui.mark(type='interval', x='=product', y='=price', y_min=0)]),
                data=data(fields='product price', rows=[(c, x) for c, x, _ in [f.next() for _ in range(n)]], pack=True),
            ),
            ui.vega_visualization(
                height='500px',
                specification=spec,
                data=data(fields=["a", "b"], rows=[
                    ["A", rnd()], ["B", rnd()], ["C", rnd()],
                    ["D", rnd()], ["E", rnd()], ["F", rnd()],
                    ["G", rnd()], ["H", rnd()], ["I", rnd()]
                ], pack=True),
            ),
        ]),
```

mturoci:
The code above seems incorrect as it sets the height of the container `ui.inline=(height='400px')` and then wants to put there 2 `500px` plots, which naturally results in an overflow. Simply removing the container height should be enough.

marek-mihok:
> Simply removing the container height should be enough.

In that case it's all from my side. I've also checked if everything works to same way across all 3 browsers. If you don't have anything else for me to test, I'll finish my QA.

mturoci:
Thanks @marek-mihok!

PR #2153, opened 2023-10-05
feat: Allow using "idx" as field name in plot data.

mturoci:


ui/src/plot.tsx
```diff
@@ -1129,7 +1129,7 @@ export const
                     if (e.stateStatus && e.state === 'selected' && model.name && e.element.geometry.customOption.interactive) {
                       const ret = Array.isArray(e.element?.data)
                         ? e.element.data.map(({ idx }: any) => ({ idx, ...originalDataRef.current[idx] }))
```
marek-mihok, line 1131: You probably forgot to update `idx` to `wave_idx` when handling events in case you got array of objects.

```suggestion
                        ? e.element.data.map(({ wave_idx }: any) => ({ idx: wave_idx, ...originalDataRef.current[wave_idx] }))
```
Property `idx` does not exist when clicking on line:

image

Or another solution could be:
```suggestion
                        ?  e.element.data.map((_item: any, idx: I) => ({ idx, ...originalDataRef.current[idx] }))
```

marek-mihok (changes requested):
I went through all of our plot examples and it seems everything is plotted correctly. However, I left you one comment regarding events, please check it out.

ui/src/plot.tsx
mturoci: Thanks for catching that.

> Or another solution could be:

I went for the first suggestion as this one relies on loop indices instead the object ones, which would be a problem in case the data is sorted in a different order than arrived.

PR #2157, opened 2023-10-09
feat: Redirect to login on WS reconnect if auth specified

mturoci:
This PR solves the problem when websockets is not able to reconnect due to expired session. Instead of noop, it now redirects users to a login screen to renew the session.

As per [WS spec](https://www.rfc-editor.org/rfc/rfc6455):

> If the status code received from the server is not 101, the client handles the response per HTTP [[RFC2616](https://www.rfc-editor.org/rfc/rfc2616)] procedures.  In particular, the client might perform authentication if it receives a 401 status code; the server might redirect the client using a 3xx status code (but clients are not required to follow them), etc.  Otherwise, proceed as follows.

Clients (browser in our case) are not required to follow HTTP redirects, so they do not. For that reason, redirect request is sent as payload which is then handled manually.

@geomodular would you mind having a look at the code once you have time?

### DIFF
--- a/socket.go
+++ b/socket.go
@@ -15,9 +15,12 @@
 package wave
 
 import (
+	"encoding/json"
 	"net/http"
 	"strings"
 	"time"
+
+	"github.com/gorilla/websocket"
 )
 
 // SocketServer represents a websocket server.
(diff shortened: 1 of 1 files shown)


socket.go
```diff
@@ -35,21 +38,30 @@ func newSocketServer(broker *Broker, auth *Auth, editable bool, baseURL string,
...
+		return
+	}
+
 	session := anonymous
 	if s.auth != nil {
 		session = s.auth.identify(r)
 		if session == nil {
-			http.Error(w, http.StatusText(http.StatusUnauthorized), http.StatusUnauthorized)
+			// As per websocket spec, clients are not required to follow HTTP redirects. So we send a redirect message.
+			if msg, err := json.Marshal(OpsD{U: s.baseURL + "_auth/logout"}); err == nil {
```
geomodular, line 52: Does the client handle this message already?

geomodular (approved):
LGTM, thanks Martin!

socket.go
mturoci: Yes - https://github.com/h2oai/wave/blob/42008b74a65520c3adc48c4f1f0bd2e22445b3e8/ui/src/ui.ts#L136.

PR #2166, opened 2023-10-19
azure openid faq update

MaksymSofer:
Update to security docs regarding setup of OpenID with Azure


mturoci (approved):
Thanks @MaksymSofer! I changed formatting a bit.

Issue #2163, opened 2023-10-18
Integrate github workflows to automate the air-gapping process


sulhicader:
#### Description

Integrate workflows in [wave-template](https://github.com/h2oai/wave-template/tree/main/.github/workflows) repo to automate the air-gapping process of wave-university.

mturoci:
@sulhicader can you elaborate? Wave University is already air-gapped IIRC.

sulhicader:
There are some new flags introduced to h2o cli to generate air-gapped bundles of wave apps so those can be integrated in to replicated bundle. this [doc](https://github.com/h2oai/h2o-ai-cloud/blob/master/documentation/docs/developerguide/cli.md#airgapped-bundles) contains more info about it. github workflows in [wave-template](https://github.com/h2oai/wave-template/blob/main/.github/workflows/) automate the end to end process of building the docker images and publishing the docker images and helm charts to ECR which are generated from h2o cli command. This workflows should be integrated in the core wave apps listed in the [issue](https://github.com/h2oai/cloud-platform/issues/97)

mturoci:
Closed via #2181

PR #2181, opened 2023-11-02
packaging: Integrate github workflows to automate the air-gapping process of wave university

sulhicader:

This closes issue #2163


mturoci:
Thanks for the PR @sulhicader!

What's the reason for splitting into multiple files/workflows? It seems like an unnecessary abstraction that introduces plenty of duplication which will be hard to maintain for a few relatively trivial tasks - build and publish docker/helm.

Can we put it into existing workflow files instead? Also, it seems like you are only doing this for Wave University app so there is no need to make it overly parametrized, and hardcoding for simplicity should be fine for now.

sulhicader:
@mturoci Thanks you for the inputs. 

.github/workflows/wave-bundle-docker-build-publish.yaml, .github/workflows/wave-bundle-helm-release.yaml workflows will going to be common for all the apps going to pusblish from wave repo ( for now university and tour ). Also I mainly using the workflows from https://github.com/h2oai/wave-template/tree/main/.github/workflows template repo and in future if any additions comes to it, it will be easier to reflect those here. Here new workflows are triggered from .github/workflows/publish-university.yml which is existing workflow. As you suggest We can reduce the workflow files but putting into one file, but it also deviate from the standard template. WDYT?

mturoci:
The template is designed to build an arbitrary app for scratch, not to be integrated into existing workflows like here so it doesn't make sense to introduce it since all the apps have their builds already defined.

* The builds only need to be adjusted to use h2o cli instead of plain zip.
* Add steps to push to ECR / Helm as needed.

Correct me if I missed something in ^^.

Also, if anything changes, I can guarantee you (or anyone) will be able to resolve it much faster in 2 files with proper context than juggling across 6 files.

sulhicader:
Sure @mturoci, I am working on this.

sulhicader:
Hi @mturoci, I addressed the review comments and took the functionalities into one script. Please feel free to add comments  if any changes needed.

PR #2195, opened 2023-11-16
fix: Do not show selected tags in picker suggestions when no input is provided #1985

### LINKED ISSUE #1985: Same value can be picked multiple times in ui.picker
#### Wave SDK Version, OS

Wave SDK: 0.24.2 , OS: MacOs

#### Actual behavior

https://github.com/h2oai/wave/assets/89178049/6b887789-3500-4d6e-8e3d-16512b97cae9

When using a `ui.picker`, we can select a choice more than one time.
Here, after the selection of a value, the suggestions list doesn't show the selected value when typing that selected value within the box. That is fine. 
But the selected value still visible and selectable in the suggestions list when clicking on the box.

#### Expected behavior

When using a `ui.picker`, we won't be able to select a choice more than one time.
After the selection of a value, when clicking on the box,
1. The selected value should be visible but shouldn't be selectable in the suggestions list. (Disable the selection.)
2. Or The selected value should be removed from the suggestions list. (If the name of an ui.choice exists in the values list then that choice shouldn't be shown in the suggestions list.)

So far, It works well when typing the selected value within the box.

#### Steps To Reproduce

```py


@app('/')
async def serve(q: Q):
        box='1 1 4 5',
        items=[
            ui.picker(
                name='picker',
                label='Tags:',
                choices=[
                    ui.choice(name='spam', label='Spam'),
                    ui.choice(name='eggs', label='Eggs'),
                    ui.choice(name='ham', label='Ham'),
                    ui.choice(name='cheese', label='Cheese'),
                    ui.choice(name='beans', label='Beans'),
                    ui.choice(name='toast', label='Toast'),
                ],
                values=['eggs'],
            ),
        ]
    )
    await q.page.save()
```

marek-mihok:

___ 

The problem was the implementation of `onEmptyResolveSuggestions` callback. As Fluent API description says, it is

> A callback for what should happen when suggestions are shown without input provided. Returns the already selected items so the resolver can filter them out.

In our case it was returning initial tags (`() => tags`) instead of filtered ones.

https://github.com/h2oai/wave/assets/23740173/6c7ab461-ca83-418a-bed7-3c3d166d6b89

Closes #1985

### DIFF
--- a/ui/src/picker.test.tsx
+++ b/ui/src/picker.test.tsx
@@ -160,15 +160,20 @@ describe('Picker.tsx', () => {
   })
 
   it('Filters correctly - does not offer already selected - without input provided', () => {
-    const { getByRole, queryByRole } = render(<XPicker model={pickerProps} />)
+    const { getByRole, queryByRole, getAllByRole } = render(<XPicker model={pickerProps} />)
     const input = (getByRole('combobox') as HTMLInputElement)
 
+    fireEvent.click(input)
+    expect(getAllByRole('option')).toHaveLength(2)
     typeToInput(input, name)
     fireEvent.click(getByRole('option'))
+
+    fireEvent.click(input)
+    expect(getAllByRole('option')).toHaveLength(1)
     typeToInput(input, altName)
     fireEvent.click(getByRole('option'))
-    fireEvent.click(input)
 
+    fireEvent.click(input)
     expect(queryByRole('option')?.querySelector('.ms-Suggestions-none')).toBeInTheDocument()
   })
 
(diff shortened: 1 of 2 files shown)


ui/src/picker.test.tsx
```diff
@@ -159,6 +159,19 @@ describe('Picker.tsx', () => {
...
 
+  it('Filters correctly - does not offer already selected - without input provided', () => {
+    const { getByRole, queryByRole } = render(<XPicker model={pickerProps} />)
+    const input = (getByRole('combobox') as HTMLInputElement)
+
+    typeToInput(input, name)
+    fireEvent.click(getByRole('option'))
+    typeToInput(input, altName)
+    fireEvent.click(getByRole('option'))
+    fireEvent.click(input)
```
mturoci, line 170: Let's also check if we get the expected options step by step, e.g. render 2 choices, select 1, 1 should be present, select the last one, none should be present.

mturoci (approved):

PR #2223, opened 2023-12-12
chore: Add more debug logging.

### LINKED ISSUE #2185: Waved panicked
#### Wave SDK Version, OS

0.26.2, Kubernetes (Managed Cloud)

#### Actual behavior
A wave app crashed but for some reason the container stayed up. This caused outage for at least one customer. 

```go
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x77b60a]

goroutine 9288 [running]:
github.com/h2oai/wave.(*App).send(0xc0001aa000, {0x0, 0x0}, 0xc0001781e0, {0xb07f40, 0x2a, 0x2a})
	/home/runner/work/wave/wave/app.go:112 +0x54a
github.com/h2oai/wave.(*App).forward(0xc0001aa000, {0x0?, 0x0?}, 0x0?, {0xb07f40?, 0xc0004e0050?, 0x8ef900?})
	/home/runner/work/wave/wave/app.go:89 +0x2f
github.com/h2oai/wave.(*Broker).resetClients.func1(0xc0004e8390?)
	/home/runner/work/wave/wave/broker.go:214 +0x36
created by github.com/h2oai/wave.(*Broker).resetClients
	/home/runner/work/wave/wave/broker.go:213 +0x98
```

A very similar panic happened at least once before: https://github.com/h2oai/wave/discussions/1949

mturoci:
Closes #2185

* TMP debug for Keycloak investigation purposes. Will cut a new internal release for MLOPs team to deploy, get logs and make next steps based on findings.
* Silences the current error to get more time for investigation.

### DIFF
--- a/app.go
+++ b/app.go
@@ -116,7 +116,14 @@ func (app *App) send(clientID string, session *Session, data []byte) error {
 			req.Header.Set("Wave-Refresh-Token", session.token.RefreshToken)
 		} else {
 			// Should never happen.
-			echo(Log{"t": "app", "error": "missing access token in send", "session": session.id})
+			echo(Log{
+				"t":          "app",
+				"error":      "missing access token in send",
+				"sessionID":  session.id,
+				"subject":    session.subject,
+				"username":   session.username,
+				"access_url": session.successURL,
+			})
 		}
 	}
 
--- a/app.go
+++ b/app.go
@@ -109,9 +109,15 @@ func (app *App) send(clientID string, session *Session, data []byte) error {
 	req.Header.Set("Wave-Subject-ID", session.subject)
 	req.Header.Set("Wave-Username", session.username)
 	if session.subject != anon {
-		req.Header.Set("Wave-Access-Token", session.token.AccessToken)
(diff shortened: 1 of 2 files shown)


mwysokin:
Anything more we could get here? Like printing the entire session info?

mturoci:
@mwysokin sure. Anything else that comes to your mind?

mwysokin:
@mturoci  Literally everything that we could print from that place in code would be helpful. Also I don't know if it's something that's present in go frameworks but maybe there's some special env variable or flag we could provide to the auth framework so that we could lower its log level to TRACE or to the lowest one possible or maybe you could hardcode it in this special version of the framework to get more information directly from the framework.

mturoci:
> maybe there's some special env variable or flag we could provide to the auth framework so that we could lower its log level to TRACE

Unfortunately there is not. It's just a [lib](https://pkg.go.dev/golang.org/x/oauth2).

PR #2229, opened 2024-01-04
feat: Allow installing custom packages inside Wave Studio #2193

### LINKED ISSUE #2193: Allow for installing arbitrary packages in Wave-studio
A short term solution could be a button that would bring up a dialog with:

* package name textbox
* package version textbox (optional, defaults to latest)
* A submit button.

Once the button is submitted, install the required package, display progress and final result. Then prompt user to install another or cancel the dialog.

Nice to have: Allow uploading requirements.txt file in addition to one-by-one installs.

marek-mihok:

___

https://github.com/h2oai/wave/assets/23740173/1a7943b5-9471-4615-b33c-99107036e35f

# Code explanation

## The core concept

The core function doing the magic of package installation is our `pip` function which uses `subprocess.Popen`. In contrast to `subprocess.run` waiting for the result, Popen allows us to handle custom logic during its execution. This allows us to read its progress output and show display it to user. For reading the output from within the app, we forward stdout into `subprocess.PIPE`. We also use `text=True` attribute to get output as a string instead of bytes object. 
Popen is used inside context manager to properly close the pipes after the execution and handle other cleanup logic. Inside the Popen context we run the event loop checking whether task is finished. If it is not, it updates the current progress displayed to user. It it is finished - if `p.poll()` has some value - it shows success/error message.
The whole logic is wrapped inside `try` and `except` block to handle UI updates when task is cancelled.

The `pip` function has following arguments: `q: Q, command: str, args: [str], on_success: Callable = None, on_error: Callable = None, on_cancel: Callable = None`. The `q` is used for UI updates, `command` stands for `pip` command, in our case it is `install` or `uninstall`, `args` are additional arguments that can be passed into `pip` command, e.g. `-r` for installing from `requirements.txt` file or `-y` to automatically bypass questions with `yes` answers during uninstall process.

## Step by step

Clicking the button **Manage packages** in header items opens the side panel with the package management interface.

This interface consists of a table of currently installed packages loaded from `project/requirements.txt` file and two buttons - **Add package** and **Add from requirements.txt**. Please note that side panel items are obtained via `get_side_panel_items()` function. Having a separate function is useful when updating side panel with currently installed packages after install/uninstall.

### Installing single package

By clicking on `Add package` the `q.args.show_add_package_fields:` handler part is executed showing `package name` and an optional `package version` fields with the `Add` button triggering the installation process.

When the `Add` button is clicked and the package name field is not empty, the new asyncio task is created inside `q.args.add_package:` handler:
```py
q.client.task = asyncio.create_task(install(q, q.args.package_name, q.args.package_version))
```
Please note that return value is saved into `q.client.task` for having its reference in case one needs to cancel the execution.

The `install` function then calls our `pip` function, blocking the side panel, displaying current console progress to the user and running the `pip install` command.

Once the `pip` finishes successfuly, `on_install_success` is called,
which calls 
```py
update_requirements({f'{package_name}': version(package_name)})
```
The `update_requirements` function adds installed package into `requirements.txt` file if it exists otherwise new one is created. If the package was instaled before, but only the version has changed, the version is updated.

When the install process is finished, the appropriate (error or success) message is shown and one must press `Go back to package manager` button which calls the `q.args.finish_message_dismiss:` handler. The `on_pip_finish()` is called unblocking the side panel, making it closable again and updating it with currently installed packages.

### Installing from requirements.txt

By clicking on `Add from requirements.txt` the `q.args.show_add_requirements:` handler is called showing the file upload component. Once the requirements file is uploaded `q.args.upload_requirements:` handler is called saving the file into `project/requirements.txt` and running the `install` function.

The rest of the process is the same as during single package installation with the 2 differences. The `update_requirements` is called without any parameters just to update the uploaded requirements file with the installed versions. The second difference is that if the installation process fails, the requirements file is removed.

### Uninstalling the package(s)

Select the table rows containing packages you want to uninstall. The `Remove selected` button disabled state is based whether any item is selected using the table `select` event and the `q.events.table and q.events.table.select is not None:` handler. Once the buttons is clicked, `asyncio.create_task(uninstall(q, q.client.selected_packages))` is called.

Closes #2193

### DIFF
--- a/studio/studio.py
+++ b/studio/studio.py
@@ -316,14 +316,14 @@ async def on_pip_finish(q: Q):
     await q.page.save()
     
 
-def update_requirements(packages: dict, remove=False):
+def update_requirements(packages: dict = {}, remove=False):
     with open('project/requirements.txt', 'a+') as file:
         file.seek(0)
         lines = file.readlines()
 
     current_packages = {}
     for line in lines:
-        package_name, package_version = line.strip().split('==')
+        package_name, package_version = line.strip().split('==') if '==' in line else [line.strip(), version(line.strip())]
         current_packages[package_name] = package_version
 
     if not remove:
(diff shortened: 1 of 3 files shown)


studio/studio.py
```diff
@@ -237,6 +245,171 @@ async def export(q: Q):
...
+```
+'''
+
+
+async def show_progress(q: Q, msg: str):
+    q.page['meta'].dialog.items = [
+        ui.progress(label=msg),
+        ui.expander(name='console_expander', label='Detail', items=[
+            ui.text(name='console_out', content=get_output(''), width='100%'),
+        ]),
```
mturoci, line 291: The UX can be improved here. After hitting the install button, the (pip) command output should be streamed automatically instead of showing a loading indicator and then require extra click.

studio/studio.py
```diff
@@ -237,6 +245,171 @@ async def export(q: Q):
...
+              success_message: str, error_message: str, on_success: Callable = None, on_error: Callable = None):
+    try:
+        version = f'=={package_version}' if package_version else ''
+        args = [sys.executable, '-m', 'pip', command, flag, f'{package_name}{version}']
+        p = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        output = ''
+        await block_dialog(q)
+        await show_progress(q, progress_message)
+        while True:
+            if p.poll() is not None:
```
mturoci, line 336: Doesn't this kill the main thread?

studio/studio.py
```diff
@@ -237,6 +245,171 @@ async def export(q: Q):
...
+                break
+            else:
+                output += f'\n{p.stdout.readline().decode("utf-8")}'
+                await update_progress(q, output)
+    finally:
+        if p.poll() is None:
+            # Task canceled.
+            p.stdout.close()
+            p.stderr.close()
+            p.kill()
```
mturoci, line 356: Wouldn't proper context manager usage solve this for us?

studio/studio.py
```diff
@@ -237,6 +245,171 @@ async def export(q: Q):
...
+            await on_pip_finish(q)
+
+
+def update_requirements(packages: dict, remove=False):
```
mturoci, line 360: Let's make this simpler:

* Open the file.
* Read and parse its packages.
* Add/remove the unnecessary package.
* Write (save) packages back to file.

studio/studio.py
```diff
@@ -237,6 +245,171 @@ async def export(q: Q):
...
+    update_requirements({f'{package_name}': version(package_name)})
+
+
+async def on_uninstall_success(package_name: str):
+    update_requirements({f'{package_name}': ''}, True)
+
+
+async def on_requirements_install_success(package_name: str):
+    packages = {}
+    with open('project/requirements_tmp.txt', 'r') as file_tmp:
```
mturoci, line 389: Why do we need tmp requirements?

studio/studio.py
```diff
@@ -291,11 +465,91 @@ async def serve(q: Q):
                 editor.update_file_tree(q, project.dir)
                 await q.page.save()
                 editor.open_file(q, project.entry_point)
+                requirements_file = open('project/requirements.txt', 'r') if os.path.exists(
+                    'project/requirements.txt') else None
+                if requirements_file:
```
mturoci, line 470: Use proper context manager.

studio/studio.py
```diff
@@ -291,11 +465,91 @@ async def serve(q: Q):
...
+                    await show_packages_dialog(q)
+                    q.client.task = asyncio.create_task(pip(
```
mturoci, line 472: Name better.

studio/studio.py
```diff
@@ -291,11 +465,91 @@ async def serve(q: Q):
...
+                    ))
             else:
                 q.page['meta'].dialog.items = [
                     ui.message_bar(type='error', text='There must be exactly 1 root folder.'),
                     ui.button(name='import_project', label='Import again', icon='Upload'),
                 ]
+    elif q.args.show_packages_dialog:
+        await show_packages_dialog(q)
+    elif q.events.package_dialog and q.events.package_dialog.dismissed:
+        q.page['meta'].dialog = None
```
mturoci, line 492: There is always going to be just one dialog so no need to distinguish between `package_dialog` and `dialog` in terms of events. WIll make code simpler as there would be no need for separate handlers.

studio/studio.py
```diff
@@ -291,11 +465,91 @@ async def serve(q: Q):
...
+    elif q.args.show_add_requirements:
+        q.page['meta'].dialog.items[2] = ui.file_upload(name='upload_requirements', file_extensions=['txt'],
+                                                        label='Install packages',
+                                                        tooltip='Packages from requirements.txt will be added to already installed packages.')
+    elif q.args.upload_requirements:
+        os.mkdir('project/tmp')
+        file = await q.site.download(q.args.upload_requirements[0], os.path.join(os.getcwd(), 'project/tmp'))
+        shutil.copy(file, 'project/requirements_tmp.txt')
+        shutil.rmtree('project/tmp')
```
mturoci, line 501: Why is `tmp` needed?

mturoci (changes requested):
Thanks @marek-mihok, needs some work.

* Think whether side_panel wouldn't be a better fit than a dialog.
* The UX shall be closer to the terminal experience etc. - as a user, I would like to see the streamed output of my pip install rightaway. "Installing..." message has not much meaning for me.
* The code is hard to follow. Could you add a description in the PR?

studio/studio.py
marek-mihok: The original implementation was for uploaded `requirements.txt` to update existing one and this was preventing of the file overwrite. 

Removing and adding `Packages from requirements.txt will replace currently installed packages.` into file upload tooltip.

studio/studio.py
marek-mihok: The `subprocess.Popen()` is a non-blocking call. Using this kind of event loop inside its context should not be killing the main thread because if task is still running `p.poll()` returns `None` and UI is updated with `await update_progress(q, output)`. In case task is finished, `p.poll()` returns some value and the UI is updated accordingly as well.

marek-mihok:
@mturoci thanks for the valuable feedback and for the patience with my first "real" python task 🙂

> Think whether side_panel wouldn't be a better fit than a dialog.

Good idea, definitely more suitable when containing more content.
Done ✅

> The UX shall be closer to the terminal experience etc. - as a user, I would like to see the streamed output of my pip install rightaway. "Installing..." message has not much meaning for me.

The idea was to keep it simple and if user needs more detail, it is available one a single click.
No problem with your idea though.
Done ✅

> The code is hard to follow. Could you add a description in the PR?

Done ✅

mturoci (changes requested):
Thanks @marek-mihok. Looking better. Still needs some UX improvements:

* When I first click on Manage Packages, nothing happens. Need to click it twice to open the side panel.
* When adding a new package, there is no way to go back if I change my mind and decide not to install it. Split the UI into tabs: installed (table), new package, import requirements.txt
* New package form can be vertical instead of horizontal (we have enough space).
* New package submit button should be disabled until a package name is provided.
* Put "Remove selected" btn above the table, set its height to fill the rest of the side panel.

mturoci:
> Unfortunately, setting height=1 does not work when table is inside the side panel.

You can use CSS calc then. Same for terminal output. Would be also good to have the installation terminal always scrolled to the bottom (so that users do not need to scroll manually), the UX should be as close to the real terminal as possible as mentioned in the prev comment. There is `scrollLogsToBottom function already so implementation should be a matter of couple of minutes.

Once these are done, ready to merge. Tks!

marek-mihok:
> You can use CSS calc then.  Same for terminal output.

For table it works, but for terminal output I had to use custom JS, since text does no support `height` attribute. 

Maybe it would be good to add support for it in the future; it's definitely useful for markdown text.

PR #2246, opened 2024-01-31
fix: Ignore nil token err.

mturoci:
A tmp fix for #2185. Still did not manage to repro thus couldn't find the exact cause for why is the `token` nil. In theory, it's fine since this happens during logout anyway, but I would be happier if we managed to resolve the real cause so that there are no surprises in future.

### DIFF
--- a/app.go
+++ b/app.go
@@ -114,6 +114,10 @@ func (app *App) send(clientID string, session *Session, data []byte) error {
 		if session.token != nil {
 			req.Header.Set("Wave-Access-Token", session.token.AccessToken)
 			req.Header.Set("Wave-Refresh-Token", session.token.RefreshToken)
+		} else {
+			if app.broker.debug {
+				echo(Log{"t": "app_send", "error": "session token is nil"})
+			}
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


app.go
```diff
@@ -109,9 +109,12 @@ func (app *App) send(clientID string, session *Session, data []byte) error {
...
 	req.Header.Set("Wave-Username", session.username)
 	if session.subject != anon {
-		req.Header.Set("Wave-Access-Token", session.token.AccessToken)
-		req.Header.Set("Wave-Refresh-Token", session.token.RefreshToken)
 		req.Header.Set("Wave-Session-ID", session.id)
+		// TODO: Figure out how can the token be nil.
+		if session.token != nil {
+			req.Header.Set("Wave-Access-Token", session.token.AccessToken)
+			req.Header.Set("Wave-Refresh-Token", session.token.RefreshToken)
+		}
```
codyharris-h2o-ai, line 117: Could log a warning message here when this condition happens - we could then monitor for occurrences ie in Managed Cloud

codyharris-h2o-ai (approved):

app.go
mturoci: Good point. Added some logging (turned on by setting `H2O_WAVE_DEBUG`).

PR #2247, opened 2024-01-31
fix: Make conda package support multiple python versions #2227

### LINKED ISSUE #2227: Conda install works with python 3.11 only
#### Wave SDK Version, OS

1.0.0

#### Actual behavior

Running

```sh
conda create -n venv
conda activate venv
conda config --append channels conda-forge
conda install -c h2oai h2o-wave
```

results in Module not found err when trying to use `wave` command.

Using python 3.11 on the other hand works as expected

```sh
conda create -n venv python=3.11
conda activate venv
conda config --append channels conda-forge
conda install -c h2oai h2o-wave
```

#### Expected behavior

Should work python >= 3.8 as specified [here](https://github.com/h2oai/wave/blob/7887ef65cf037aea5cc6e845ea8a3193ccdb8261/py/h2o_wave/pyproject.toml#L8).

marek-mihok:

____

The problem was that the package only supports the version of python it was built in. It can be solved by specifying `noarch: python` inside `meta.yaml` conda config file. The result is the single platform independent package supporting multiple python versions.

Also update of the pipeline build action (`amauryval/publish_conda_package_action`) to version 2.0.3 is necessary, because it allows specifying `noarch` value as the `Platforms` parameter.

Closes #2227

### DIFF
--- a/.github/workflows/publish-conda.yml
+++ b/.github/workflows/publish-conda.yml
@@ -23,7 +23,10 @@ jobs:
       - name: Publish to Conda
         uses: amauryval/publish_conda_package_action@2.0.3
         with:
-          CondaDir: 'py/conda'
+          CondaDir: 'py/h2o_wave/conda'
           Platforms: 'noarch'
           CondaUsername: ${{ secrets.CONDA_USERNAME }}
           CondaPassword: ${{ secrets.CONDA_PASSWORD }}
+        env:
+          # VERSION clashes with conda build. Use PKG_VERSION instead.
+          PKG_VERSION: ${{ env.VERSION }}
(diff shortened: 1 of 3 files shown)


mturoci (approved):
Thanks @marek-mihok! We probably want to sync publish conda step in `release-wave` with `publish-conda` job as well.

mturoci:
> Is it for cases when you want to publish conda without going through all release process?

Yes.

> is CondaDir: 'py/conda' in publish-conda.yml instead of CondaDir: 'py/h2o_wave/conda' intentional?

Probably not. Seems like a leftover, but better check.

> And what exactly do you mean by "sync"?

They should be the same. No matter which workflow I use, I want the output to be identical.

marek-mihok:
@mturoci synced. Can we try to run `publish-conda` separately first to prevent `release-wave` from potential failing?

mturoci (approved):

PR #2253, opened 2024-02-05
feat: Allow reconnects without losing state/focus.

### LINKED ISSUE #2252: Make Wave more resilient to unstable networks
As of today, Wave requires a stable websocket connection. However, some environments cut websocket connections on purpose to save on resources which results in subpar UX.

The goal of this issue is to handle underlying reconnects in the background so that users can use the app undisturbed without notice.

* Requires UI change to only rerender the page if reconnect failed. First reconnect should be with zero backoff (immediate).
* Server needs to allow resumability - distinguish between reconnects and new connects.

cc @codyharris-h2o-ai @mmalohlava

mturoci:
## Demo (disconnect every 5s)

https://github.com/h2oai/wave/assets/64769322/6054d127-04c8-4476-996d-96913d3f7c08

Closes #2252


client.go
```diff
@@ -157,8 +171,10 @@ func (c *Client) listen() {
 				c.broker.sendAll(c.broker.clients[app.route], clearStateMsg)
 			}
 		case watchMsgT:
-			c.subscribe(m.addr) // subscribe even if page is currently NA
-
+			if c.isReconnect {
```
mturoci, line 174: Maybe `isReconnect` is redundant and I could use `c.cancel != nil`?

socket.go
```diff
@@ -70,8 +70,28 @@ func (s *SocketServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			}
 		}
 	}
+	clientID := r.URL.Query().Get("client-id")
```
mturoci, line 73: JS websocket API does not allow for setting custom headers so used query params instead.

client.go
```diff
@@ -90,15 +93,26 @@ func (c *Client) refreshToken() error {
...
-		if app != nil {
-			app.forward(c.id, c.session, disconnectMsg)
-			if err := app.disconnect(c.id); err != nil {
-				echo(Log{"t": "disconnect", "client": c.addr, "route": c.appPath, "err": err.Error()})
+		ctx, cancel := context.WithCancel(context.Background())
+		c.cancel = cancel
+		go func(ctx context.Context) {
+			select {
+			// Send disconnect message only if client doesn't reconnect within 2s.
+			case <-time.After(2 * time.Second):
```
codyharris-h2o-ai, line 101: Can this be configurable?

client.go
codyharris-h2o-ai: I think using `isReconnect` is the right call - imo it makes it clearer what's going on and in the future debugging a bit easier

socket.go
```diff
@@ -70,8 +70,28 @@ func (s *SocketServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	client, ok := s.broker.clientsByID[clientID]
+	if ok {
+		client.conn = conn
+		client.isReconnect = true
+		if client.cancel != nil {
+			client.cancel()
+		}
```
codyharris-h2o-ai, line 80: I think it might be worth logging something here (I would say debug to start, and then maybe reduce to trace) just to see how often it's happening.

codyharris-h2o-ai (approved):
lgtm, but it would be great if we could tune that reconnection window duration - 2s feels pretty short

client.go
mturoci: Yea, should be harmless I think.

socket.go
codyharris-h2o-ai: Other option would be to send the id on the first message or something but I think this is better

mturoci: Yep, I think as well. No need to clutter our custom protocol even more.

geomodular (approved):
LGTM Martin, good job!

codyharris-h2o-ai (approved):

PR #2262, opened 2024-02-08
feat: Make ui.command to support path and download props #2224

### LINKED ISSUE #2224: Add Download ability to ui.command
In our documentation, we show making a `ui.command` that has the icon and label of "Download".  This assumes that `ui.command` can perform the download function.

Can we add this ability to `ui.command` and other components where it would make sense?

marek-mihok:

___

Adds `path` and `download` attributes to **ui.command** to support the same behavior as `ui.link()`:

```tsx
export interface Command {
  /** An identifying name for this component. If the name is prefixed with a '#', the command sets the location hash to the name when executed. */
  name: Id
  /** The text displayed for this command. */
  label?: S
  /** The caption for this command (typically a tooltip). */
  caption?: S
  /** The icon to be displayed for this command. */
  icon?: S
  /** Sub-commands, if any */
  items?: Command[]
  /** Data associated with this command, if any. */
  value?: S
  /** The path or URL to link to. The 'items' and 'value' props are ignored when specified. */
  path?: S
  /** True if the link should prompt the user to save the linked URL instead of navigating to it. */
  download?: B
}
```

Closes #2224

### DIFF
--- a/ui/src/button.test.tsx
+++ b/ui/src/button.test.tsx
@@ -16,7 +16,6 @@ import { fireEvent, render } from '@testing-library/react'
 import React from 'react'
 import { Buttons, MiniButtons, XButtons, XMiniButtons } from './button'
 import { wave } from './ui'
-import userEvent from '@testing-library/user-event'
 
 const
   name = 'test-btn',
@@ -257,7 +256,7 @@ describe('Button.tsx', () => {
       expect(container.querySelector('i[data-icon-name="ChevronDown"]') as HTMLLIElement).not.toBeInTheDocument()
     })
 
-    it('Does not set args and calls sync on click when download link is specified', () => {
+    it('Does not set args and calls sync on click when command has download link specified', () => {
       // TODO: Use "true" only instead of value
       const
         value = 'value',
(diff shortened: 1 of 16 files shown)


ui/src/button.test.tsx
```diff
@@ -254,5 +255,93 @@ describe('Button.tsx', () => {
...
+
+      expect(wave.args['command2']).toBe(false)
+      fireEvent.click(contextMenuButton)
+      fireEvent.click(getByText('Command 2'))
+      expect(wave.args['command2']).toBe(false)
+
+      expect(pushMock).toHaveBeenCalledTimes(1)
+    })
+
+    it('Ignores items when command has download link specified', () => {
```
mturoci, line 292: What does this test do? I couldn't figure it out.

mturoci (approved):
LGTM. Just one question.

ui/src/button.test.tsx
marek-mihok: It tests [this line of code ](https://github.com/h2oai/wave/blob/6712a4746994f69b3cae5f38052c037defa8a686/ui/src/card_menu.tsx#L84) to verify that command having `path` specified cannot have context menu with another sub-commands.

At least for me it does not make sense for command to be clickable in case of having context menu.

mturoci: Are subcommands invoked by a click? I thought they appear on hover.

mturoci: Got it, makes sense. Ty.

PR #2263, opened 2024-02-09
feat: Image annotation shapes improvements #1860

### LINKED ISSUE #1860: Image Annotator line width could be thinner during zoom

While zooming into the image to annotate smaller objects the line and mark points are too thick.

image

Line and markers could have the same width during annotation

### LINKED ISSUE #1862: Image Annotator default box/polygon background fill for all objects

During box annotation and especially for polygon segmentation it could be a bit more difficult to see the objects if we only show the border. It would be nice to fill the object background as well with a small alpha.

image

image

For selected boxes we already fill the background with small alpha. We could do the same buy default for all objects.

It would be also nice to apply the same edge color for all the poligon markers. Having too many points it might be difficult to see the line color.

marek-mihok:

___

This PR brings the following enhancements:

- Both points and aux points match the tag colors.
- Line thickness and point size does not change when zooming.

https://github.com/h2oai/wave/assets/23740173/096ac635-955d-4b00-bec0-b029b5d6f077

Closes #1860
Closes #1862

### DIFF
--- a/ui/src/image_annotator_polygon.ts
+++ b/ui/src/image_annotator_polygon.ts
@@ -214,7 +214,7 @@ export class PolygonAnnotator {
     if (!this.ctx) return
 
     const path = new Path2D()
-    path.arc(x, y, this.ctx.lineWidth * 2, 0, 2 * Math.PI)
+    path.arc(x, y, (ARC_RADIUS_DEFAULT * this.ctx.lineWidth) / 2, 0, 2 * Math.PI)
     this.ctx.strokeStyle = isAux ? color.substring(0, color.length - 2) + '0.5)' : color
     this.ctx.fillStyle = isAux ? '#b8b8b8' : '#FFF'
     this.ctx.fill(path)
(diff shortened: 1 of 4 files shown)


ui/src/image_annotator_polygon.ts
```diff
@@ -1,12 +1,13 @@
...
 import { DrawnPoint, DrawnShape, ImageAnnotatorPoint, ImageAnnotatorRect } from "./image_annotator"
-import { ARC_RADIUS } from "./image_annotator_rect"
+import { ARC_RADIUS_DEFAULT } from "./image_annotator_rect"
 
 export class PolygonAnnotator {
   private currPolygonPoints: ImageAnnotatorPoint[] = []
   private boundaryRect: ImageAnnotatorRect | null = null
   private draggedPoint: DrawnPoint | null = null
   private draggedShape: DrawnShape | null = null
+  private getArcRadius = () => this.ctx ? ARC_RADIUS_DEFAULT / this.ctx.getTransform().a : ARC_RADIUS_DEFAULT
```
mturoci, line 10: Can be updated in the constructor when `ctx` is set. That way arcRadius can be computed just once and thus be made a constant instead.

mturoci (changes requested):
Good job @marek-mihok, just 1 small comment from me.

ui/src/image_annotator_polygon.ts
marek-mihok: The arcRadius is dependant of the currently applied transform (zoom) but constructor is called only once on image load. Therefore it cannot be constant.

PR #2271, opened 2024-02-20
feat: Add data-test attribute to picker item remove button #273

### LINKED ISSUE #273: Add data-test to picker item remove button
**Q SDK Version, OS**
0.8.0, 0.8.2 @ OS X

**Actual behavior**
Prior to 0.8.0 release the following code used to produce a DOM element with `data-test="country"` property:
`ui.picker(name='country', label='Region', values=selected_country, max_choices=1, choices=city_choices)`
 After 0.8.0, there's no DOM element with `data-test="country"`, which breaks e2e cypress tests

**Expected behavior**
If UI element has `name` property defined, it should be translated to DOM element with `data-test` property with the same value to enable cypress testing

**Steps To Reproduce**
1. create an app with `ui.picker(name='country', ....)`
2. run with wave 0.8.x
3. inspect the DOM of the generated UI

dbranley:
Closes #273 

As explained in the issue, the picker `data-test` attribute is missing from the DOM once the `max_choices` defined for the FluentUI TagPicker component is reached. With the change in this PR, a new `data-test` attribute is added to the remove button for the choices selected in the picker. This allows a Cypress test to locate the picker remove button with the `data-test` attribute and click it. Since the `max_choices` is no longer satisfied, the `input` element with the picker `data-test` attribute is added back to the DOM.

This was accomplished by using `removeButtonIconProps` on `TagPicker` to add a new `data_test` attribute that includes the picker name prefixed with `remove_`. The code in `ui/src/picker.tsx` looks like this:

```{typescript}
removeButtonIconProps={{ ...removeButtonIconProps, 'data-test' : 'remove_'+m.name } as Fluent.IIconProps}
```

There are 2 new test cases in `picker.test.tsx`. First is to confirm that the picker `data-test` attribute is unavailable when `max_choices` is met. The other is to confirm that the new `data-test` attribute for the remove button can be located and clicked, which returns the original picker `data-test` attribute back in DOM.

wave_ui_test_issue_273

Lastly, I also created a new Cypress test to confirm this would work. To support this, I modified the `picker_selection.py` example so that it defined a `max_choices` of 3. This allows the Cypress test to use an existing example. I noticed that there are no Python test definitions for Cypress in the repository, so I wasn't sure what to do with the test. Here is the test, you can see it is very simple:

```{python}
from h2o_wave import cypress

@cypress('Use picker and reach maximum number of choices')
def test_picker(cy):
    cy.visit('/demo')
    cy.locate('picker').type('Ham').type('{enter}')
    cy.locate('picker').should('not.exist')
    cy.locate('remove_picker').should('exist').first().click()
    cy.locate('picker').should('exist').type('Beans').type('{enter}')
    cy.locate('picker').should('not.exist')
```

Here is video showing the Cypress test in action:
https://github.com/h2oai/wave/assets/61979454/6cf63b01-1b8e-461d-a1b6-4b019c65e4cf

### DIFF
--- a/py/examples/picker_selection.py
+++ b/py/examples/picker_selection.py
@@ -14,14 +14,14 @@ async def serve(q: Q):
         ]
     else:
         q.page['example'] = ui.form_card(box='1 1 4 5', items=[
-            ui.picker(name='picker', label='Picker with initial values', choices=[
+            ui.picker(name='picker', label='Picker with initial values and max of 3 choices', choices=[
                 ui.choice(name='spam', label='Spam'),
                 ui.choice(name='eggs', label='Eggs'),
                 ui.choice(name='ham', label='Ham'),
                 ui.choice(name='cheese', label='Cheese'),
                 ui.choice(name='beans', label='Beans'),
                 ui.choice(name='toast', label='Toast'),
-            ], values=['spam', 'eggs']),
+            ], max_choices=3, values=['spam', 'eggs']),
             ui.button(name='show_inputs', label='Submit', primary=True),
         ])
     await q.page.save()
--- a/py/examples/picker_selection.py
+++ b/py/examples/picker_selection.py
@@ -14,14 +14,14 @@ async def serve(q: Q):
         ]
     else:
         q.page['example'] = ui.form_card(box='1 1 4 5', items=[
-            ui.picker(name='picker', label='Picker with initial values and max of 3 choices', choices=[
+            ui.picker(name='picker', label='Picker with initial values', choices=[
                 ui.choice(name='spam', label='Spam'),
                 ui.choice(name='eggs', label='Eggs'),
                 ui.choice(name='ham', label='Ham'),
                 ui.choice(name='cheese', label='Cheese'),
                 ui.choice(name='beans', label='Beans'),
                 ui.choice(name='toast', label='Toast'),
-            ], max_choices=3, values=['spam', 'eggs']),
+            ], values=['spam', 'eggs']),
             ui.button(name='show_inputs', label='Submit', primary=True),
         ])
     await q.page.save()
--- a/ui/src/picker.test.tsx
+++ b/ui/src/picker.test.tsx
@@ -41,6 +41,25 @@ describe('Picker.tsx', () => {
     expect(queryByTestId(name)).toBeInTheDocument()
   })
 
+  it('Does not render data-test attr when max choices reached', () => {
+    const { queryByTestId } = render(<XPicker model={{...pickerProps, values: [name], max_choices: 1}} />)
+    expect(queryByTestId(name)).toBe(null)
+  })
+
+  it('Remove selection so data-test attr for picker renders', () => {
+    const { queryByTestId } = render(<XPicker model={{...pickerProps, values: [name], max_choices: 1}} />)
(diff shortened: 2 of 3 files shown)


ui/src/picker.test.tsx
```diff
@@ -41,6 +41,25 @@ describe('Picker.tsx', () => {
     expect(queryByTestId(name)).toBeInTheDocument()
   })
 
+  it('Does not render data-test attr when max choices reached', () => {
```
mturoci, line 44: Not sure I understand these tests. We do want to render `data-test` every time a picker is rendered, irrespective of `max_choices`. `data-test_remove` should be rendered every time a value is selected (has a cancel button).

ui/src/picker.tsx
```diff
@@ -79,6 +83,7 @@ export const XPicker = ({ model: m }: { model: Picker }) => {
       {m.label && <Fluent.Label required={m.required}>{m.label}</Fluent.Label>}
       <Fluent.TagPicker
         inputProps={{ 'data-test': m.name } as Fluent.IInputProps}
+        removeButtonIconProps={{ ...removeButtonIconProps, 'data-test' : 'remove_'+m.name } as Fluent.IIconProps}
```
mturoci, line 86: Can be simplified

```suggestion
        removeButtonIconProps={{ iconName: 'Cancel', 'data-test': 'remove_' + m.name } as Fluent.IIconProps}
```

py/examples/picker_selection.py
```diff
@@ -14,14 +14,14 @@ async def serve(q: Q):
...
-            ui.picker(name='picker', label='Picker with initial values', choices=[
+            ui.picker(name='picker', label='Picker with initial values and max of 3 choices', choices=[
                 ui.choice(name='spam', label='Spam'),
                 ui.choice(name='eggs', label='Eggs'),
                 ui.choice(name='ham', label='Ham'),
                 ui.choice(name='cheese', label='Cheese'),
                 ui.choice(name='beans', label='Beans'),
                 ui.choice(name='toast', label='Toast'),
-            ], values=['spam', 'eggs']),
+            ], max_choices=3, values=['spam', 'eggs']),
```
mturoci, line 24: I am assuming this is a leftover from development. Please remove as these changes are not relevant to the feature this PR addresses.

mturoci (changes requested):
Thanks @dbranley!

Looks good overall, just not sure why `max_choices` need to be involved. See my comments for more details.

ui/src/picker.test.tsx
dbranley: You're right, `max_choices` is irrelevant here since the request is about introducing a `data-test` attribute on the picker's remove button. I have updated the test to remove references to `max_choices`.

py/examples/picker_selection.py
dbranley: You are correct. I added `max_choices` to the example for the Cypress test I created to test the changes. Of course that was unnecessary, per your other observation! I've reverted all the changes to the example.

dbranley:
Thanks @mturoci for all your great feedback! I've responded to each of your observations and pushed another commit with the changes. Let me know what you think!

mturoci (approved):
Now this is perfection :) Good job!

PR #2273, opened 2024-02-21
fix: Include spacing in width for inline horizontal components #2251

### LINKED ISSUE #2251: Changes in width handling since v0.26.3 makes handling of inline objects difficult
Hi,

here's a minimal example of inline objects that add up to 100%
```
page["form"] = ui.form_card(
    box="1 1 4 2",
    items=[
        ui.inline(
            items=[
                ui.textbox(name="textbox1", label="Textbox 1", width="34%"),
                ui.textbox(name="textbox2", label="Textbox 2", width="33%"),
                ui.textbox(name="textbox3", label="Textbox 3", width="33%"),
            ]
        )
    ],
)
```

1.0.1
image

0.26.3 (actually the screenshots are taken with 0.22.0 but I realized I'm using 0.26.3 right now and it's still fine)
image

Until version 0.26.3 100% remained the same total width no matter how crazy it gets
image

After 0.26.3 the 100% become unreliable, here examples for 1.0.1 and 100% and 50%/50%
image
image

All the best
Hannes

marek-mihok:

___

The problem was that left margin for ui.inline horizontally laid components adds to total component width causing the overflow. Since something like `boxSizing: "margin-box"` does not exist, the issue is fixed by replacing left margin with left padding and using `boxSizing: "border-box"` which accounts for both borders and paddings when calculating component dimensions.

image

This PR also includes a fix for firefox visual regression tests incompatibility with new `H2O_WAVE_RECONNECT_TIMEOUT` (#2252).

Solution tested across all browsers.
Also both unit and visual regression tests pass.

Closes #2251

### DIFF
--- a/tools/showcase/showcase.py
+++ b/tools/showcase/showcase.py
@@ -201,7 +201,9 @@ def main():
 
     try:
         os.makedirs(example_file_path)
-        wave_server = subprocess.Popen(['make', 'run'], cwd=os.path.join('..', '..'), stderr=subprocess.DEVNULL, preexec_fn=os.setsid) # noqa
+        env = os.environ.copy()
+        env['H2O_WAVE_RECONNECT_TIMEOUT'] = "0s"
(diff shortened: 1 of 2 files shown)


mturoci (approved):
LGTM

Issue #2276, opened 2024-02-26
Pre-release QA before v1.1.0


marek-mihok:
# Wave 1.1.0 pre-release QA

## Test results

- **Unit tests**

All pass

- **Visual regression tests**

As expected, only minor (barely noticeable) changes were introduced: left-side cells in markdown tables has no left padding anymore and links are little bit bolder.

![image](https://github.com/h2oai/wave/assets/23740173/db424e9c-cf47-4e1d-93ff-9f28712088ff)

<img width="1451" alt="image" src="https://github.com/h2oai/wave/assets/23740173/ea7039a4-ce45-4914-84ca-e820acf07c90">

## Manual testing 

Components changed since 1.0.0 were tested.

ui.audio_annotator ❌
ui.chatbot ✅
ui.card_menu ✅
ui.inline ✅
ui.image_annotator ✅
ui.link ✅
ui.markdown ✅
ui.notification_bar ✅
ui.picker ✅
ui.spin_box ✅
ui.stats ✅
ui.table ✅
ui.tall_stats  ✅
ui.textbox ✅
ui.toolbar ❌

### Audio annotator

1. Timeline works inconsistently across browsers, different files and file formats on Safari and Firefox:

While trying [aac sample](https://filesamples.com/samples/audio/aac/sample3.aac), cursor stucks few seconds before end on Safari and jumps to the end on Firefox:

https://github.com/h2oai/wave/assets/23740173/6380f24f-4220-48aa-864e-cd6d0a3f1ca9

It gets even more interesting for [flac](https://filesamples.com/samples/audio/flac/sample3.flac) files. The music is still playing, but the cursor is jumping on one place until the music ends. This was reproduced only for Safari:

https://github.com/h2oai/wave/assets/23740173/47af5498-73ba-427f-af47-6d72f216692a

However when trying [different aac file](https://dl.espressif.com/dl/audio/ff-16b-2c-44100hz.aac), issue persists for Safari only.

[Ogg](https://file-examples.com/wp-content/storage/2017/11/file_example_OOG_1MG.ogg) file was working for Chrome and Firefox, however on Safari it couldn't be decoded:

<img width="1415" alt="image" src="https://github.com/h2oai/wave/assets/23740173/77c3622e-9283-44e5-90b7-7884ee098d2b">

The [mp4 ](https://download.samplelib.com/mp4/sample-5s.mp4) file cannot be played on Safari:

[...]

PR #2278, opened 2024-03-05
fix: Close websocket connections as per spec.

mturoci:
Fixes FF websocket issue. @marek-mihok please verify if works for you as well.

Part of #2276


marek-mihok:
@mturoci works for me! Good job.

PR #2285, opened 2024-03-12
fix: Eliminate race conditions in reconnect feature.

mturoci:
Reconnect feature stabilization.

This PR adds R/W synchronization to client state shared across multiple goroutines, preventing race conditions. A simple state machine is used for the synchronization.

### DIFF
--- a/client.go
+++ b/client.go
@@ -111,14 +111,16 @@ func (c *Client) listen() {
 		if c.state != STATE_DISCONNECT {
 			return
 		}
-		c.state = STATE_TIMEOUT
+		// This defer runs to completion. If the client drops, reconnects and drops out again, ignore first drop timeout.
+		timeoutID := STATE_TIMEOUT + c.addr
+		c.state = timeoutID
 		c.lock.Unlock()
 
 		select {
 		// Send disconnect message only if client doesn't reconnect within the specified timeframe.
 		case <-time.After(c.reconnectTimeout):
 			c.lock.Lock()
-			if c.state != STATE_TIMEOUT {
+			if c.state != timeoutID {
 				return
 			}
 			app := c.broker.getApp(c.appPath)
--- a/client.go
+++ b/client.go
@@ -26,24 +26,22 @@ import (
 )
 
 const (
-	// Time allowed to write a message to the peer.
-	writeWait = 10 * time.Second
-
-	// Maximum message size allowed from peer.
-	maxMessageSize = 1 * 1024 * 1024 // bytes
-)
-
-var (
-	newline          = []byte{'\n'}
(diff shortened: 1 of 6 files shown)


client.go
```diff
@@ -90,29 +98,44 @@ func (c *Client) refreshToken() error {
...
-				}
+		c.lock.Lock()
+		defer c.lock.Unlock()
+		if c.state != STATE_DISCONNECT {
+			return
+		}
+		// This defer runs to completion. If the client drops, reconnects and drops out again, ignore first drop timeout.
+		timeoutID := STATE_TIMEOUT + c.addr
+		c.state = timeoutID
+		c.lock.Unlock()
```
geomodular, line 117: This line will make `c.lock` to unlock twice (due to `defer`).

client.go
```diff
@@ -90,29 +98,44 @@ func (c *Client) refreshToken() error {
 	return nil
 }
 
+func (c *Client) setState(newState string) {
+	c.lock.Lock()
+	c.state = newState
```
geomodular, line 103: Is it necessary to set a variable in a locked environment?

client.go
```diff
@@ -33,10 +34,16 @@ const (
...
 var (
-	newline       = []byte{'\n'}
-	notFoundMsg   = []byte(`{"e":"not_found"}`)
-	disconnectMsg = []byte(`{"data": {"":{"@system":{"client_disconnect":true}}}}`)
-	clearStateMsg = []byte(`{"c":1}`)
+	newline          = []byte{'\n'}
+	notFoundMsg      = []byte(`{"e":"not_found"}`)
+	disconnectMsg    = []byte(`{"data": {"":{"@system":{"client_disconnect":true}}}}`)
+	clearStateMsg    = []byte(`{"c":1}`)
+	STATE_CREATED    = "CREATED"
```
geomodular, line 41: The idiomatic way to represent a state (enumeration) would be a `const` value (not `var`) generated by `iota`(are those values in variables needed?).

client.go
mturoci: Wouldn't the deferred unlock run only after `select` statement is finished?

client.go
mturoci: Yes, because the variable is accessed across different goroutines. Better ideas welcome ofc.

client.go
mturoci: This was easier to debug, but since it works, can be refactored into `iota`.

client.go
geomodular: Ah, sorry, now I see what you meant!

geomodular (approved):
LGTM, thanks Martin!

PR #2303, opened 2024-03-27
feat: Include waved binary in Conda distributions #2266

### LINKED ISSUE #2266: Make waved part of conda bundle
Same as we do for Pypi wheels.

marek-mihok:

___

### Waved binaries are now included inside conda packages! 🎉 

As of today building of platform-specific packages supporting multiple python versions [is not supported by Conda](https://github.com/conda/conda-build/issues/2611#issuecomment-1239338538), so we need to build packages for each python version individually and we end up with 21 packages total (4 architectures * 5 python versions + 1 noarch package without waved binary):

image

This PR also features a [new gh action](https://github.com/marek-mihok/upload_to_anaconda_cloud_action/tree/v1.0.0) which only uploads already built packages into Anaconda cloud.

Closes #2266


marek-mihok:
@mturoci I didn't have chance to test new gh workflows so there is a possibility of failure, however I've tried to prepare everything for it to pass on the first try.

py/h2o_wave/Makefile
```diff
@@ -1,3 +1,19 @@
+build_conda_package = \
```
mturoci, line 1: Is it possible to create 2 separate make targets (conda-noarch and conda-arch) instead? And maybe a 3rd one for moving stuff around (the last section of the function)

py/h2o_wave/conda/meta.yaml
```diff
@@ -15,20 +15,33 @@ about:
...
       - {{ module }} = {{ entrypoints }}
-    {% endfor %}
\ No newline at end of file
+    {% endfor %}
+  include_recipe: False
+  script_env:
+    - NO_ARCH
+  {% if NO_ARCH == "1" %}
+  noarch: python
+  string: "0"
```
mturoci, line 46: What is this `string` for?

py/h2o_wave/hatch_build.py
```diff
@@ -26,6 +26,10 @@ def initialize(self, _version, build_data):
         if not platform:
             # Create a default metadata file in case of noarch builds.
             create_metadata_file('linux', 'amd64')
+            # If conda build, copy binaries into package.
```
mturoci, line 29: I though conda has its own build pipeline and doesn't use hatch. How does this work?

.github/workflows/publish-conda.yml
```diff
@@ -20,13 +19,39 @@ jobs:
...
+
+      - name: Build Release
+        run: make release
         env:
           # VERSION clashes with conda build. Use PKG_VERSION instead.
           PKG_VERSION: ${{ env.VERSION }}
+          NODE_OPTIONS: '--max-old-space-size=8192'
+
+      - name: Publish to Conda
+        uses: marek-mihok/upload_to_anaconda_cloud_action@v1.0.0
```
mturoci, line 53: What was the reason for switching from the previous conda publish action?

mturoci (commented):
Thanks @marek-mihok! Looks great. A few questions.

py/h2o_wave/conda/meta.yaml
marek-mihok: To override `string: py{{ python_version | replace(".", "") }}` in case of noarch builds.
However better would be doing something like this
```yaml
{% if NO_ARCH == "0" %}
string: py{{ python_version | replace(".", "") }}
{% endif %}
```
and remove `string: "0"` to let conda do its job with naming noarch builds.

.github/workflows/publish-conda.yml
marek-mihok: The previous action was building a package (the only input was path to conda dir with yaml files), converting it for every platform and uploading it to anaconda cloud. But you couldn't have control over moving files before each build.

That's why I added a workflow `Setup Miniconda` step and then the packages are built during `make release`. The new action then only uploads packages to Anaconda cloud.

I can also think of another solution and that is running the previous gh action multiple times with moving the files before each run. But this way if something fails, only part of the packages will be uploaded to Anaconda cloud.

mturoci: Sounds reasonable. Thanks!

PR #2329, opened 2024-05-10
feat: Support decimal values in table progress cell type #824

### LINKED ISSUE #824: Support decimal values in table progress cell type
I am using a `ui.progress_table_cell_type` to show some statistics in a table that has need for displaying decimal values.

Support for showing additional decimals instead of rounding to an integer.
Or
Support some other cell type that is suitable to decimal stats.

#### Comments
It should be optional / configurable of course, in most other cases I think rounding works great.

dbranley:
Closes #824 

As explained in the issue, the table progress cell currently rounds values to whole numbers when diplaying them as a percentage. With this change, the percentage will now render with up to 2 decimal places. This is accomplished by using toFixed(2) and parseFloat() together to ensure data is rounded properly and then trailing zeros are removed. The code change in `ui/src/progress_table_cell_type.tsx` looks like this:

```{typescript}
<div className={css.percent}>{`${Number.parseFloat((progress * 100).toFixed(2))}%`}</div>
```

Here is a partial screen shot of the Table example where you can see the progress data now being displayed with up to 2 decimal places:

output_example_issue_824

There is a new test case in `progress_table_cell_type.test.tsx`. The test rerenders the `XProgressTableCellType` component in a loop to make sure that the component is rendering the following input decimal values correctly:

```{typescript}
  progressValues = [
    {input: 0.88, output: '88%'},
    {input: 0.888, output: '88.8%'},
    {input: 0.8888, output: '88.88%'},
    {input: 0.88888, output: '88.89%'},
    {input: 0.88899, output: '88.9%'},
    {input: 0.88999, output: '89%'},]
```

wave_ui_test_issue_824

### DIFF
--- a/ui/src/progress_table_cell_type.test.tsx
+++ b/ui/src/progress_table_cell_type.test.tsx
@@ -25,8 +25,8 @@ const
     {input: 0.888, output: '88.8%'},
     {input: 0.8888, output: '88.88%'},
     {input: 0.88888, output: '88.89%'},
-    {input: 0.88899, output: '88.9%'},
-    {input: 0.88999, output: '89%'},]  
+    {input: 0.88899, output: '88.90%'},
+    {input: 0.88999, output: '89.00%'},]  
 
 describe('ProgressTableCellType.tsx', () => {
 
--- a/ui/src/progress_table_cell_type.test.tsx
+++ b/ui/src/progress_table_cell_type.test.tsx
@@ -19,7 +19,14 @@ import { ProgressTableCellType, XProgressTableCellType } from './progress_table_
 const
   name = 'progress-cell',
   progress = 0.0,
-  progressCellProps: ProgressTableCellType = { name }
+  progressCellProps: ProgressTableCellType = { name },
+  progressValues = [
+    {input: 0.88, output: '88%'},
+    {input: 0.888, output: '88.8%'},
+    {input: 0.8888, output: '88.88%'},
+    {input: 0.88888, output: '88.89%'},
+    {input: 0.88899, output: '88.9%'},
+    {input: 0.88999, output: '89%'},]  
 
 describe('ProgressTableCellType.tsx', () => {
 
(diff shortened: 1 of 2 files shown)


ui/src/progress_table_cell_type.test.tsx
```diff
@@ -19,7 +19,14 @@ import { ProgressTableCellType, XProgressTableCellType } from './progress_table_
...
   progress = 0.0,
-  progressCellProps: ProgressTableCellType = { name }
+  progressCellProps: ProgressTableCellType = { name },
+  progressValues = [
+    {input: 0.88, output: '88%'},
+    {input: 0.888, output: '88.8%'},
+    {input: 0.8888, output: '88.88%'},
+    {input: 0.88888, output: '88.89%'},
+    {input: 0.88899, output: '88.9%'},
+    {input: 0.88999, output: '89%'},]
```
mturoci, line 29: I would expect this to be:

```suggestion
    {input: 0.88899, output: '88.90%'},
    {input: 0.88999, output: '89.00%'},]  
```

The rationale being that if I have a list of floats, I want them all to have the same precision displayed (max 2 decimal places).

mturoci (changes requested):
Thanks @dbranley! Just 1 change required, otherwise like it a lot :)

ui/src/progress_table_cell_type.test.tsx
dbranley: @mturoci Thanks for the feedback and clarification. I had assumed that trailing 0's should be trimmed. But I can see now that you want progress data rendered with the same precision for consistency. I'll make the change and resubmit for review later today. Thanks!

dbranley: @mturoci I have a clarification question for you. Could you confirm that you do NOT expect to see the trailing zeros when the incoming percentage has less than 2 decimal places? In other words, the first 2 `progressValues` in the test case I added look like this:

```{typescript}
    {input: 0.88, output: '88%'},
    {input: 0.888, output: '88.8%'},
```

I assume this is what you want instead of seeing output such as `88.00%` and `88.80%` for these 2 examples? I assume that's the case since your comment was only on the last 2 items in the array, but I just wanted to double check before I commit something.

thanks!

mturoci: >  Could you confirm that you do NOT expect to see the trailing zeros when the incoming percentage has less than 2 decimal places?

Correct.

dbranley: @mturoci I have updated the PR with the change you requested. Here is a new screen shot from the Table example where you can see the trailing 0's for floats that are used in the example allow the data to be nicely aligned.

output_example_issue_824_20240514

I have also re-run all the unit tests and you can see they still pass:

wave_ui_test_issue_824_20240514

Let me know if you need any futher changes. Thanks!

mturoci (approved):
LGTM. Great job @dbranley!

Especially appreciate your independence, quick and constructive discussion and flawless execution (+ you included tests without me explicitly asking!). Prime example  of how OSS contributions should look like.

ui/src/progress_table_cell_type.test.tsx
mturoci: Perfect.

PR #2336, opened 2024-05-22
feat: Allow alignment for ui.text #1370

### LINKED ISSUE #1370: Allow alignment for ui.text

I am using HO2 wave to make a web app which need to collate the text to captured subtile images:

image

but textbox dont have a centered alignment option yet.

Need start/end/center alignment options for text input applications.

(A clear and concise description of any alternative solutions or features you've considered.)

dbranley:
Closes #1370

**The PR fulfills these requirements:** (check all that apply)

As requested in the ticket, I have added a new `align` prop in `text.tsx`. The possible values are `start`, `end`, `center`, and `justify`, with a default of `start`. This new prop is then passed down through the `style` prop for `Fluent.Text`. The code change in `ui/src/text.tsx` looks like this:

```typescript
  /** The alignment of the text content. Defaults to 'start'. */
  align?: 'start' | 'end' | 'center' | 'justify'
```

```typescript
    <Fluent.Text data-test={name} variant={toTextVariant(size || 'm')} style={{ textAlign: align || 'start'}} block className='w-text'>
```

I then ran the make generate target to update the generated python and R code to reflect this new attribute in files, such as `py/h2o_wave/h2o_wave/ui.py`.

I have updated the documentation in `website/widgets/form/text.md` with a new section called `Alignment`. I ran the make targets to regenerate the documentation artifacts and you can see a screen shot of that here:

doc_issue_1370

I tested this with a new custom python example that uses the new `align` attribute available for `ui.text`. You can find this at `py/examples/text_alignment.py`. And here is a screen shot of the new example:

output_example_issue_1370

I created a new unit test in `text.test.tsx`, where it tests each of the possible values for the new prop, along with confirming it defaults to `start`. Here is a snipit from the test:

```{typescript}
const textAlignPropValues: Text['align'][] = ['start', 'end', 'center', 'justify']
```

```{typescript}
  it('Renders Text with the align prop', () => {
    const { queryByTestId, rerender } = render(<XText {...textProps} />)
    expect(queryByTestId(name)).toHaveAttribute('style','text-align: start;')     
    textAlignPropValues.forEach(textAlignPropValue => {
      rerender(<XText {...textProps} align={textAlignPropValue} />)
      expect(queryByTestId(name)).toHaveAttribute('style','text-align: '+textAlignPropValue+';')      
    })
  }) 
```

wave_ui_test_issue_1370


ui/src/text.test.tsx
```diff
@@ -31,4 +32,15 @@ describe('Text.tsx', () => {
...
   })
 
+  it('Renders Text with the align prop', () => {
+    const { queryByTestId, rerender } = render(<XText {...textProps} />)
+    expect(queryByTestId(name)).toBeInTheDocument()
+    expect(queryByTestId(name)).toHaveAttribute('style','text-align: start;')     
+    textAlignPropValues.forEach(textAlignPropValue => {
+      rerender(<XText {...textProps} align={textAlignPropValue} />)
+      expect(queryByTestId(name)).toBeInTheDocument()
+      expect(queryByTestId(name)).toHaveAttribute('style','text-align: '+textAlignPropValue+';')
```
mturoci, line 42: TLDR: This unit test can be removed.

No need to unit test visual things as they are very coupled to the implementation - will break on refactor. We have visual regression tests for this scenario that can be used to verify no other components changed unexpectedly.

If you are interested, you can checkout to main, run `make build-ui` to build the UI, then `cd tools/showcase` and `make visual-regression`. If you have never run it before, it will create base snapshots that will be used for comparison in the next run. Now checkout your feature branch, build UI again and run visual regression tests again. It parses all the examples in documentation and compares against your baseline (main). It's super strict so false positives are very likely, but still good enough to quickly see if something broke or not. One more thing - if new examples are inserted (like it is now in `text.md`), the order will not match and you will get a lot of mismatches. In that case, you just check if the screenshots look "normal" since there is no way around that.

py/examples/text_alignment.py
```diff
@@ -0,0 +1,33 @@
+# Form / Text / Alignment
```
mturoci, line 1: Please add a screenshot of this example to `website/docs/examples/assets` to make docs build pass.

py/examples/text_alignment.py
```diff
@@ -0,0 +1,33 @@
...
+        ui.separator('Align - center'),
+        ui.text(text, align=ui.TextAlign.CENTER),
+        ui.separator('Align - justify'),
+        ui.text(text, align=ui.TextAlign.JUSTIFY),
+        ui.separator('Various sizes - align center'),
+        ui.text_xl('Extra large text', align=ui.TextXlAlign.CENTER),
+        ui.text_l('Large text', align=ui.TextLAlign.CENTER),
+        ui.text_m('Medium text', align=ui.TextMAlign.CENTER),
+        ui.text_s('Small text', align=ui.TextSAlign.CENTER),
+        ui.text_xs('Extra small text', align=ui.TextXsAlign.CENTER),
```
mturoci, line 29: No need to also include sizes, showing different alignments with default size is enough. The more concise the examples the better.

mturoci (changes requested):
Good job @dbranley! A few minor comments and good to go.

ui/src/text.test.tsx
dbranley: @mturoci Thanks for explaining this about the visual regression tests, I hadn't noticed it yet! I ran through this and think this is a good testing tool. I'll be sure to run this as a check before submitting any UI changes in the future.

py/examples/text_alignment.py
dbranley: @mturoci Thanks for letting me know about committing a screen shot for the new example. Here's a screen shot of the page after I rebuilt all the docs:

doc_example_issue_1370_20240523

py/examples/text_alignment.py
dbranley: @mturoci Done! Here's a screen shot of the new example:

output_example_issue_1370_20240523

dbranley:
@mturoci Thanks for all the great feedback! I think I've take care of all the points, but please let me know if I missed something or you notice something else. Thanks!

PR #2343, opened 2024-06-05
fix: Trigger multi combobox args after they are set, not before.

### LINKED ISSUE #2342: New values in ui.combobox trigger a button submission
#### Wave SDK Version, OS

h2o-wave 1.2.0

#### Actual behavior

Minimum code example:

```python
from h2o_wave import Q, app, main, ui


@app("/")
async def serve(q: Q):
    print(q.args)
    q.page['example'] = ui.form_card(box='1 1 2 2', items=[
        ui.combobox(
            name="my_coolest_combobox",
            label="Selection",
            values=["choice1"],
            required=False,
            choices=["choice1", "choice2"],
            trigger=True,
        ),
        ui.button(
            name="i_am_a_button",
            label=(
                "Don't click me"
            ),
            primary=True,
        )
    ])
    await q.page.save()
```

When adding a new value to the combobox, I am getting: 
`q.args: __wave_submission_name__:'i_am_a_button'`

When selecting an existing value in the combobox, I am getting: 
`q.args: __wave_submission_name__:'my_coolest_combobox'`

#### Expected behavior

In both cases, I would expect the `__wave_submission_name__` to be `'my_coolest_combobox'``

#### Steps To Reproduce

1. Use minimum code example above
2. Add an new value to the combobox and wait for the trigger
3. Select a value from the combobox and wait for the trigger
4. Compare q.args prints in the console

mturoci:
Closes #2342

### DIFF
--- a/ui/src/combobox.tsx
+++ b/ui/src/combobox.tsx
@@ -17,6 +17,7 @@ import { IComboBox, IComboBoxOption } from '@fluentui/react'
 import { B, Id, S, U } from './core'
 import React from 'react'
 import { wave } from './ui'
+import { flushSync } from 'react-dom'
 
 /**
  * Create a combobox.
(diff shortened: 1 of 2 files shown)


ui/src/combobox.tsx
```diff
@@ -110,22 +111,23 @@ const ComboboxMultiSelect = ({ model: m }: { model: Omit<Combobox, 'value'> }) =
...
-        return result
+      flushSync(() => {
+        setSelected(keys => {
+          const result = option.selected ? [...keys, String(option.key)] : keys.filter(key => key !== option.key)
+          wave.args[m.name] = result
+          return result
+        })
       })
+
+      if (m.trigger) wave.push()
```
marek-mihok, line 122: Can using the `flushSync` be avoided? This should be enough:
```suggestion
      const result = option.selected ? [...selected, String(option.key)] : selected.filter(key => key !== option.key)
      setSelected(result)
      wave.args[m.name] = result
      if (m.trigger) wave.push()
```

marek-mihok (changes requested):
Good job @mturoci! Everything is working as expected. Just one minor thing - using `flushSync` [can be risky](https://react.dev/reference/react-dom/flushSync) and it is a last resort solution. In this case I think we could go without it (see the suggestion).

ui/src/combobox.tsx
mturoci: Our `selected` state depends on its prev state thus a callback is necessary to make it safe.

flushSync is not risky, it just avoids batching state updates, but since we need to know the prev state, we do want sync updates anyway.

mturoci: Using your suggestion should work as well, but there are no guarantees from React that each user event like blur, tab, enter etc. commit state so I prefer to not depend on the internal implementation, especially when we are in the maintenance mode and I want each of us to spend at little time on this project as possible.

marek-mihok: Thanks for clarification! Your points make sense.

PR #2353, opened 2024-06-20
build: 🔒Enable and enforce `boringcrypto` experiment

zoido:
`GOEXPERIMENT=boringcrypto` switches `crypto` packages to call BoringCrypto library. It is included as a precompiled object with Go distibution so can be linked to statically.

We want to have all  binaries consistently built this way to get closer to FIPS 140 compliance.


mturoci (approved):
Thanks @zoido! Code LGTM, I also verified this works by running waved against self-signed certificate.

PR #2387, opened 2024-08-27
fix: Select All on 2nd filter in table not showing checked checkboxes #2386

### LINKED ISSUE #2386: Select All on 2nd filter in table not showing checked checkboxes
#### Wave SDK Version, OS

v1.5.0, MacOS

#### Actual behavior

This was initially reported during a [discussion](https://github.com/h2oai/wave/pull/2371#pullrequestreview-2245034119) in the PR for table filter counts.

When you have a table with 2 or more filterable columns, you can get into a situation where clicking the `Select All` button on the 2nd filter context menu does not change the state of the checkboxes for the filters. This happens when you filter something in one of the columns and then try to filter in another column.

#### Expected behavior

Clicking `Select All` or `Deselect All` on a filter context menu should always change the rendered state of the checkboxes displayed. 

#### Steps To Reproduce

1. Go to a table with 2 filterable columns.
2. Click the filter icon on the 1st column and from the filter context menu selected one of the filter checkboxes
3. Notice that the table updates as expected and the filter count for the column displays 1
4. Click the filter icon on the 2nd column and from the filter context menu click the `Select All` button
5. Notice that none of the checkboxes on the filter context menu are selected, this is incorrect
6. Notice that the filter count for the column displays the correct number, which means filtering worked even though the checkboxes were not checked
7. Click the filter icon on the 2nd column again
8. Notice all the checkboxes are rendered as checked

Here is a short video showing the issue:

https://github.com/user-attachments/assets/094ab2a6-57b2-4757-ac27-c5e9dcec3d7a

dbranley:
Closes #2386

I was able to find the root cause of this bug. The issue is in the definition of the `ContextualMenu` component that is used to display the column filtering options. The bug is in the `isFilterChecked` function in `ContextualMenu`:

```typescript
  ContextualMenu = ({ onFilterChange, col, listProps, selectedFilters, setFiltersInBulk }: ContextualMenuProps) => {
    const
      isFilterChecked = (data: S, key: S) => !!selectedFilters && selectedFilters[data]?.includes(key),
      [menuFilters, setMenuFilters] = React.useState(col.cellType?.tag
        ? Array.from(listProps.items.reduce((_filters, { key, text, data }) => {
          key.split(',').forEach(key => _filters.set(key, { key, text, data, checked: isFilterChecked(data, key) }))
          return _filters
        }, new Map<S, Fluent.IContextualMenuItem>()).values())
        : listProps.items.map(i => ({ ...i, checked: isFilterChecked(i.data, i.key) }))
      ),
```

If the `selectedFilters` prop is not null, but does NOT contain an entry for the current column, then the `isFilteredChecked` function will return `undefined`. This eventually becomes the value of the `checked` prop where `Fluent.Checkbox` is defined. As a result, eventually changing that to `true` when checked, doesn't get recognized and therefore looks like nothing has changed.

The solution is to modify the `isFilteredChecked` function to explicitly return either `true` or `false`. Here's a code snipit for that fix:

```typescript
      isFilterChecked = (data: S, key: S) => !!selectedFilters && selectedFilters[data]?.includes(key) ? true : false,
``` 

With this change, the `checked` prop is always set correctly, and the issue is resolved. 

Here's a short video showing the fix:

https://github.com/user-attachments/assets/668b17e1-7000-4160-8267-1c60c01af30d

I re-ran all the existing unit tests and they still pass:

Issue2386_ui_tests_after_fix

### New unit test

I have been unable to create a new test for this issue. I've worked on this for several days and cannot find a solution. I suspect this might be an issue with @testing-library/react, but can't be sure. I did try upgrading the library to the latest version of 12.x, which is 12.1.5, but no luck. I'm unable to upgrade any higher since there is a dependency on React 18 in more recent versions.

Let me try to explain the problem I'm having: When the filter icon is clicked, the `ContextualMenu` component defined in `table.tsx` is rendered. On this component, there is a list of checkboxes based on the `menuFilters` state variable defined in the component. The problem is that from the unit test, when a filter icon is clicked for a different column, the `ContextualMenu` component is initialized with the `menuFilters` from the 1st column where the filter icon was clicked. I have verified this by debugging the code in `table.tsx` and in the unit test. I have also proven to myself that this problem never happens when you do the same thing in the browser, it only happens when I do this from the unit test.

And I'm fairly confident that my unit test doesn't have an issue, at least as far as I could figure out. I even tried adding an await thinking that it might be some timing issue waiting for the `menuFilters` state variable to be updated, but nothing works. I put a lot of debug statements in the test and in `table.tsx` and am confident that the code in there is all fine.

Here is a snipit of the test that I wrote to illustrate the problem:

```typescript
    it('Filters correctly - multiple filters with select all', () => {
      tableProps = {
        ...tableProps,
        columns: [
          { name: 'colname1', label: 'col1', searchable: true },
          { name: 'colname2', label: 'col2', filterable: true },
          { name: 'colname3', label: 'col3', filterable: true },
        ],
        rows: [
          { name: 'rowname1', cells: [cell11, 'col2-val2', 'On'] },
          { name: 'rowname2', cells: [cell21, 'col2-val1', 'Off'] },
          { name: 'rowname3', cells: [cell31, 'col2-val3', 'On'] },     
        ]
      }
      const { container, getByLabelText, getAllByRole, queryAllByTestId, queryAllByLabelText, getByText, getByTestId } = render(<XTable model={tableProps} />)

      expect(getAllByRole('row')).toHaveLength(tableProps.rows!.length + headerRow)

      //Make a selection on the filter in the 2nd column
      console.log("Unit test - about to select filter for 'col2'")
      fireEvent.click(container.querySelectorAll(filterSelectorName)[0]!)
      fireEvent.click(getByLabelText('col2-val3'))      
      expect(getAllByRole('row')).toHaveLength(1 + headerRow)
      expect(queryAllByTestId('filter-count')[0]).toHaveTextContent('1')
      expect(getAllByRole('checkbox', {checked: true})).toHaveLength(1)

      //Now click somewhere so ContextualMenu is dismissed
      console.log("Unit test - about to click the search field, but don't actually search for anything - trying to force onDismiss() for context menu")
      fireEvent.click(getByTestId('search'))

      //Now select the filter on the 3rd column
      console.log("Unit test - about to select filter for 'col3'")
      fireEvent.click(container.querySelectorAll(filterSelectorName)[1]!)

      //But notice the menu filter options on the ContextualMenu component contain options from column 2, not column 3
      //This should give lenth of 1 since there should be a checkbox label of 'Off' - but it gives length of 0 
      console.log("Unit test - number of 'Off' labels is ="+queryAllByLabelText('Off').length+"=, but should be 1 for col3")
      //This should give lenth of 0 since there should be NO checkbox label of 'col2-val1' - but it gives length of 1 
      console.log("Unit test - number of 'col2-val1' labels is ="+queryAllByLabelText('col2-val1').length+"=, but should be 0 for col3")

      console.log("Unit test - about to click 'Select All'")
      fireEvent.click(getByText('Select All'))
      console.log("Unit test - number of 'Off' labels is ="+queryAllByLabelText('Off').length+"=, but should be 1 for col3")
      console.log("Unit test - number of 'col2-val1' labels is ="+queryAllByLabelText('col2-val1').length+"=, but should be 0 for col3")
    })  
```

And here's a snipit from `table.tsx` where you can see that I've added console.log() to print out the `listProps` at the top and then `menuFilters` state variable after initialization.

```typescript
  ContextualMenu = ({ onFilterChange, col, listProps, selectedFilters, setFiltersInBulk }: ContextualMenuProps) => {
    console.log("ContextualMenu - at top for col.name="+col.name+"=, listProps is:")
    console.log(listProps)
    
    const
      isFilterChecked = (data: S, key: S) => !!selectedFilters && selectedFilters[data]?.includes(key) ? true : false,
      [menuFilters, setMenuFilters] = React.useState(col.cellType?.tag
        ? Array.from(listProps.items.reduce((_filters, { key, text, data }) => {
          key.split(',').forEach(key => _filters.set(key, { key, text, data, checked: isFilterChecked(data, key) }))
          return _filters
        }, new Map<S, Fluent.IContextualMenuItem>()).values())
        : listProps.items.map(i => ({ ...i, checked: isFilterChecked(i.data, i.key) }))
      ),
      selectAll = () => {
        setMenuFilters(menuFilters.map(i => ({ ...i, checked: true })))
        setFiltersInBulk(col.key, menuFilters.map(f => f.key))
      },
      deselectAll = () => {
        setMenuFilters(menuFilters.map(i => ({ ...i, checked: false })))
        setFiltersInBulk(col.key, [])
      },
      getOnFilterChangeHandler = (data: S, key: S) => (_ev?: React.FormEvent<HTMLInputElement | HTMLElement>, checked?: B) => {
        setMenuFilters(filters => filters.map(f => f.key === key ? ({ ...f, checked }) : f))
        onFilterChange(data, key, checked)
      }
    console.log("ContextualMenu - after initialization for col.name="+col.name+"=, menuFilters is:")
    console.log(menuFilters)

    return (
      <div style={{ padding: 10 }}>
        <Fluent.Text variant='mediumPlus' styles={{ root: { paddingTop: 10, paddingBottom: 10, fontWeight: 'bold' } }} block>Show only</Fluent.Text>
```

And here is the output after running the test:

````
  console.log
    Unit test - about to select filter for 'col2'

      at Object.log (src/table.test.tsx:1035:15)

  console.log
    ContextualMenu - at top for col.name=col2=, listProps is:

      at log (src/table.tsx:296:13)

  console.log
    {
      ariaLabel: undefined,
      items: [
        { key: 'col2-val2', text: 'col2-val2', data: 'colname2' },
        { key: 'col2-val1', text: 'col2-val1', data: 'colname2' },
        { key: 'col2-val3', text: 'col2-val3', data: 'colname2' }
      ],
      totalItemCount: 3,
      hasCheckmarks: false,
      hasIcons: false,
      defaultMenuItemRenderer: [Function: defaultMenuItemRenderer],
      labelElementId: undefined
    }

      at log (src/table.tsx:297:13)

  console.log
    ContextualMenu - after initialization for col.name=col2=, menuFilters is:

      at log (src/table.tsx:320:13)

  console.log
    [
      {
        key: 'col2-val2',
        text: 'col2-val2',
        data: 'colname2',
        checked: false
      },
      {
        key: 'col2-val1',
        text: 'col2-val1',
        data: 'colname2',
        checked: false
      },
      {
        key: 'col2-val3',
        text: 'col2-val3',
        data: 'colname2',
        checked: false
      }
    ]

      at log (src/table.tsx:321:13)

  console.log
    ContextualMenu - at top for col.name=col2=, listProps is:

      at log (src/table.tsx:296:13)

  console.log
    {
      ariaLabel: undefined,
      items: [
        { key: 'col2-val2', text: 'col2-val2', data: 'colname2' },
        { key: 'col2-val1', text: 'col2-val1', data: 'colname2' },
        { key: 'col2-val3', text: 'col2-val3', data: 'colname2' }
      ],
      totalItemCount: 3,
      hasCheckmarks: false,
      hasIcons: false,
      defaultMenuItemRenderer: [Function: defaultMenuItemRenderer],
      labelElementId: undefined
    }

      at log (src/table.tsx:297:13)

  console.log
    ContextualMenu - after initialization for col.name=col2=, menuFilters is:

      at log (src/table.tsx:320:13)

  console.log
    [
      {
        key: 'col2-val2',
        text: 'col2-val2',
        data: 'colname2',
        checked: false
      },
      {
        key: 'col2-val1',
        text: 'col2-val1',
        data: 'colname2',
        checked: false
      },
      {
        key: 'col2-val3',
        text: 'col2-val3',
        data: 'colname2',
        checked: true
      }
    ]

      at log (src/table.tsx:321:13)

  console.log
    Unit test - about to click the search field, but don't actually search for anything - trying to force onDismiss() for context menu

      at Object.log (src/table.test.tsx:1043:15)

  console.log
    Unit test - about to select filter for 'col3'

      at Object.log (src/table.test.tsx:1047:15)

  console.log
    ContextualMenu - at top for col.name=col3=, listProps is:

      at log (src/table.tsx:296:13)

  console.log
    {
      ariaLabel: undefined,
      items: [
        { key: 'On', text: 'On', data: 'colname3' },
        { key: 'Off', text: 'Off', data: 'colname3' }
      ],
      totalItemCount: 2,
      hasCheckmarks: false,
      hasIcons: false,
      defaultMenuItemRenderer: [Function: defaultMenuItemRenderer],
      labelElementId: undefined
    }

      at log (src/table.tsx:297:13)

  console.log
    ContextualMenu - after initialization for col.name=col3=, menuFilters is:

      at log (src/table.tsx:320:13)

  console.log
    [
      {
        key: 'col2-val2',
        text: 'col2-val2',
        data: 'colname2',
        checked: false
      },
      {
        key: 'col2-val1',
        text: 'col2-val1',
        data: 'colname2',
        checked: false
      },
      {
        key: 'col2-val3',
        text: 'col2-val3',
        data: 'colname2',
        checked: true
      }
    ]

      at log (src/table.tsx:321:13)

  console.log
    Unit test - number of 'Off' labels is =0=, but should be 1 for col3

      at Object.log (src/table.test.tsx:1052:15)

  console.log
    Unit test - number of 'col2-val1' labels is =1=, but should be 0 for col3

      at Object.log (src/table.test.tsx:1054:15)

  console.log
    Unit test - about to click 'Select All'

      at Object.log (src/table.test.tsx:1056:15)

  console.log
    ContextualMenu - at top for col.name=col3=, listProps is:

      at log (src/table.tsx:296:13)

  console.log
    {
      ariaLabel: undefined,
      items: [
        { key: 'On', text: 'On', data: 'colname3' },
        { key: 'Off', text: 'Off', data: 'colname3' }
      ],
      totalItemCount: 2,
      hasCheckmarks: false,
      hasIcons: false,
      defaultMenuItemRenderer: [Function: defaultMenuItemRenderer],
      labelElementId: undefined
    }

      at log (src/table.tsx:297:13)

  console.log
    ContextualMenu - after initialization for col.name=col3=, menuFilters is:

      at log (src/table.tsx:320:13)

  console.log
    [
      {
        key: 'col2-val2',
        text: 'col2-val2',
        data: 'colname2',
        checked: true
      },
      {
        key: 'col2-val1',
        text: 'col2-val1',
        data: 'colname2',
        checked: true
      },
      {
        key: 'col2-val3',
        text: 'col2-val3',
        data: 'colname2',
        checked: true
      }
    ]

      at log (src/table.tsx:321:13)

  console.log
    Unit test - number of 'Off' labels is =0=, but should be 1 for col3

      at Object.log (src/table.test.tsx:1058:15)

  console.log
    Unit test - number of 'col2-val1' labels is =1=, but should be 0 for col3

      at Object.log (src/table.test.tsx:1059:15)
````

If you look at the debug statements carefully, you'll notice that when the 2nd filter icon is clicked for col3, it is getting passed the correct data in the `listProps` prop, but the `menuFilters` state variable is initialized for the col2 values.  

Based on the debugging I've done, I think this is happening because the `onDismiss()` on the context menu is never called when running from the unit test. This gets called fine from the UI, but from the unit test `onDismiss()` isn't called, hence the context menu doesn't get initialized properly when it is opened for a different column. I even added a step in the test to click into the Search field, hoping that might trigger `onDismiss()` on the context menu, but that didn't help.

Also, I confirmed that clicking the `Select All` button anyway doesn't work as expected since it ends up selecting menu filter values which don't actually exist.

I hope this all makes sense. And if I'm overlooking something in the unit test or if you see an issue with the fix, please let me know. Thanks!

### DIFF
--- a/ui/src/table.tsx
+++ b/ui/src/table.tsx
@@ -294,7 +294,7 @@ const
   },
   ContextualMenu = ({ onFilterChange, col, listProps, selectedFilters, setFiltersInBulk }: ContextualMenuProps) => {
     const
-      isFilterChecked = (data: S, key: S) => !!selectedFilters && selectedFilters[data]?.includes(key),
+      isFilterChecked = (data: S, key: S) => !!selectedFilters && selectedFilters[data]?.includes(key) ? true : false,
       [menuFilters, setMenuFilters] = React.useState(col.cellType?.tag
         ? Array.from(listProps.items.reduce((_filters, { key, text, data }) => {
           key.split(',').forEach(key => _filters.set(key, { key, text, data, checked: isFilterChecked(data, key) }))
--- a/ui/src/table.tsx
+++ b/ui/src/table.tsx
@@ -294,7 +294,7 @@ const
   },
   ContextualMenu = ({ onFilterChange, col, listProps, selectedFilters, setFiltersInBulk }: ContextualMenuProps) => {
     const
-      isFilterChecked = (data: S, key: S) => !!selectedFilters && selectedFilters[data]?.includes(key) ? true : false,
+      isFilterChecked = (data: S, key: S) => !!selectedFilters && !!selectedFilters[data]?.includes(key),
       [menuFilters, setMenuFilters] = React.useState(col.cellType?.tag
         ? Array.from(listProps.items.reduce((_filters, { key, text, data }) => {
           key.split(',').forEach(key => _filters.set(key, { key, text, data, checked: isFilterChecked(data, key) }))
--- a/ui/src/table.test.tsx
+++ b/ui/src/table.test.tsx
@@ -16,6 +16,7 @@ import { fireEvent, render } from '@testing-library/react'
 import React from 'react'
 import { Table, XTable } from './table'
 import { wave } from './ui'
+import { KeyCodes } from '@fluentui/react'
 
 const
   name = 'table',
@@ -1012,7 +1013,7 @@ describe('Table.tsx', () => {
       expect(getAllByRole('row')).toHaveLength(2 + headerRow)
     })
 
-    it('Select All on 2nd filter selects all filter checkboxes', () => {
+    it('Select All on 2nd filter selects all filter checkboxes', async () => {
       tableProps = {
         ...tableProps,
         columns: [
@@ -1023,31 +1024,29 @@ describe('Table.tsx', () => {
         rows: [
           { name: 'rowname1', cells: [cell11, 'col2-val2', 'On'] },
           { name: 'rowname2', cells: [cell21, 'col2-val1', 'Off'] },
-          { name: 'rowname3', cells: [cell31, 'col2-val3', 'On'] },     
+          { name: 'rowname3', cells: [cell31, 'col2-val3', 'On'] },
         ]
       }
-      const { container, getByLabelText, getAllByRole, queryAllByTestId, getByText } = render(<XTable model={tableProps} />)
+      const { container, getByLabelText, getAllByRole, getByText, queryByText } = render(<XTable model={tableProps} />)
 
-      expect(getAllByRole('row')).toHaveLength(tableProps.rows!.length + headerRow)
-
-      //Make a selection on the filter in the 2nd column
       fireEvent.click(container.querySelectorAll(filterSelectorName)[0]!)
-      fireEvent.click(getByLabelText('col2-val3'))      
-      expect(getAllByRole('row')).toHaveLength(1 + headerRow)
-      expect(queryAllByTestId('filter-count')[0]).toHaveTextContent('1')
-      expect(getAllByRole('checkbox', {checked: true})).toHaveLength(1)
+      fireEvent.click(getByLabelText('col2-val3'))
+      expect(getAllByRole('checkbox', { checked: true })).toHaveLength(1)
 
-      //Now select the filter on the 3rd column and then click the 'Select All' button
+      // FluentUI uses a deprecated 'which' property instead of the 'key' prop.
+      // Close the menu with escape (does not close when clicking other menu in test).
+      fireEvent.keyDown(window, { which: KeyCodes.escape })
+      expect(queryByText('Show only')).not.toBeInTheDocument()
+
       fireEvent.click(container.querySelectorAll(filterSelectorName)[1]!)
       fireEvent.click(getByText('Select All'))
-
-      //Make sure all the checkboxes available are now checked
-      expect(getAllByRole('checkbox', {checked: true})).toHaveLength(getAllByRole('checkbox').length)
-    })    
+      expect(queryByText('Show only')).toBeInTheDocument()
+      expect(getAllByRole('checkbox', { checked: true })).toHaveLength(getAllByRole('checkbox').length)
+    })
 
     it('Fires event when pagination enabled', () => {
       const { container, getAllByText } = render(<XTable model={{ ...tableProps, pagination: { total_rows: 10, rows_per_page: 5 }, events: ['filter'] }} />)
-      
+
       fireEvent.click(container.querySelector(filterSelectorName) as HTMLDivElement)
       fireEvent.click(getAllByText('1')[3].parentElement as HTMLDivElement)
 
@@ -1087,7 +1086,7 @@ describe('Table.tsx', () => {
       fireEvent.click(getByLabelText('2'))
       expect(getAllByRole('row')).toHaveLength(2 + headerRow)
       expect(queryByTestId('filter-count')).toHaveTextContent('2')
-    })    
+    })
 
     it('Filter counts - manual deselect', () => {
       const { container, getAllByRole, getByLabelText, queryByTestId } = render(<XTable model={tableProps} />)
@@ -1109,25 +1108,25 @@ describe('Table.tsx', () => {
       fireEvent.click(getByLabelText('2'))
       expect(getAllByRole('row')).toHaveLength(tableProps.rows!.length + headerRow)
       expect(queryByTestId('filter-count')).toBeNull
-    })    
+    })
 
     it('Filter counts - select all', () => {
       const { container, getAllByRole, queryByTestId, getByText } = render(<XTable model={tableProps} />)
 
       expect(getAllByRole('row')).toHaveLength(tableProps.rows!.length + headerRow)
       fireEvent.click(container.querySelector(filterSelectorName)!)
-      
+
       fireEvent.click(getByText('Select All'))
       expect(getAllByRole('row')).toHaveLength(tableProps.rows!.length + headerRow)
       expect(queryByTestId('filter-count')).toHaveTextContent(tableProps.rows!.length.toString())
-    })  
+    })
 
     it('Filter counts - deselect all', () => {
       const { container, getAllByRole, queryByTestId, getByText } = render(<XTable model={tableProps} />)
 
       expect(getAllByRole('row')).toHaveLength(tableProps.rows!.length + headerRow)
       fireEvent.click(container.querySelector(filterSelectorName)!)
-      
+
       fireEvent.click(getByText('Select All'))
       expect(getAllByRole('row')).toHaveLength(tableProps.rows!.length + headerRow)
       expect(queryByTestId('filter-count')).toHaveTextContent(tableProps.rows!.length.toString())
@@ -1159,7 +1158,7 @@ describe('Table.tsx', () => {
 
       expect(getAllByRole('row')).toHaveLength(tableProps.rows!.length + headerRow)
       fireEvent.click(container.querySelector(filterSelectorName)!)
-      
+
       fireEvent.click(getByText('Select All'))
       expect(getAllByRole('row')).toHaveLength(tableProps.rows!.length + headerRow)
       expect(queryByTestId('filter-count')).toHaveTextContent('9+')
@@ -1167,7 +1166,7 @@ describe('Table.tsx', () => {
       fireEvent.click(getByText('Deselect All'))
       expect(getAllByRole('row')).toHaveLength(tableProps.rows!.length + headerRow)
       expect(queryByTestId('filter-count')).toBeNull
-    })     
+    })
 
     it('Filter counts - clear selection with reset', () => {
       const { container, getAllByRole, getByLabelText, queryByTestId, getByText } = render(<XTable model={{ ...tableProps, resettable: true }} />)
@@ -1333,7 +1332,7 @@ describe('Table.tsx', () => {
     })
   })
 
-  
+
   describe('Group by', () => {
     beforeEach(() => {
       tableProps = {
--- a/ui/src/table.test.tsx
+++ b/ui/src/table.test.tsx
@@ -1013,7 +1013,7 @@ describe('Table.tsx', () => {
       expect(getAllByRole('row')).toHaveLength(2 + headerRow)
     })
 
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: ui/src/table.tsx [hunk 1]
```diff
@@ -294,7 +294,7 @@ const
   },
   ContextualMenu = ({ onFilterChange, col, listProps, selectedFilters, setFiltersInBulk }: ContextualMenuProps) => {
     const
-      isFilterChecked = (data: S, key: S) => !!selectedFilters && selectedFilters[data]?.includes(key),
+      isFilterChecked = (data: S, key: S) => !!selectedFilters && selectedFilters[data]?.includes(key) ? true : false,
```
mturoci, line 297: Small stylistic update. `!!` is a boolean conversion which turns `undefined` to `false` since it's a falsy value.

```suggestion
      isFilterChecked = (data: S, key: S) => !!selectedFilters && !!selectedFilters[data]?.includes(key),
```

mturoci (changes requested):
Great catch @dbranley! The fix looks reasonable to me. Can you please also add a unit test verifying this behavior so that we can be sure we never run into this in future?

ui/src/table.tsx
dbranley: Done! I generally lean towards explicit return of true/false, but since the first part of this AND statement uses a boolean conversion then I can see that it makes sense to use it for both sides of the AND. Thanks for the suggestion!

dbranley:
@mturoci I actually spent several days trying to create a meaningful unit test for this issue. I left a comment about this in the original PR yesterday, along with snipits from the test code and snipits from `table.tsx` where I added console.log() debug statements. You'll find this at the bottom of the original PR comment. 

Based on the debugging I've done, I'm pretty sure the test isn't working because `onDismiss()` for the context menu doesn't get called when running from a unit test. As a result, I think this is leaving state values from previous instances of the `ContextualMenu` component. So this means in the unit test, the filter selection list on the popup is actually values from the first column that was selected. 

I did go ahead and add a new test anyway, which just makes sure that all the checkboxes are selected after clicking `Select All`, but I'm not sure the test adds much value since these are selections for the wrong column. As a result, the row count will always be 0 + 1 for the header.

Let me know if you think the test should be removed or if I'm overlooking something in the post from yesterday about the issues with the unit test. Thanks!

mturoci (approved):
> I left a comment about this in the original PR yesterday

Oh didn't notice, sorry about that.

I see. I updated your test to make it behave the way we want: fail without your fix and succeed with it. You were right, for some reason, the first context menu wouldn't close within the test, despite clicking on the second already. I managed to close it by emulating hitting `Esc` key, inspired by @marek-mihok's [work](https://github.com/h2oai/wave/blob/927d6d695f96459b5a2f917bcf719cfcee04fdae/ui/src/menu.test.tsx#L71). Didn't dig deeper though.

Thanks for all the hard work @dbranley! Merging it in.

dbranley:
@mturoci That's awesome that you found a way to get past this from the test! I tried just about everything, but didn't think about simulating the `Esc` key. 

I know the PR is merged now, but I'm still curious to get to root cause of the problem with the menu not going away. I think I found the issue for this in FluentUI and it was resolved late last year. Tomorrow I'm going to replicate the problem locally without Wave and then upgrade to the latest FluentUI to see if this fix resolved it. I'll let you know how that goes. Like I said, I know this isn't necessary, I'm just curious.

mturoci:
>  I know this isn't necessary, I'm just curious.

Love the passion :) Feel free to make a PR to update the test in case you find the root cause.

dbranley:
@mturoci So it looks like the solution I had in mind is only available in Fluent v9, so not applicable to Wave. I've probably spent enough time in this rabbit hole, so I'm going to leave it there. On to something else now!

PR #2390, opened 2024-09-12
fix: Conform standalone button width to its container #2361

### LINKED ISSUE #2361: Regression from earlier version, inline Buttons are misplaced
#### Wave SDK Version, OS

Wave v1.3.4

#### Actual behavior

From wave v1.0.0 to wave v1.3.4 the placement of buttons has regressed.

```
ui.inline(
        ui.button(
            name="b1",
            label="Button1",
            primary=False,
            width="150px",
        ),
        ui.button(
            name="b2",
            label="Button2",
            primary=False,
            width="150px",
        ),
        ui.button(
            name="b3",
            label="Button3",
            primary=True,
            width="200px",
        ),
        ui.button(
            name="b4",
            label="Button4",
            primary=False,
            width="200px",
        ),
        justify="start",
)
```

v1.3.4
image

#### Expected behavior

v1.0.0
image

marek-mihok:

____
The idea of this fix is to make the standalone button together with its left margin to fit its container width. This is by specifying `display: 'flex` for the button wrapper. 

**Example:**
When the standalone button with has specified width of `150px` is followed by another standalone button with the same width, the first button will be 150px wide but the second one and every other will be 142px wide as it accounts for 8px left padding. This marginal difference can only be caught by a perfectionist eye. This is demonstrated in the recording below:

https://github.com/user-attachments/assets/019abfb7-5192-463e-9d49-db65f4d459fe

The `alignItems: 'center'` is specified not to break existing button stylings.

The solution was manually and visual regression tested.

Closes #2361

### DIFF
--- a/ui/src/button.tsx
+++ b/ui/src/button.tsx
@@ -108,6 +108,10 @@ const
       boxSizing: 'border-box',
       overflowX: 'auto'
     },
+    standaloneButton: {
+      display: 'flex',
+      alignItems: 'center',
+    }
   }),
   justifications: Dict<Fluent.Alignment> = {
     start: 'start',
@@ -201,7 +205,7 @@ export const
     )
   },
   XStandAloneButton = ({ model: m }: { model: Button }) => (
-    <div className={css.buttons}>
+    <div className={css.standaloneButton}>
       <XButton key={m.name} model={m} />
     </div>
   ),


mturoci (approved):
LGTM, thanks @marek-mihok!

PR #2402, opened 2024-09-30
Emit event on table group open/collapse #2133

### LINKED ISSUE #2133: Emit event on table group open/collapse

The problem is described in this [discussion.](https://github.com/h2oai/wave/discussions/2131)

I'd like to be able to update the table groups and rows without resetting the collapsed/expanded state of the table groups.

dbranley:

Closes #2133

As requested in the ticket, I have made changes so that the table component will emit an event called `group_change` whenever a group is collapsed or opened. The `group_change` event data sent is a list of the group names that had their collapse state changed. The expectation is that the client py code will use this as a toggle to keep track of the collapsed states when the table has to be updated.  

The change for this is in by both the `onToggleCollapseAll` and `onToggleCollapse` functions in `table.tsx`.

But before changing this, I had to understand `expandedRefs.current`. When set to `null`, it means all the groups are currently open. And when it is not null, then it means all the groups are collapsed, except for the groups whose names are in the `expandedRefs.current` dictionary with a value of `false`. So if you imagine a table with 2 groups called 'Bob' and 'John', here is what various definitions of `expandedRefs.current` look like:

- `null` - all groups open
- `{'Bob': false}` - 'Bob' is open and 'John' is collapsed
- `{}` - all groups collapsed

Here's an explanation of the logic changes in the 2 functions:

**onToggleCollapseAll:**

When collapsing all the groups:

- if `expandedRefs.current` is `null`, then it means all the groups are now open. Therefore, all the group names will be included in the event.
- if `expandedRefs.current` is not `null`, then it means only the groups in the `expandedRefs.current` dict are open. Therefore, the group names in the dict will be included in the event.

When opening all the groups:

- this can only happen when all the groups are collapsed. Therefore, all the group names will be included in the event.

**onToggleCollapse:**

The event from here will always just be the group name whose collapse state is changing.

While testing this, I found a bug which I had to fix. The issue was in the else-block in the code that was updating `expandedRefs.current`:

```typescript
        if (expandedRefs.current) {
          isCollapsed
            ? expandedRefs.current[key] = false
            : delete expandedRefs.current[key]
        } else {
          expandedRefs.current = { [key]: false }
        }
```

As I mentioned above, when `expandedRefs.current` is `null`, that means all the groups are currently open and the user must be closing the current group. But the code in the else-block is setting `expandedRefs.current` to `{ 'group_name': false }`, which is wrong. In order to properly reflect that 1 group is collapsed while all the others are open, the dict needs to contain all those other groups with a value of `false`. I made that fix and also added a new unit test that replicated the steps to reproduce the bug - `Collapses all groups when some already collapsed - fire event`.

I updated the documentation of the `events` prop to mention the new event type available to listen for. I also updated the text a bit since before it suggested that all events depended on pagination being set, but both `select` and now `group_change` are sent regardless of pagination.

```typescript
  /** The events to capture on this table. When pagination is set, one of 'search' | 'sort' | 'filter' | 'download' | 'page_change' | 'reset'. These events are available regardless of pagination: 'select' | 'group_change'. */
  events?: S[]
```

I then ran `make generate` to update/generate the corresponding py code to reflect the prop change.

I created several new unit tests in `table.test.tsx` to test this change both with predefined groups and also when group by is used. I also ran the entire test suite and everything passes fine:

issue2133_unit-tests

I also created a new py example called `table_events_group.py` to test the changes. It has a table with 2 predefined groups and is listening on the new `group_change` event. The events are used to maintain the collapse state for the groups so that the collapsed state is maintained when the table is updated. There is also a button to kick off a background task to add new rows to the table. Using this, you can see how the collapsed state is maintained as new rows are added to the table. Here's a short video of the new example:

https://github.com/user-attachments/assets/103d401e-eff9-414b-a7f1-3c0e7876a528

I decided to include this new example in the PR and you can see it is reflected in my local build of the wave website:

issue2133_doc_example

I also locally modified the existing py table example to listen on the new event so that I can test it working for `Group By` groups. I did not include this in the PR. Here's a short video of that:

https://github.com/user-attachments/assets/cb135e07-3ca0-4678-9c70-cb4d927b5c04

Finally, I updated the `table.md` widget documentation to mention the new event. I added a this to the existing section called `With collapsed groups`. You can see a screen shot of that here:

issue2133_doc_widget


ui/src/table.tsx
```diff
@@ -491,14 +491,39 @@ const
...
+          let changingGroups 
+          if (isAllCollapsed){
+            if (expandedRefs.current && Object.keys(expandedRefs.current).length > 0){
+              changingGroups = Object.keys(expandedRefs.current)
+            } else {
+              changingGroups = groups?.map(group => group.name)
+            }
+          } else {
+            changingGroups = groups?.map(group => group.name)
+          }
```
mturoci, line 505: Would this work as well?

```suggestion
          const changedGroups =
            isAllCollapsed && expandedRefs.current && Object.keys(expandedRefs.current).length > 0
              ? Object.keys(expandedRefs.current)
              : groups?.map(group => group.name)
```

ui/src/table.test.tsx
```diff
@@ -1676,6 +1676,58 @@ describe('Table.tsx', () => {
       expect(container.querySelectorAll('.ms-GroupHeader-title')[0]).toHaveTextContent('1/20/1970, 4:58:47 AM(0)')
       expect(container.querySelectorAll('.ms-GroupHeader-title')[1]).toHaveTextContent('6/22/2022, 8:47:51 PM(1)')
     })
+
+    it('Collapses all group by list - fire event', () => {
```
mturoci, line 1680: Why do we need "duplicate" tests? Wouldn't the ones in the "Groups" suite be enough?

website/widgets/form/table.md
```diff
@@ -401,7 +401,7 @@ q.page['example'] = ui.form_card(box='1 1 3 4', items=[
 
 ### With collapsed groups
 
-Groups are shown in a collapsed state by default. With the `collapsed` attribute you can change this behavior.
+Groups are shown in a collapsed state by default. With the `collapsed` attribute you can change this behavior. You also can keep track of the collapsed states by registering a `'group_change'` [event](/docs/examples/table-events-group) (populated in `q.events`). This is useful when needing to refresh the table.
```
mturoci, line 404:
```suggestion
Groups are shown in a collapsed state by default. With the `collapsed` attribute you can change this behavior. You can also keep track of the collapsed states by registering a `'group_change'` [event](/docs/examples/table-events-group) (populated in `q.events`). This is useful when needing to refresh the table and persist collapsed state.
```

py/examples/table_events_group.py
```diff
@@ -0,0 +1,114 @@
...
+    if (issue_cnt % 2) == 0:
+        bobrows.append({"name":"row"+str(issue_cnt), "cell":"Issue"+str(issue_cnt)})
+    else:
+        johnrows.append({"name":"row"+str(issue_cnt), "cell":"Issue"+str(issue_cnt)})   
+    update_table_groups(q)
+    await q.page.save()       
+
+
+@app('/demo')
+async def serve(q: Q):
```
mturoci, line 49: Love this example, but it might be too complex. Let's strip it down to a bare minimum - registering group change event and displaying the currently changed groups (ie without background tasks adding new rows).

mturoci (changes requested):
Phenomenal job @dbranley! Looks good overall, just a few small comments.

ui/src/table.tsx
dbranley: Done. Thanks for the suggestion - I'm usually lean towards explicit if-else blocks, but what you have is definitely cleaner and more concise!

ui/src/table.test.tsx
dbranley: I thought these would be useful since the way to end up with these groups are different - explicitly defined vs group by. Even though this new 'group_change' event code is the same regardless, since the steps getting there are different it seemed worth having the tests to protect from future changes that might result in this breaking only for 'group by' - not sure what that would be now, but seems useful to protect against that. 

If you disagree, I can definitely remove the 'group by' tests, so just let me know!

website/widgets/form/table.md
dbranley: Done!

py/examples/table_events_group.py
dbranley: Yeah it is a complex example, but I was trying to replicate the original [discussion](https://github.com/h2oai/wave/discussions/2131) from when this ticket was originally created. It was certainly helpful for me when testing.

I've remove the bits related to adding new rows and updated the screenshot under ./website/docs/examples/assets.

dbranley:
@mturoci Thanks for the great feedback! I've made all the changes you suggested, although I kept the unit test as-is for now. See my in-line response to your suggestion and let me know what you think. 

I re-ran all the unit tests and everything still passes. I then re-tested with the new example and all good. I also rebuilt the docs and reviewed those too. 

Thanks!

PR #2460, opened 2025-05-14
security: Allow setting stricter CSP.

mturoci:
* All inline scripts/styles include a per-response generated `nonce`, which is injected into CSP header if specified.
* Replaced plot templating lib handlebars with mustache which is slower, but doesn't require any inlining.
* npm audit fix for a couple of mid CVEs


web_server.go
```diff
@@ -58,20 +60,55 @@ func newWebServer(
...
+		return "", fmt.Errorf("failed reading default index.html page: %v", err)
+	}
+	return base64.StdEncoding.EncodeToString(nonce), nil
+}
+
+func mungeIndexPage(baseURL, html, nonce string) (string, error) {
+	// HACK: Set base URL as a body tag attribute, to be used by the front-end for deducing hash-routing and websocket addresses.
+	html = strings.Replace(html, "<body", `<body data-base-url="`+baseURL+`" data-nonce="`+nonce+`"`, 1)
 	// ./wave-static/a/b/c.d -> /base-url/wave-static/a/b/c.d
 	html = strings.ReplaceAll(html, `="./wave-static/`, `="`+baseURL+"wave-static/")
```
marek-mihok, line 108: Once the UI is built, there is a `<script type="module" crossorigin src="./wave-static/index-658bc908.js"></script>` added into the `<header>` of `/wave/ui/build/index.html` file, which does not contain `nonce` attribute. Isn't this a possible vulnerability?

```js
<!doctype html>
<html lang="en">

<head>
...
  <title>H2O Wave</title>
  <script type="module" crossorigin src="./wave-static/index-658bc908.js"></script>
...
</head>
```

ui/index.html
```diff
@@ -9,23 +9,32 @@
...
+    const originalAppendChild = Element.prototype.appendChild
+    window.CSPSettings = { nonce }
+    Element.prototype.insertBefore = function (newNode, referenceNode) {
+      if (nonce && newNode.nodeType === 1 && newNode.hasAttribute("data-merge-styles")) {
+        newNode.setAttribute("nonce", nonce)
+      }
+      return originalInsertBefore.call(this, newNode, referenceNode)
+    }
+    Element.prototype.appendChild = function (newNode) {
+      if (nonce && newNode.nodeType === 1 && newNode.nodeName === 'SCRIPT' || newNode.nodeName === 'STYLE') {
```
marek-mihok, line 32:
```suggestion
      if (nonce && newNode.nodeType === 1 && (newNode.nodeName === 'SCRIPT' || newNode.nodeName === 'STYLE')) {
```

Is the condition correct? It is evaluated like this
```js
if (nonce && newNode.nodeType === 1 && newNode.nodeName === 'SCRIPT') || newNode.nodeName === 'STYLE'
```
This means even if `nonce` is false, there will be nonce assigned to <style> elements anyway. Adding the braces as suggested would fix that.

marek-mihok (commented):
I was not able to find any issues/regressions in the functionality, however I have 2 concerns addressed in comments below.

web_server.go
mturoci: > Isn't this a possible vulnerability?

It isn't. Nonce is only needed for inline scipts since they are vulnerable to XSS. The one you show above is a regular 1st party script with a content that is already set.

marek-mihok: Makes sense. Thanks for clarification!

Issue #524, opened 2021-01-23
Table with height of 100% doenst show up in 0.11.0 layout 


tkmaker:
#### Wave SDK Version, OS

0.11.0 Wave. Using Chrome. 

I have a `ui.table` with `height='100%'` and using the new layout in 0.11.0. If I have a single card with the table and the 100% value specified, the table doesnt appear in the card. If I get rid of the `height` param altogether, the table shows up fine. 

#### Actual behavior

(A clear and concise description of what happened.)

#### Expected behavior

(A clear and concise description of what you expected to happen.)

#### Steps To Reproduce

1. Do this
2. Do that
3. Do something else

mturoci:
Hi @tkmaker,

CSSwise, this behavior is caused because percentage values for height work only if parent element has fixed height specified. In wave terms, this works for grid layout, because the direct parent of the table (card) has height specified (last `box` param), but it doesn't in responsive layout due to introduction of vertical/horizontal positioning at card level. You can overcome this by specifying `size` prop on `ui.zone` and setting table's `height` to 100%. This solution will not work in Safari though, as it has different height 100% implementation compared to normal browsers.

The general advice is to use fixed units like `px` instead. However, if your desired height is something general like taking up the whole screen, you can use `vh` units which stand for 'viewport height' - range from 0 to 100. In case you want to occupy whole screen and also account for card spacing, you can use `calc(100vh - 50px)` which allows you to calculate in a cross-unit way.

The code I used:
```py
# Table
# Use a #table to display tabular data.
# ---
import random

from faker import Faker

from h2o_wave import main, app, Q, ui

fake = Faker()

_id = 0


class Issue:
    def __init__(self, text: str, status: str, progress: float, icon: str, notifications: str, created: str):
        global _id
        _id += 1
        self.id = f'I{_id}'
        self.text = text
        self.status = status
        self.views = 0
        self.progress = progress
        self.icon = icon
        self.notifications = notifications
        self.created = created


# Create some issues
issues = [
    Issue(
        text=fake.sentence(),
        status=('Closed' if i % 2 == 0 else 'Open'),
        progress=random.random(),
        icon=('BoxCheckmarkSolid' if random.random() > 0.5 else 'BoxMultiplySolid'),
        notifications=('Off' if random.random() > 0.5 else 'On'),
        created=fake.iso8601()) for i in range(100)
]

# Create columns for our issue table.
columns = [
    ui.table_column(name='text', label='Issue', sortable=True, max_width='300'),
[...]

PR #2478, opened 2025-09-22
Layout.md Doc update for issue #524

zbanks23:

- [X] When resolving a specific issue, the PR description includes `Closes #xxx`, where "xxx" is the issue number.

Closes Issue #524

### DIFF
--- a/website/docs/layout.md
+++ b/website/docs/layout.md
@@ -100,9 +100,9 @@ q.page['tab_card'] = ui.tab_card(
 
 ![zone card shrink](assets/zone-card-shrink.png)
 
-### Percentage Width and Height
+### Width and Height
 
-By default, we encourage static units like `px` or `rem` for setting widths and heights. All units can be found [here](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Values_and_Units). In flex layouts, percent values such as `100%` only work if every ancestor has an explicit `size`, otherwise, the computed height can be zero and the content may not render as expected.
+By default, we encourage static units like `px` or `rem` for setting widths and heights (see [MDN](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Values_and_units#lengths) for more info). In flex layouts, percent values such as `100%` only work if every ancestor has an explicit `size`, otherwise, the computed height can be zero and the content may not render as expected.
 
 ### Responsive Layout
 


zbanks23:
I'm a first-time contributor for open-source projects in general. Let me know if I messed up in any way!

website/docs/layout.md
```diff
@@ -100,6 +100,10 @@ q.page['tab_card'] = ui.tab_card(
 
 ![zone card shrink](assets/zone-card-shrink.png)
 
+### Setting Width and Height
+
+By default, we encourage static units like `px` or `rem` for setting widths and heights. In flex layouts, percent values such as `100%` only work if every ancestor has an explicit size; otherwise, the computed height can be zero and the content may not render as expected.
```
mturoci, line 105: Would be good to also include an MDN link to all the available units.
```suggestion
By default, we encourage static units like `px` or `rem` for setting widths and heights. In flex layouts, percent values such as `100%` only work if every ancestor has an explicit `size`, otherwise, the computed height can be zero and the content may not render as expected.
```

website/docs/layout.md
```diff
@@ -100,6 +100,10 @@ q.page['tab_card'] = ui.tab_card(
 
 ![zone card shrink](assets/zone-card-shrink.png)
 
+### Setting Width and Height
```
mturoci, line 103:
```suggestion
### Percentage width and height
```

website/docs/layout.md
```diff
@@ -55,7 +55,7 @@ A breakpoint value of `xs` matches all viewport widths, unless other breakpoints
 
 ### Zones
 
-A *zone* is a named area inside the layout where cards can be placed. A zone is defined using `ui.zone()`.
+A _zone_ is a named area inside the layout where cards can be placed. A zone is defined using `ui.zone()`.
```
mturoci, line 58: Any specific reason to go for italics instead of bold here?
```suggestion
A *zone* is a named area inside the layout where cards can be placed. A zone is defined using `ui.zone()`.
```

mturoci (changes requested):
Thanks @zbanks23! Looks good overall, just could use more clarity.

website/docs/layout.md
zbanks23: I am not sure why this line had a change. The original doc had "zone" as italic.

PR #2488, opened 2026-04-04
Security: Server-Side Script Injection from File Content in Inline JavaScript

tuanaiseo:
## Problem

File contents are interpolated directly into an inline JavaScript template literal: `editor.setValue(`{file_content}`)`. If a file contains backticks or `${...}`, it can break out of the string/template context and execute arbitrary JavaScript in the browser (XSS).

**Severity**: `high`
**File**: `studio/studio_editor.py`

## Solution

Never embed raw file content in executable JS strings. Send content as JSON-encoded data (`json.dumps`) and parse safely client-side, or deliver via API response and set editor text without code interpolation.

## Changes

- `studio/studio_editor.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/studio/studio_editor.py
+++ b/studio/studio_editor.py
@@ -7,8 +7,8 @@ def update_file_tree(q: Q, root: str) -> None:
 
 def open_file(q: Q, file: str) -> None:
     q.page['meta'].script = ui.inline_script(f'''
-editor.setValue(`{file_utils.read_file(file)}`)
-eventBus.emit('activeFile', '{file}')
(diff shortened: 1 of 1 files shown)


mturoci:
Thanks @tuanaiseo! The Wave Studio app is a single-user app that allows remote code execution on purpose, giving people full access to the underlying k8s pod (or their local dev env) it's currently running on.

This change may fix a small security hole, but there is a much larger one anyway.

Issue #2243, opened 2024-01-31
Add link to `ui.tab` to direct user to external URL when clicked on


nipunchamikara:
#### Is your feature request related to a problem? Please describe

I have a bunch of tabs which I use for navigation. I want the last one to link to documentation which is an external URL. However, I don't see a way to open an external URL using a tab.

#### Describe the solution you'd like

Ideally, the ability to specify an external URL in the tab, maybe a property called `link` just like a button.

#### Describe alternatives you've considered

For now, external URLs were opened by setting inline script to `f'window.open("{external_url}");'`. However, this is not ideal since the page has to be re-rendered. Do mention if there are any _more elegant_ solutions to this.

rutujmirzapure:
Introduce a link property in ui.tab, similar to ui.button. When provided, the link property would override the tab's default behavior and redirect users to the specified URL. If not provided, the tab functions as it currently does, emitting an event to the backend.
`from h2o_wave import main, app, Q, ui

@app('/')
async def serve(q: Q):
    if not q.client.initialized:
        # Example usage of the proposed 'link' property in ui.tab
        q.page['tabs'] = ui.tab_card(
            box='1 1 4 1',
            items=[
                ui.tab(name='home', label='Home'),  # Regular tab
                ui.tab(name='about', label='About'),  # Regular tab
                ui.tab(name='documentation', label='Documentation', link='https://wave.h2o.ai')  # Tab with external link
            ]
        )
        q.page['content'] = ui.markdown_card(
            box='1 2 4 4',
            title='Welcome',
            content='Select a tab to navigate.'
        )
        q.client.initialized = True

    # Backend logic for tabs that do not use the 'link' property
    if q.args.home:
        q.page['content'].content = 'You are on the Home tab.'
    elif q.args.about:
        q.page['content'].content = 'You are on the About tab.'

    await q.page.save()
`


It will Simplifies navigation to external URLs without the need for custom scripts.Enhances the flexibility and usability of the ui.tab component.Reduces re-rendering overhead for static external links.

If this feature is not immediately available, users can achieve similar behavior with ui.inline_script:

`q.page['script'] = ui.inline_script("""
    document.querySelector('[data-tab-name="documentation"]').addEventListener('click', () => {
        window.open('https://wave.h2o.ai', '_blank');
    });
""")
`

PR #2501, opened 2026-08-20
feat: Add external links to tabs #2243

himaSH97:

### DIFF
--- a/ui/src/tab.test.tsx
+++ b/ui/src/tab.test.tsx
@@ -39,6 +39,9 @@ describe('Tab.tsx', () => {
     wave.args[name] = null
     jest.clearAllMocks()
   })
+  afterEach(() => {
+    jest.restoreAllMocks()
+  })
 
   it('Renders data-test attr', () => {
     const { queryByTestId } = render(<View {...getProps()} />)
@@ -194,12 +197,16 @@ describe('Tab.tsx', () => {
     const path = 'https://wave.h2o.ai/docs/getting-started'
     const windowOpenMock = jest.spyOn(window, 'open').mockImplementation(() => null)
     const { getAllByRole } = render(<View {...{ ...getProps(), state: { items: [{ name: 'tab1' }, { name: 'docs', label: 'Docs', path }] } }} />)
-    fireEvent.click(getAllByRole('tab')[1])
+    const link = getAllByRole('tab')[1]
 
+    expect(link.tagName).toBe('A')
+    expect(link).toHaveAttribute('href', path)
+
+    fireEvent.click(link)
+    
     expect(windowOpenMock).toHaveBeenCalledTimes(1)
     expect(windowOpenMock).toHaveBeenCalledWith(path, '_blank')
     expect(pushMock).toHaveBeenCalledTimes(0)
-    windowOpenMock.mockRestore()
   })
 
 })
--- a/ui/src/tab.test.tsx
+++ b/ui/src/tab.test.tsx
@@ -190,4 +190,15 @@ describe('Tab.tsx', () => {
     expect(getAllByRole('tab')[0]).toHaveClass('is-selected')
   })
 
+  it('Opens link in a new tab when tab has path specified', () => {
+    const path = 'https://wave.h2o.ai/docs/getting-started'
+    const windowOpenMock = jest.fn()
+    window.open = windowOpenMock
+    const { getAllByRole } = render(<View {...{ ...getProps(), state: { items: [{ name: 'tab1' }, { name: 'docs', label: 'Docs', path }] } }} />)
+    fireEvent.click(getAllByRole('tab')[1])
+
+    expect(windowOpenMock).toHaveBeenCalledWith(path, '_blank')
+    expect(pushMock).toHaveBeenCalledTimes(0)
+  })
+
 })
--- a/ui/src/tab.test.tsx
+++ b/ui/src/tab.test.tsx
@@ -38,8 +38,6 @@ describe('Tab.tsx', () => {
     wave.args = [] as any
     wave.args[name] = null
     jest.clearAllMocks()
-  })
-  afterEach(() => {
     jest.restoreAllMocks()
   })
 
--- a/ui/src/tab.test.tsx
+++ b/ui/src/tab.test.tsx
@@ -192,13 +192,14 @@ describe('Tab.tsx', () => {
 
   it('Opens link in a new tab when tab has path specified', () => {
     const path = 'https://wave.h2o.ai/docs/getting-started'
-    const windowOpenMock = jest.fn()
-    window.open = windowOpenMock
+    const windowOpenMock = jest.spyOn(window, 'open').mockImplementation(() => null)
     const { getAllByRole } = render(<View {...{ ...getProps(), state: { items: [{ name: 'tab1' }, { name: 'docs', label: 'Docs', path }] } }} />)
     fireEvent.click(getAllByRole('tab')[1])
 
+    expect(windowOpenMock).toHaveBeenCalledTimes(1)
     expect(windowOpenMock).toHaveBeenCalledWith(path, '_blank')
     expect(pushMock).toHaveBeenCalledTimes(0)
+    windowOpenMock.mockRestore()
   })
 
 })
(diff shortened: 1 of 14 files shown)


himaSH97:
Fixes #2243

## Summary

Adds an optional `link` property to `ui.tab`.

When `link` is set, clicking the tab navigates to the supplied external URL.
Tabs without `link` retain their current behavior and continue emitting backend
events.

### Tests
image

ui/src/tab.test.tsx
```diff
@@ -190,4 +190,15 @@ describe('Tab.tsx', () => {
     expect(getAllByRole('tab')[0]).toHaveClass('is-selected')
   })
 
+  it('Opens link in a new tab when tab has path specified', () => {
+    const path = 'https://wave.h2o.ai/docs/getting-started'
+    const windowOpenMock = jest.fn()
+    window.open = windowOpenMock
```
mturoci: This alters `window.open` for all the tab tests and never cleans it up to the original impl. Maybe we could use `spyOn` instead since you just need to check how it was called?

ui/src/tab.test.tsx
```diff
@@ -190,4 +190,15 @@ describe('Tab.tsx', () => {
...
+    const { getAllByRole } = render(<View {...{ ...getProps(), state: { items: [{ name: 'tab1' }, { name: 'docs', label: 'Docs', path }] } }} />)
+    fireEvent.click(getAllByRole('tab')[1])
+
+    expect(windowOpenMock).toHaveBeenCalledWith(path, '_blank')
```
mturoci, line 206: Let's also make the test stricter and assert the mock has been called exactly once.

ui/src/tabs.test.tsx
```diff
@@ -176,4 +176,15 @@ describe('Tabs.tsx', () => {
     expect(getAllByRole('tab')[0]).toHaveClass('is-selected')
   })
 
+  it('Opens link in a new tab when tab has path specified', () => {
```
mturoci, line 180: Same comments as in the test above.

ui/src/tab.tsx
```diff
@@ -49,6 +49,11 @@ export const
       onLinkClick = (item?: PivotItem) => {
         const name = item?.props.itemKey
         if (!name || valueB() === name) return
+        const tab = state.items.find(t => t.name === name)
+        if (tab?.path) {
+          window.open(tab.path, '_blank')
+          return
+        }
```
mturoci, line 56: why do we need to handle this within `onLinkClick`? Isn't `headerButtonProps` enough?

website/widgets/content/tabs.md
```diff
@@ -44,3 +44,18 @@ q.page['example'] = ui.tab_card( box='1 1 4 1', link=True, value='#about', items
...
+## External link tab
+
+Use the `path` attribute to open an external URL in a new browser tab when clicked.
+
+```py
+q.page['example'] = ui.tab_card(box='1 1 4 1', items=[
+        ui.tab(name='#menu/spam', label='Spam'),
+        ui.tab(name='#menu/ham', label='Ham'),
+        ui.tab(name='#menu/eggs', label='Eggs'),
+        ui.tab(name='#about', label='About'),
```
mturoci:
```suggestion
```

mturoci (changes requested):
Thanks @himaSH97! The PR looks good, just a few comments/questions.

website/widgets/content/tabs.md
himaSH97: Removed from the docs

ui/src/tab.tsx
himaSH97: Yes, my mistake. 
I’ll remove the manual window.open call and add an early return for tabs with a path, since onClick is always called.

ui/src/tab.test.tsx
himaSH97: I replaced the direct window.open assignment with jest.spyOn(window, 'open') and restore the original implementation with mockRestore() after the test.

ui/src/tab.test.tsx
himaSH97: Added

ui/src/tabs.test.tsx
himaSH97: Similar changes added

ui/src/tab.test.tsx
mturoci: What if the test fails before reaching the mockRestore? The spy needs to be within module scope and the cleanup called in afterEach.

ui/src/tabs.test.tsx
mturoci: The test is missing the assertion for whether the new window (the desired behavior) was opened with given link. Do we need to spy on something else then?

ui/src/tab.tsx
himaSH97: Hey @mturoci, 
I tested both `tab.tsx` and `tabs.tsx` components. It seems like `headerButtonProps` is enough for navigation but it actually does not open a new tab. 

This issue is as described below,

```<Pivot onLinkClick={onLinkClick}>
  <PivotItem
    headerButtonProps={{
      href: path,
      target: '_blank',
      rel: 'noopener noreferrer',
    }}
  />
</Pivot>
```
When we add a `href` here it renders as an `<a>` tag but in `Pivot` component from FluentUI has a preventDefault attched to its `onLinkClick` function. Hence default navigations won't work.

I would like to know your opinion on this before addressing the tests you have mentioned.

mturoci: I see. That means your original solution was indeed correct, right? `headerButtonProps` is needed to get the browser link rendered, while `window.open` does the actual job of opening the url.

himaSH97: Yes, I am thinking of using both window.open for open the new tab and headerButtonProps to make sure the external navigation is rendered as an `a` tag. 

Also will update the test function to check both these requirements

mturoci: Sounds good to me. Thanks!

ui/src/tab.test.tsx
himaSH97: I found this on jest docs 
https://archive.jestjs.io/docs/en/23.x/jest-object#jestrestoreallmocks

According to the docs it restores all the mocks created by `jest.spyOn` and wont affect existing `pushMock`.

Let me know your thoughts

ui/src/tabs.test.tsx
himaSH97: Updated both tests for new specs

mturoci (commented):
Code LGTM.

Found 1 minor styling bug for visited link state. Would be better if the visited link style would be the same as non visited (respecting theme color).

https://github.com/user-attachments/assets/58c0c794-1406-479d-bc99-b6abc8daeae1

ui/src/tab.test.tsx
mturoci: Seems fine, thanks!

himaSH97:
Added a small fix to above styling mismatch. Let me know if this works.

I have added a prevent default for the `mouseDown` event. it breaks the forcus from the rendered `a` tag and use the default styling.

mturoci:
Isn't there a way to just override :visited style for the link? Would be more explicit about the actual intent. The JS solution may also potentially turn off other browser behavior we would desire. Better to just fix styling and not touch anything else.

himaSH97:
Hi @mturoci,
I tried adding a custom class with the existing css styling but this method gets overridden by pivot link styles

```className: css(className, isSelected && classNames.linkIsSelected) ```

I found it is possible to do this with custom html attribute (`'data-external-tab': true`) and use that to render the styles which pivot won't override

```
'.ms-Pivot-link[data-external-tab]:hover, .ms-Pivot-link[data-external-tab]:hover .ms-Pivot-linkContent': {
            color: cssVar('$card')
   }, 
```
          
is this a good approach to use here?

mturoci:
Hey @himaSH97. See https://github.com/h2oai/wave/pull/2501/changes/b958521b024a04dd4f576004dc156da1c2cc845d for the proper solution. 

I assume your attempts just needed an ` !important` rule. However, we do not want to scope the fix to just tab components, but potentially globally to all the places that framework may need to render links within a Pivot items. The commit above fixes them all uniformly.

I tested around and couldn't find any more issues. Please do the same and let me know in case you still find some bug. If not, I we are good to merge this.

mturoci (approved):

PR #2504, opened 2026-09-12
fix: Percent-encode file URLs in Site/AsyncSite download() and unload()

### LINKED ISSUE #2503: site.download()/unload() fail for uploaded files whose name contains '#' or '?'
#### Wave SDK Version, OS

1.8.13 (also reproduced on 1.8.10), Linux

#### Actual behavior

When a user uploads a file through `ui.file_upload` whose name contains `#` (e.g. `guide #1.pdf`), waved stores it fine and returns the URL `/_f/<uuid>/guide #1.pdf` in `q.args`. Passing that URL to `q.site.download()` raises:

```
h2o_wave.core.ServiceError: Download failed (code=404): Not Found
```

`q.site.unload()` fails the same way, so the file also cannot be cleaned up. The same happens for `?` in the filename. Both `Site` and `AsyncSite` are affected.

#### Expected behavior

`download()` and `unload()` accept the URL exactly as waved returned it, whatever characters the filename contains.

#### Root cause

`Site.download`/`unload` and `AsyncSite.download`/`unload` build the request with plain string concatenation:

```python
async with self._http.stream('GET', f'{_config.hub_host_address}{url}') as res:
```

httpx then parses `#` as the start of a URL fragment (and `?` as a query), so the request that reaches waved is `GET /_f/<uuid>/guide ` and the file server answers 404.

#### Steps To Reproduce

```python
import asyncio, httpx, json
from h2o_wave import AsyncSite

async def main():
    r = httpx.post('http://localhost:10101/_f/', files=[('files', ('guide #1.pdf', b'%PDF-1.4 fake'))])
    url = json.loads(r.text)['files'][0]   # '/_f/<uuid>/guide #1.pdf'
    await AsyncSite().download(url, 'out.pdf')  # ServiceError: Download failed (code=404)

asyncio.run(main())
```

Percent-encoding the path before the request (`urllib.parse.quote(url)`) makes the same download succeed, because Go decodes the escaped path on the server side. A PR with that fix for the Python client follows.

A related but separate issue: since waved echoes the raw filename in the upload URL, the value in `q.args` is not a valid URL and also breaks when used directly as `ui.link(path=...)`. Having the file server return `url.PathEscape(filename)` would fix that end to end.

pascal-pfeiffer:
waved echoes the raw filename in the upload URL, and download()/unload() built the request by plain string concatenation, so a '#' or '?' in the filename was parsed as a URL fragment/query and the request 404'd.

Quote the URL before the request (unquoting first so already-encoded URLs are not double-encoded) and keep the decoded basename when downloading into a directory.

closes #2503


mturoci:
Thanks @pascal-pfeiffer. IMO this is something that should be fixed on the app-level code. We could change 404 to 400 with a more descriptive error tho.

Even though this change seems innocent, I would prefer avoiding any (no matter how small) risk of breaking existing apps, given the current maintenance mode of the project.

pascal-pfeiffer:
> Thanks @pascal-pfeiffer. IMO this is something that should be fixed on the app-level code. We could change 404 to 400 with a more descriptive error tho.
> 
> Even though this change seems innocent, I would prefer avoiding any (no matter how small) risk of breaking existing apps give the current maintenance mode of the project.

Thank you for the explanation and I can totally understand why you would avoid any risk of breaking existing apps. Please feel free to dismiss and close this PR and close the issue as unplanned. 
We did apply the fix on our app-level code, so there isn't any blocker for us.

mturoci:
Good to hear, thanks!

